PR #35, opened 2023-09-18
feat: Basic Windows support

gmchale79:

The DLC worker agent did not start on Windows, and numerous unit tests did not pass. 

Numerous small changes related to removing `NotImplemented` exceptions, using more appropriate file paths etc. 

User impersonation is not intended to work in this branch at this time, and several of the changes relate to ensuring tests pass when impersonation is not used. 

**As of 11/6/23 this is targeted at a feature branch, not mainline.** 

The worker agent starts, then receives and executes jobs on Windows. 

Submitting jobs to gamma and ensuring they are picked up and executed by a Windows worker. 

No currently. 

It is not intended to be.

### DIFF
--- a/src/deadline_worker_agent/aws_credentials/aws_configs.py
+++ b/src/deadline_worker_agent/aws_credentials/aws_configs.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import stat
-
+import os
 import logging
 from abc import ABC, abstractmethod
 from configparser import ConfigParser
@@ -11,6 +11,7 @@ from pathlib import Path
 from typing import Optional
 from openjd.sessions import PosixSessionUser, SessionUser
 from subprocess import run, DEVNULL, PIPE, STDOUT
+from ..set_windows_permissions import set_user_restricted_path_permissions
 
 __all__ = [
     "AWSConfig",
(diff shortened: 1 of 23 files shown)


jusiskin (changes requested):
Thank you for taking on the first pass of adding Windows support to the worker agent. This is a great first step towards an important feature. It seems like this is a partial implementation, which is fine.

~~The biggest blocker to shipping this code is setting expectations for those using it. Specifically, if we don't yet support isolation of OS users then that limitation and its associated risks need to be made clear~~

EDIT: Sorry, I missed that this was targeting the `feature_windows` branch. I presume that we will have OS user impersonation when we merge this feature branch into `mainline`. On that basis, we can disregard my concern above.

ddneilson (changes requested):

ddneilson (approved):

jusiskin (changes requested):

ddneilson (dismissed):

mwiebe (changes requested):

mwiebe (approved):

anandgo1 (approved):

jusiskin (approved):
This looks good to me.

A few small remaining nit-picks but they are not critical for merging

PR #58, opened 2023-10-16
feat(install): Adding vfs install path to install parameters

natmatn:
When launching the worker agent moved to systemd being spun up in the user data of the EC2 launch template, the FUS3_PATH environment variable in the profile.d file was no longer being exported. This resulted in deadline-cloud being unable to locate the VFS executable causing it to use the fallback loading method.

Add a `--vfs-install-path` argument to the install method which is then added to the env section of the systemd process definition. This will be set by the AMI builder so it can match the install location of the Fus3 package.

Deadline VFS install location will now be properly exported when the worker agent is started. Logging will be added to the deadline worker to make it more obvious next time the install location isn't available via the expected environment variable.
https://github.com/casillas2/deadline-cloud/pull/75

* hatch run test
* Created a CMF using the deadline worker AMI, used the install script to install the deadline-worker service on said ami, launch the systemd service through the user data in Ec2 launch template user data section, and then confirmed through the cloudwatch logs that the VFS install location was found using the environment variable set in the service environment variables.

No

No

### DIFF
--- a/src/deadline_worker_agent/installer/__init__.py
+++ b/src/deadline_worker_agent/installer/__init__.py
@@ -13,8 +13,6 @@ INSTALLER_PATH = {
     "linux": Path(__file__).parent / "install.sh",
 }
 
-VFS_DEFAULT_INSTALL_PATH = "/opt/deadline_vfs"
-
 
 def install() -> None:
     """Installer entrypoint for the Amazon Deadline Cloud Worker Agent"""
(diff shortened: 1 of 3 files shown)


ddneilson (changes requested):
I don't like the idea of having an app-specific configuration in the agent's installation; especially when that app is specific to one type of worker. Specifically, I don't like the precedent that it sets -- do we add app-specific configurations for every possible application that a job might need to interact with?

If we, instead, solve the more general problem (needing to have environment variable definitions added to the systemd script) then the installation can use the general mechanism. 

What do you think about changing this from a vfs-specific feature to one where we allow the user to define a set of environment variables as part of the installation? The check for existence of the vfs binary would then be done outside of this script, and need to be done by the scripting that is calling the agent installation script prior to installing the agent.

ddneilson (approved):

viknith (changes requested):

viknith (approved):
Lgtm!

PR #289, opened 2024-04-02
fix: provide region in queue AWS configuration

jusiskin:

The worker agent was not specifying a `region` in the `config` file it provides in the [queue AWS credentials directory](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/docs/state.md#queue-aws-credentials).

If the worker host had not been provisioned in a way where AWS SDKs could resolve the region, then this would cause jobs to fail with:

```
ERROR: You must specify a region.
```

Write the `config` file with a `region` set to the same region that the agent is configured to use.

The actions within worker sessions will be configured to have the same region as the worker agent is using. Customers can override this behavior with a queue environment,

*   Unit tests were updated and pass
*   Ran the agent in a customer-managed fleet and confirmed that adaptors could successfully make boto calls that require a region

No. This was the intended behavior. Session actions should be pre-configured such that if they use AWS SDKs, those SDKs will use the same region as the worker.

No

----

### DIFF
--- a/src/deadline_worker_agent/aws_credentials/aws_configs.py
+++ b/src/deadline_worker_agent/aws_credentials/aws_configs.py
@@ -107,7 +107,11 @@ class _AWSConfigBase(ABC):
         # finally, read the config
         self._config_parser.read(self.path)
 
-    def install_credential_process(self, profile_name: str, script_path: Path) -> None:
+    def install_credential_process(
+        self,
+        profile_name: str,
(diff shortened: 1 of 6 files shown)


marofke (approved):
Thanks for your quick work here Josh, much appreciated!

:ship: 🐋

prestomation (approved):

PR #293, opened 2024-04-10
fix: handle case where BatchGetJobEntity returns no jobRunAsUser

ddneilson:

The jobRunAsUser response field of the JobDetailsEntity structure within a BatchGetJobEntity request is an optional field. This Worker Agent currently treats it as a required field; resulting in failed Sessions when the field is absent.

Handle the case of missing the jobRunAsUser field in the response. It is a service invariant that this field is always be present when the Worker is within a CMF. When the Worker is running within an SMF, then the field may be optional; but, further, in this case the Agent will always be running with a local jobRunAsUser override. We fail Sessions when this invariant is violated.

Sessions should no longer fail if a Queue is configured to not have a `jobRunAsUser` and the Worker is running within a service managed fleet.

I've added unit test coverage of all cases, and also ran the code locally against a CMF under a variety of configurations and submitted a sleep job to queues to test.

In these tests `queue-fcda40de72c64476aac2addf8e997a85` is configured with `QUEUE_CONFIGURED_USER`, and `queue-e17782957b0d49a5acef1a6ed690b392` is configured for `WORKER_AGENT_USER`.

1. Submitting the sleep job to both queues, and picking them up with a "normally" configured worker:
```
+ deadline-worker-agent --no-shutdown --farm-id farm-591c678b072c493c90622087500add9c --fleet-id fleet-e1736871b91b47148d1b25bc7929b866
...
[2024-04-10 16:29:10,931][INFO    ] 🔷 Session.User 🔷 [session-9ec374d3d45b4ee19132965165258560] Running as Queue's jobRunAsUser. (User: jobuser) [queue-fcda40de72c64476aac2addf8e997a85/job-fb01ad60cfd040e491fd64127897f1f6]
...
[2024-04-10 16:29:44,091][WARNING ] 🔷 Session.User 🔷 [session-33279e091c2a4ef89d4cef5a1f85ba13] Running as the Worker Agent's user. (User: agentuser) [queue-e17782957b0d49a5acef1a6ed690b392/job-6bed0407db9b4320a579cf7788f0b814]
...
```

2. Configuring the worker agent to run all as the agent user and submitting to the `QUEUE_CONFIGURED_USER` queue:

```
+ deadline-worker-agent --no-shutdown --farm-id farm-591c678b072c493c90622087500add9c --fleet-id fleet-e1736871b91b47148d1b25bc7929b866 --run-jobs-as-agent-user
...
[2024-04-10 16:32:36,092][WARNING ] 🔷 Session.User 🔷 [session-b442b4d20eba45eaa672588307d07025] Running as the Worker Agent's user. (User: agentuser) [queue-fcda40de72c64476aac2addf8e997a85/job-bffead121b674cea89ae6fdfc29c2b10]
...
```

3. Configuring the worker agent to override the jobRunAsUser and submitting to the `WORKER_AGENT_USER` queue:

```
+ deadline-worker-agent --no-shutdown --farm-id farm-591c678b072c493c90622087500add9c --fleet-id fleet-e1736871b91b47148d1b25bc7929b866 --posix-job-user jobuser:jobuser
...
[2024-04-10 16:35:46,601][INFO    ] 🔷 Session.User 🔷 [session-91b6bf4e58fd47de83d9c3e358611dfc] Running as host-configured override user. (User: jobuser) [queue-e17782957b0d49a5acef1a6ed690b392/job-a2798841947c4cca85824aed8674429e]
...
```

N/A

No, it should not be.

----

### DIFF
--- a/src/deadline_worker_agent/scheduler/scheduler.py
+++ b/src/deadline_worker_agent/scheduler/scheduler.py
@@ -756,7 +756,7 @@ class WorkerScheduler:
                     job_id=job_id,
                     session_id=session_id,
                     user=user_to_log,
-                    message="Running as the Worker Agent's user.",
+                    message="Running as the Worker Agent's user. This configuration is not recommended; please see the Security chapter of the User Guide.",
                 )
             )
         return os_user
--- a/src/deadline_worker_agent/scheduler/scheduler.py
+++ b/src/deadline_worker_agent/scheduler/scheduler.py
@@ -37,6 +37,7 @@ from ..sessions.log_config import (
     LogProvisioningError,
     SessionLogConfigurationParameters,
 )
+from ..sessions.job_entities.job_details import JobRunAsUser
 from ..api_models import (
     AssignedSession,
     UpdateWorkerScheduleResponse,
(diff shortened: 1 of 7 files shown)


src/deadline_worker_agent/scheduler/scheduler.py
```diff
@@ -668,6 +669,98 @@ def _fail_all_actions(
...
+                            message="Running as Queue's jobRunAsUser.",
+                        )
+                    )
+        if os_user is None:
+            try:
+                user_to_log = getpass.getuser()
+            except Exception:
+                # This is best-effort. If we cannot determine the user we will not log
+                user_to_log = "UNKNOWN"
+            logger.warning(
```
epmog, line 752: Not sure the user/operator has any insight why this is a warning log and not just an info. Should we swap it to info? Or add additional information?

src/deadline_worker_agent/sessions/job_entities/job_details.py
```diff
@@ -91,33 +91,39 @@ def job_run_as_user_api_model_to_worker_agent(
     """Converts the 'JobRunAsUser' api model to the 'JobRunAsUser' dataclass
     expected by the Worker Agent.
     """
-    if "runAs" in job_run_as_user_data and job_run_as_user_data["runAs"] == "WORKER_AGENT_USER":
+    # Only two options for "runAs": WORKER_AGENT_USER & QUEUE_CONFIGURED_USER
+    if job_run_as_user_data["runAs"] == "WORKER_AGENT_USER":
```
epmog, line 95: `runAs` is required now, so I believe we have to update the api_model file to indicate as such? it's currently marked as `NotRequired`

src/deadline_worker_agent/scheduler/scheduler.py
ddneilson: It's a warning 'cause running as the agent user is not the recommended configuration. We could elaborate in the logging message a bit to make that clearer.

src/deadline_worker_agent/sessions/job_entities/job_details.py
ddneilson: Good catch. I'll update that.

src/deadline_worker_agent/scheduler/scheduler.py
```diff
@@ -668,6 +669,98 @@ def _fail_all_actions(
...
+                        user=os_user.user,
+                        message="Running as host-configured override user.",
+                    )
+                )
+            elif job_run_as_user is None:
+                # Terminal error. We need to fail the Session.
+                # This should *never* happen; it occuring would mean that a service invariant has
+                # been violated.
+                message = (
+                    "FATAL: Queue does not have a jobRunAsUser. This should not be possible. "
```
jusiskin, line 704: In this same message we should also mention that there is no job user override, since both of these conditions need to be true.

epmog (approved):

src/deadline_worker_agent/scheduler/scheduler.py
```diff
@@ -668,6 +669,98 @@ def _fail_all_actions(
...
+                if host_is_posix:
+                    os_user = job_run_as_user.posix
+                else:
+                    os_user = job_run_as_user.windows
+                if os_user is None:
+                    # Terminal error. We need to fail the Session.
+                    # This should *never* happen; it occuring would mean that a service invariant has
+                    # been violated.
+                    message = (
+                        "FATAL: Queue's jobRunAsUser does not define a QUEUE_CONFIGURED_USER for this platform. "
```
jusiskin, line 731: Should we include which platform in the message for helping diagnose? I suppose the agent logs will include the platform in the startup logging, right?

jusiskin (approved):
Approved, but posted two suggestions to improve the error messages slightly.

src/deadline_worker_agent/scheduler/scheduler.py
ddneilson: The platform is indeed already in the log.

src/deadline_worker_agent/scheduler/scheduler.py
ddneilson: True, but I'm not too worried since this will only happen if there's a service bug.

PR #305, opened 2024-04-25
ci: change to reusable workflows

moorec-aws:
We have moved the functional code for the workflows to the organization [.github repository](https://github.com/aws-deadline/.github/tree/mainline/.github/workflows) to be called as reusable workflows. This allows us to reduce the amount of locations code needs to be modified for workflow changes.

1. Remove reuse_python_build.yml since this is now stored in the .github repository
2. Change the bump, publish, code quality, integration test workflows to call the reusable workflows in the .github repository

workflows in this repository now call reusable repositories from the .gihtub repository

Tested in a development org and repository

no

no

----


.github/workflows/code_quality.yml
```diff
@@ -3,9 +3,20 @@ name: Code Quality
...
 jobs:
   Test:
-    name: Code Quality
-    uses: ./.github/workflows/reuse_python_build.yml
-    secrets: inherit
+    name: Python
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ['3.9', '3.10', '3.11']
```
jusiskin, line 18: not in scope for this PR since I know we didn't have this in the matrix previously, but can we add `3.12` here?

jusiskin (approved):

AWS-Samuel (approved):

PR #323, opened 2024-05-29
chore: fix retry formula to conform to botocore(70514ef)

amarsjac:
[Delay tests](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/test/unit/boto/test_retries.py#L52) are failing due to a change up steam in [botocore](https://github.com/boto/botocore/commit/70514effeb6a77e12993c0fbaa907c0b92aae342#diff-d3e80b391c9677c720b0d2d6328856e2c4aae989f737c78f0acf7f36bd706688L266)

change the `expected` formula in the unit tests to match the botocore formula

tests fixed, PRs and release workflows are unblocked

`hatch run test`

no

no
----

### DIFF
--- a/test/unit/boto/test_retries.py
+++ b/test/unit/boto/test_retries.py
@@ -66,7 +66,7 @@ class TestNoOverflowExponentialBackoff:
             random_between=random_between,
         )
         if is_exponential:
-            expected = min(max_backoff, random_value * base ** (attempt_number - 1))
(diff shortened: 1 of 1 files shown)


moorec-aws (approved):

crowecawcaw (approved):
Approved. But do we really need these unit tests? It looks like they're just testing botocore functionality. And in practice when botocore isn't matching our tests, we're just updating the tests to match what botocore actually does.

PR #324, opened 2024-05-31
docs: fix incorrect default log path on Linux/MacOS

jusiskin:

The documentation stated an incorrect default log path for Linux/MacOS:

```
/var/log/amazon/deadline/logs
```

The source of truth for the actual default log path is here:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/68dfc2c9c0521cca47ad1ff3bbe06fd323c18852/src/deadline_worker_agent/startup/settings.py#L18

Fix the documented default log path for Linux/MacOS with the correct value.

People reading the worker agent documentation will have accurate information about the default log paths on Linux/MacOS

View the change in GitHub markdown preview and visually confirm the path matches the  code

Yes

No

----

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -14,8 +14,8 @@ The default locations for worker logs are:
 
 | Platform | Default worker log path |
 | --- | --- |
-| Linux | `/var/log/amazon/deadline/logs` |
-| MacOS | `/var/log/amazon/deadline/logs` |
+| Linux | `/var/log/amazon/deadline` |
+| MacOS | `/var/log/amazon/deadline` |
 | Windows | `C:\ProgramData\Amazon\Deadline\Logs` |
 
 The worker agent maintains two sets of rotating log files:


ddneilson (approved):

docs/logging.md
```diff
@@ -14,8 +14,8 @@ The default locations for worker logs are:
 
 | Platform | Default worker log path |
 | --- | --- |
-| Linux | `/var/log/amazon/deadline/logs` |
-| MacOS | `/var/log/amazon/deadline/logs` |
+| Linux | `/var/log/amazon/deadline` |
+| MacOS | `/var/log/amazon/deadline` |
```
eherozhao, line 18: why does Windows have a Logs folder but the other two don't? Should we keep consistent for them?

ddneilson: It's under `/var/log` on posix-based systems. That's universally understood to be for logging files.

godobyte (approved):

PR #327, opened 2024-06-04
test: Configure integ tests to use persistent Deadline Cloud resources

moorec-aws:
Instead of using the API to create resources for tests, Deadline Cloud resource ids such as Fleets, Queues and Farms will be passed via environment variables.

Additional queues and fleets need to be added to cover a large range of tests. 

* Deadline Cloud resources, and the worker configuration are defined in the conftest since all repositories will not have the same requirements.
* Added a scaling queue and fleet for scaling test cases
* Added a non scaling queue_a, queue_b, fleet for testing non scaling uses cases as well as queue switching tests.

Tests will now use pre-existing resources

Provided farm, queue and fleet environment variables and ran the integration tests.

`hatch run integ-test`

no

no

----

### DIFF
--- a/test/integ/conftest.py
+++ b/test/integ/conftest.py
@@ -1,87 +1,281 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-import dataclasses
+import boto3
+import glob
+import json
+import logging
+import os
+import pathlib
+import posixpath
+import pytest
+import tempfile
+from dataclasses import dataclass, field, InitVar
 from typing import Generator
 
 from deadline_test_fixtures import (
-    DeadlineClient,
-    DeadlineResources,
+    DeadlineWorker,
     DeadlineWorkerConfiguration,
+    DockerContainerWorker,
     Farm,
     Fleet,
-    JobRunAsUser,
     PosixSessionUser,
     Queue,
-    QueueFleetAssociation,
+    PipInstall,
+    EC2InstanceWorker,
+    BootstrapResources,
 )
-import pytest
+
+
+LOG = logging.getLogger(__name__)
 
 
 pytest_plugins = ["deadline_test_fixtures.pytest_hooks"]
 
 
-@pytest.fixture(scope="session")
-def farm(deadline_resources: DeadlineResources) -> Farm:
-    return deadline_resources.farm
+@dataclass(frozen=True)
+class DeadlineResources:
+    farm: Farm = field(init=False)
+    queue_a: Queue = field(init=False)
+    queue_b: Queue = field(init=False)
+    fleet: Fleet = field(init=False)
+    scaling_queue: Queue = field(init=False)
+    scaling_fleet: Fleet = field(init=False)
+
+    farm_id: InitVar[str]
+    queue_a_id: InitVar[str]
+    queue_b_id: InitVar[str]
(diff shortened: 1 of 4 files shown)


ttblanchard:
Does it make more sense to create these new fixtures in the test fixtures repository?

moorec-aws:
> Does it make more sense to create these new fixtures in the test fixtures repository?
The idea was to pull these fixtures out of the test-fixtures repository since each repository could have different requirements, resources.

test/integ/test_job_submissions.py
```diff
@@ -59,16 +54,15 @@ def test_success(
...
+        deadline_resources,
+        job_run_as_user,
     ) -> None:
         # WHEN
         job = Job.submit(
             client=deadline_client,
-            farm=farm,
-            queue=queue_with_job_run_as_user,
+            farm=deadline_resources.farm,
+            queue=deadline_resources.scaling_queue,
```
jusiskin: Should we be using the `scaling_queue` here? Do we have the logic that sets up the scaling fleet?

moorec-aws: It is simply a queue with an associated autoscaling fleet with the autoscaling mode enabled at the moment. Should this be switched to one with no scaling?

moorec-aws: updated to use a non scaling queue

jusiskin (approved):

jusiskin (approved):

test/integ/conftest.py
```diff
@@ -1,87 +1,281 @@
...
-    deadline_client: DeadlineClient,
-    job_run_as_user: PosixSessionUser,
-) -> Generator[Queue, None, None]:
-    queue = Queue.create(
-        client=deadline_client,
-        display_name=f"Queue with jobsRunAsUser {job_run_as_user.user}",
-        farm=farm,
-        job_run_as_user=JobRunAsUser(posix=job_run_as_user, runAs="QUEUE_CONFIGURED_USER"),
-    )
+def worker(request: pytest.FixtureRequest, worker_config) -> Generator[DeadlineWorker, None, None]:
```
ttblanchard, line 189: Why are we redefining this fixture that already exists here: https://github.com/aws-deadline/deadline-cloud-test-fixtures/blob/mainline/src/deadline_test_fixtures/fixtures.py#L473

test/integ/conftest.py
```diff
@@ -1,87 +1,281 @@
...
+        queue_a_id=queue_a_id,
+        queue_b_id=queue_b_id,
+        fleet_id=fleet_id,
+        scaling_queue_id=scaling_queue_id,
+        scaling_fleet_id=scaling_fleet_id,
     )
 
 
 @pytest.fixture(scope="session")
 def worker_config(
```
ttblanchard, line 107: And this one? https://github.com/aws-deadline/deadline-cloud-test-fixtures/blob/mainline/src/deadline_test_fixtures/fixtures.py#L381

test/integ/conftest.py
moorec-aws: This should have been defined here originally using the DeadlineWorker, and DeadlineWorkerConfig classes from deadline-cloud-test-fixtures. Since each repository could have different implementations of the worker and its configuration.

ttblanchard (approved):

PR #331, opened 2024-06-13
test: Add Windows E2E Tests

moorec-aws:
Add windows E2E tests, depends on https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/115

* Move integ tests to e2e tests folder.
* Split e2e tests into windows and linux
* Add additional hatch calls for windows/linux
* Add windows job submission test
* Add a windows e2e test workflow for testing

#### Updated Rev adds:
* Removed codebuild environment as it is an artifact from an old hatch workaround for codeartifact that is not needed anymore.
* Updated the pipeline e2e scripts to support both linux and windows. It runs the hatch command based on what operating system is detected. Environment variable is set [here](https://github.com/aws-deadline/.github/pull/10).
* Updated canary script to run the e2e test script.
* During testing I found that Windows Workers were not shutting down, and switch to the `not responding` status. This is because the service is not handling shutdown properly. We needed to add `SvcShutdown` to call `SvcStop` to shutdown the worker when an instance termination occurs.

Adds windows e2e tests and allows for calling linux or windows tests.

`hatch run linux-e2e-test`
`hatch run windows-e2e-test`

Confirmed tests succeeded with:
`./pipelines/canary.sh`
`./pipeline/e2e.sh`

No

No

----


requirements-testing.txt
```diff
@@ -1,15 +1,15 @@
 coverage[toml] ~= 7.5
 coverage-conditional-plugin == 0.9.*
-deadline-cloud-test-fixtures == 0.6.*
 pytest ~= 8.2
 pytest-cov == 5.0.*
+/home/moorec/Projects/git/deadline-cloud-test-fixtures/dist/deadline_cloud_test_fixtures-0.6.2.post18+g7526d2c.tar.gz
```
ttblanchard: typo

moorec-aws: This will be updated once https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/115 is merged.

.github/workflows/windows_e2e_test.yml
```diff
@@ -0,0 +1,18 @@
+name: Mainline Windows E2E Test
+
+on:
+  workflow_dispatch:
+
```
epmog, line 5: If I'm understanding it won't fire automatically yet and it's valid syntax even without a value?

test/e2e/conftest.py
```diff
@@ -0,0 +1,299 @@
...
+def job_run_as_user() -> PosixSessionUser:
+    return PosixSessionUser(
+        user="jobuser",
+        group="shared-group",
+    )
+
+
+@pytest.fixture(scope="session")
+def operating_system(request) -> str:
+    if request.param == "linux":
```
jusiskin: This fixture is not parametrized. Should this be:

```suggestion
@pytest.fixture(scope="session", params=("linux", "windows"))
def operating_system(request) -> str:
    if request.param == "linux":
```

test/e2e/conftest.py
```diff
@@ -0,0 +1,299 @@
...
+def region() -> str:
+    return os.getenv("REGION", os.getenv("AWS_DEFAULT_REGION", "us-west-2"))
+
+
+@pytest.fixture(scope="session")
+def job_run_as_user() -> PosixSessionUser:
+    return PosixSessionUser(
+        user="jobuser",
+        group="shared-group",
+    )
```
jusiskin: should this have different beahviour based on the `operating_system` fixture?

test/e2e/linux/test_job_submissions.py
```diff
@@ -13,30 +13,26 @@
...
+@pytest.mark.parametrize("operating_system", ["linux"], indirect=True)
 class TestJobSubmission:
     def test_success(
         self,
-        deadline_client: DeadlineClient,
-        farm: Farm,
-        queue: Queue,
+        deadline_client,
+        deadline_resources,
     ) -> None:
```
jusiskin: Why are we removing all of the type annotations?

.github/workflows/windows_e2e_test.yml
moorec-aws: That is correct.

pipeline/e2e.sh
```diff
@@ -0,0 +1,20 @@
...
+  then
+    hatch run codebuild:build
+    export WORKER_AGENT_WHL_PATH=dist/`hatch run codebuild\:metadata name | sed 's/-/_/g'`-`hatch run codebuild\:version`-py3-none-any.whl
+    echo "Set WORKER_AGENT_WHL_PATH to $WORKER_AGENT_WHL_PATH"
+  fi
+else
+  continue
+fi
+
+hatch run codebuild:integ-test
```
jusiskin: nit: can we rename this and the hatch script to `e2e-test`?

ttblanchard (dismissed):

epmog (dismissed):

jusiskin (dismissed):

pipeline/e2e.sh
moorec-aws: Yes, I've updated it to support both linux/windows e2e tets.

test/e2e/conftest.py
moorec-aws: Yes, but this will come in another PR. It doesn't effect the current tests.

test/e2e/conftest.py
moorec-aws: Updated

test/e2e/conftest.py
```diff
@@ -145,7 +146,12 @@ def worker_config(
         ), f"Expected exactly one Worker agent whl path, but got {resolved_whl_paths} (from pattern {worker_agent_whl_path})"
         resolved_whl_path = resolved_whl_paths[0]
 
-        dest_path = posixpath.join("/tmp", os.path.basename(resolved_whl_path))
+        if operating_system.name == "AL2023":
+            dest_path = posixpath.join("/tmp", os.path.basename(resolved_whl_path))
+        else:
+            dest_path = posixpath.join(
+                "%USERPROFILE%\\AppData\\Local\\Temp", os.path.basename(resolved_whl_path)
```
ttblanchard, line 153: Does this work for you? I found I had to use `$env:USERPROFILE` see my changes here https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/125/files. Looks like our only test here is on a Linux worker.

moorec-aws: Yes, I confirmed this worked with my local whl file to test the `SvcShutdown` change.

ttblanchard: ok, interesting.

ttblanchard (approved):

epmog (approved):

PR #340, opened 2024-06-20
chore: add quotes to variables in shell scripts

moorec-aws:
We have a bunch of unquoted variables in our shell scripts

Add the quotes

n/a

`hatch run test && hatch run integ-test`

Confirmed helper resource scripts are functional.

Confirmed Worker Agent Install with Telem Opt Out is working:
```
===========================================================|      AWS Deadline Cloud Worker Agent Installer       |
===========================================================

Farm ID: farm-***************
Fleet ID: fleet-***************
Region: us-west-2
Worker agent user: deadline-worker-agent
Worker agent group: deadline-worker-agent
Worker job group: deadline-job-users
Scripts path: /opt/deadline/worker/bin
Session root directory: /sessions
Worker agent program path: /opt/deadline/worker/bin/deadline-worker-agent
Deadline client program path: /opt/deadline/worker/bin/deadline
Allow worker agent shutdown: no
Start systemd service: no
Telemetry opt-out: yes
VFS install path: unset
Disallow EC2 instance profile: no
Confirm install with the above settings (y/n):y

Worker agent user deadline-worker-agent already exists
Job group deadline-job-users already exists
Worker agent user (deadline-worker-agent) is alread in job group (deadline-job-users)
No prior sudoers shutdown rule at /etc/sudoers.d/deadline-worker-shutdown
Provisioning log directory (/var/log/amazon/deadline)
Done provisioning log directory (/var/log/amazon/deadline)
Provisioning persistence directory (/var/lib/deadline)
Done provisioning persistence directory (/var/lib/deadline)
Provisioning root directory for OpenJD Sessions (/sessions)
Done provisioning root directory for OpenJD Sessions (/sessions)
Provisioning configuration directory (/etc/amazon/deadline)
Done provisioning configuration directory
The following settings will be modified:
  farm_id = farm-********
  fleet_id = fleet-***********
  allow_ec2_instance_profile = True
  shutdown_on_stop = False
  windows_job_user = None
  session_root_dir = /sessions
Installing systemd service to /etc/systemd/system/deadline-worker.service
Done installing systemd service
Reloading systemd
Done reloading systemd
Opting out
```

No

No

----


scripts/add_queue_env.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-export ENV_JSON=$(dirname $0)/queue-env.json
+export ENV_JSON=$(dirname "$0")/queue-env.jso
```
ddneilson:
```suggestion
export ENV_JSON=$(dirname "$0")/queue-env.json
```

scripts/run_posix_docker.sh
```diff
@@ -120,10 +120,10 @@ fi
...
 docker run --rm \
     --name test_worker_agent \
-    -v $(pwd):/code:ro \
-    -v ${HOME}/.aws:/aws \
-    --env-file ${TMP_ENV_FILE} \
-    ${ARGS} \
+    -v "$(pwd)":/code:ro \
+    -v "${HOME}"/.aws:/aws \
+    --env-file "${TMP_ENV_FILE}" \
+    "${ARGS}" \
```
ddneilson:
```suggestion
    ${ARGS} \
```

I don't think that this one should be quotes. `ARGS` is built-up in this script and contains multiple command-line arguments appended together. Quoting this turns that list of args into a single argument.

testing_containers/posix_ldap_multiuser/run.sh
```diff
@@ -15,4 +15,4 @@ if test "${PIP_INDEX_URL}" != ""; then
     ENVS="${ENVS} PIP_INDEX_URL=${PIP_INDEX_URL}"
 fi
 
-sudo -u agentuser -i ${ENVS} /home/agentuser/run_agent.sh
\ No newline at end of file
+sudo -u agentuser -i "${ENVS}" /home/agentuser/run_agent.sh
```
ddneilson:
```suggestion
sudo -u agentuser -i ${ENVS} /home/agentuser/run_agent.sh
```

`ENVS` is built-up in this script by appending multiple env var definitions together. This needs to be interpreted as multiple args to the shell, so it cannot be quoted.

ddneilson (changes requested):

scripts/add_queue_env.sh
moorec-aws: Fixed

scripts/run_posix_docker.sh
moorec-aws: Fixed

testing_containers/posix_ldap_multiuser/run.sh
moorec-aws: Fixed

scripts/run_posix_docker.sh
```diff
@@ -98,32 +98,32 @@ fi
...
 if test "${AWS_ENDPOINT_URL_DEADLINE}" != ""; then
     echo "AWS_ENDPOINT_URL_DEADLINE=${AWS_ENDPOINT_URL_DEADLINE}" >> $TMP_ENV_FILE
 fi
 
 docker run --rm \
-    --name test_worker_agent \
-    -v $(pwd):/code:ro \
-    -v ${HOME}/.aws:/aws \
-    --env-file ${TMP_ENV_FILE} \
+    --name test_worker_agentss \
```
jusiskin: This looks like an unintended change, right?

This container name is referenced in a few other places:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/9367c10713391caca6bf6e256a4eb595ecda0165/DEVELOPMENT.md?plain=1#L128

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/9367c10713391caca6bf6e256a4eb595ecda0165/scripts/enter_active_test_container.sh#L7

jusiskin (changes requested):

scripts/run_posix_docker.sh
moorec-aws: Yes, that was unintended. Thanks.

scripts/run_posix_docker.sh
```diff
@@ -98,32 +98,32 @@ fi
...
 if test "${AWS_ENDPOINT_URL_DEADLINE}" != ""; then
     echo "AWS_ENDPOINT_URL_DEADLINE=${AWS_ENDPOINT_URL_DEADLINE}" >> $TMP_ENV_FILE
 fi
 
 docker run --rm \
-    --name test_worker_agent \
-    -v $(pwd):/code:ro \
-    -v ${HOME}/.aws:/aws \
-    --env-file ${TMP_ENV_FILE} \
+    --name test_worker_agents \
```
jusiskin: This is still breaks the docker workflows I think. This should now have changed and should actually be `test_worker_agent`

scripts/run_posix_docker.sh
jusiskin: Looks like there is still an unintended change from `test_worker_agent` &rarr; `test_worker_agents`. We should restore the original name (`test_worker_agent`)

jusiskin (changes requested):

jusiskin (approved):

jusiskin (dismissed):

AWS-Samuel (approved):

jusiskin (approved):

PR #346, opened 2024-07-04
fix: ignore job-entity run-as value when agent has a job user override

AWS-Samuel:
The worker agent was validating the `jobRunAsUser` field in the `BatchGetJobEntity` API response, even when the worker agent was configured with a job user override. 

This would lead to issues if a queue was configured with *only* the opposite system type in the `jobRunAsUser`. For example, a linux worker with the `posix_job_user` [configuration](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/installer/worker.toml.example#L202) receives a job, which has a `jobRunAsUser` value of: 

```
{
    "runAs": "QUEUE_CONFIGURED_USER",
    "windows": {
        "user": "windows-job-user",
        "passwordArn": "<password arn>"
    }
}
```

Then the worker agent would throw an error [here](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/9367c10713391caca6bf6e256a4eb595ecda0165/src/deadline_worker_agent/sessions/job_entities/job_details.py#L390) because there is no `posix` `runAs` field available, even though it shouldn't matter due to the override.

Ignore the `jobRunAsUser` field if the worker agent has a job user override

The erroneous behavior above no longer occurs.

Unit tests added. 

Tested on a CMF worker with and without this change. 

Confirmed that the worker with this change had a successful job run, while the worker without this change encountered the  ```Expected "jobRunAs" -> "posix" to exist when "jobRunAs" -> "runAs" is "QUEUE_CONFIGURED_USER" but it was not present``` error

No

No

----

### DIFF
--- a/test/unit/sessions/job_entities/test_job_entities.py
+++ b/test/unit/sessions/job_entities/test_job_entities.py
@@ -44,6 +44,7 @@ from deadline_worker_agent.api_models import (
     StepDetailsData,
     StepDetailsIdentifier,
 )
+from deadline_worker_agent.startup.config import JobsRunAsUserOverride
 from deadline_worker_agent.sessions.job_entities import (
     EnvironmentDetails,
     JobAttachmentDetails,
@@ -153,6 +154,7 @@ def job_entities(
         job_id=job_id,
         deadline_client=deadline_client,
         windows_credentials_resolver=windows_credentials_resolver,
+        job_run_as_user_override=None,
     )
 
     yield job_entities
@@ -225,6 +227,7 @@ class TestJobEntity:
             job_id=job_id,
             deadline_client=deadline_client,
             windows_credentials_resolver=windows_credentials_resolver,
+            job_run_as_user_override=None,
         )
 
         # WHEN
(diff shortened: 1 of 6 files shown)


ddneilson (approved):

test/unit/sessions/job_entities/test_job_entities.py
```diff
@@ -280,6 +283,57 @@ def test_job_run_as_user(
...
+            "jobRunAsUser": job_run_as_user,
+            "logGroupName": "TEST",
+            "schemaVersion": TemplateSpecificationVersion.JOBTEMPLATE_v2023_09.value,
+        }
+
+        # WHEN
+        job_details_data = JobDetails.validate_entity_data(
+            api_response, job_user_override=job_run_as_user_overrides
+        )
+        entity_obj = JobDetails.from_boto(job_details_data)
```
jusiskin, line 331: nit: shouldn't this test belong in `test/unit/sessions/job_entities/test_job_details.py` instaed, since we're testing behaviour of the `JobDetails` class?

Marked as a nit because I see this is not precedent as the test above has the same problem. 😞

jusiskin (approved):

PR #349, opened 2024-07-05
fix: jobRunAsUser always removed in BatchGetJobEntity JobDetails

AWS-Samuel:

The check here https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/c0dd3b35db2faef3296277cbf66e3dc42adeb80f/src/deadline_worker_agent/sessions/job_entities/job_details.py#L379
was not quite correct because just having an override object doesn't mean that the worker agent is actually overriding the queue user.

Based on https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/c0dd3b35db2faef3296277cbf66e3dc42adeb80f/src/deadline_worker_agent/startup/config.py#L24

Now we:
  - Check that `JobsRunAsUserOverride` is not None
  - Check that it is either:
    - Using the worker agent user
    - Using an overriden job-user

Fixes a bug which was introduced by #346 

Unit tests pass

Modified `test_job_run_as_user` and confirmed that it passed, whereas it failed with the `JobsRunAsUserOverride(run_as_agent=False, job_user=None)` change without the fix in this PR

@moorec-aws confirmed the E2E tests pass with this change

no

No
----

### DIFF
--- a/src/deadline_worker_agent/sessions/job_entities/job_details.py
+++ b/src/deadline_worker_agent/sessions/job_entities/job_details.py
@@ -376,7 +376,9 @@ class JobDetails:
                     ),
                 )
 
-        if job_user_override is not None:
+        if job_user_override is not None and (
+            job_user_override.run_as_agent or job_user_override.job_user
+        ):
             # If there is an override, we don't care about the job details jobRunAsUser
             entity_data.pop("jobRunAsUser", None)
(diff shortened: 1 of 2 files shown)


moorec-aws (approved):
integration tests passed with this change.

crowecawcaw (approved):

jusiskin (approved):

PR #352, opened 2024-07-05
feat: add Deadline Client Lib to telemetry and add a success/fail decorator…

leongdl:
- Currently we lack telemetry when Job Attachments is downloaded or output is uploaded from S3. There is insufficient metrics to know if a particular release (or dependency release) has regressions.

- Add success / fail telemetry metrics to Job Attachments download and upload functions. This is a generic try-catch metric that is useful as a canary. This will provide traceability to detect any regression.

- More traceability into Worker Agent and Job Attachments execution.

- hatch build
- hatch run fmt
- hatch run test
```
------------------------------------------------------------------------------------------------------------------------
TOTAL                                                                         4808    672   2110    204    84%
Coverage HTML written to dir build/coverage
Coverage XML written to file build/coverage/coverage.xml

Required test coverage of 80.0% reached. Total coverage: 84.42%
========================================================================================================== slowest 5 durations ==========================================================================================================
```

- Not Applicable.

- No.

----

### DIFF
--- a/src/deadline_worker_agent/aws/deadline/__init__.py
+++ b/src/deadline_worker_agent/aws/deadline/__init__.py
@@ -2,7 +2,7 @@
 
 import logging
 from time import sleep, monotonic
-from typing import Any, Dict, Optional
+from typing import Any, Callable, Dict, List, Optional, TypeVar, cast
 from threading import Event
 from dataclasses import asdict, dataclass
 import random
@@ -11,6 +11,7 @@ from botocore.retries.standard import RetryContext
 from botocore.exceptions import ClientError
 
 from deadline.client.api import get_telemetry_client, TelemetryClient
+from deadline.client import version as deadline_client_lib_version
 from deadline.job_attachments.progress_tracker import SummaryStatistics
 from openjd.model import version as openjd_model_version
 from openjd.sessions import version as openjd_sessions_version
@@ -40,6 +41,9 @@ from ...log_sync.cloudwatch import (
 __cached_telemetry_client: Optional[TelemetryClient] = None
 
 _logger = logging.getLogger(__name__)
+
+# Generic function return type.
+F = TypeVar("F", bound=Callable[..., Any])
 
 
 class DeadlineRequestError(Exception):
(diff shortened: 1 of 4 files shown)


jairaws (approved):

jairaws (dismissed):

src/deadline_worker_agent/aws/deadline/__init__.py
```diff
@@ -849,3 +856,34 @@ def record_sync_inputs_fail_telemetry_event(
...
+            """
+            success: bool = True
+            try:
+                return function(*args, **kwargs)
+            except Exception as e:
+                success = False
+                raise e
+            finally:
+                _get_deadline_telemetry_client().record_event(
+                    event_type=f"com.amazon.rum.deadline.worker_agent.{function.__name__}",
```
crowecawcaw: I liked how in another PR, you also supported explicitly adding the event name so code could be refactored without changing the telemetry. Might be useful here, e.g. to cleanup the leading "_" for one of the functions.

crowecawcaw (dismissed):

src/deadline_worker_agent/aws/deadline/__init__.py
leongdl: Good point - Let me port over the code. The reason they are different is because the namespace here is different.

I wrote this one first, and then did the "event_name" change in the deadline cloud library and noticed also the `_` private method. easy add - will do this.

leongdl: Updated!

jairaws (approved):

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1203,6 +1205,7 @@ def _handle_action_update(
             )
         )
 
+    @record_success_fail_telemetry_event(metric_name="sync_asset_outputs")  # type: ignore
```
sakshie95, line 1208: what mypy issue are we ignoring here?

sakshie95 (approved):

src/deadline_worker_agent/sessions/session.py
leongdl: mypy [misc] only on OSX. It doesn't like functions with the python `**kwargs: Dict[str, Any]` type.

PR #361, opened 2024-07-18
docs: adds missed case in worker api protocol docs

AWS-Samuel:

There was a missed case in the API protocol documentation for this code path: https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/aws/deadline/__init__.py#L598

Document it!

More accurate docs

I read over it and it looks correct ;)

By definition yes

No

----

### DIFF
--- a/docs/worker_api_protocol.md
+++ b/docs/worker_api_protocol.md
@@ -52,6 +52,7 @@ The steps to take are:
     and then retry indefinitely.
     * Response: ConflictException(409) ->
         * `reason` is `STATUS_CONFLICT`:
+            * `context["status"]` is `ASSOCIATED` -> Perform exponential backoff, and then retry.
             * `context["status"]` is `STOPPING` or `NOT_COMPATIBLE` -> Worker cannot transition to `STARTED` from its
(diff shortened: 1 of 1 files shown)


docs/worker_api_protocol.md
```diff
@@ -52,6 +52,7 @@ The steps to take are:
     and then retry indefinitely.
     * Response: ConflictException(409) ->
         * `reason` is `STATUS_CONFLICT`:
+            * `context["status"]` is `ASSOCIATED` -> Perform exponential backoff
```
ddneilson:
```suggestion
            * `context["status"]` is `ASSOCIATED` -> Perform exponential backoff, and then retry.
```

ddneilson (changes requested):

moorec-aws (approved):

docs/worker_api_protocol.md
AWS-Samuel: Amended, thanks

ddneilson (approved):

PR #364, opened 2024-07-24
test: Add linux test for job submission session action fail

YutongLi291:
A basic behaviour with Worker Agent is for it to report session actions as FAILED when the session action in question failed due to incorrect or failing commands in the job template or during the job itself.

We should have a test to make sure that quality for this part of the code is always kept high.

Add test that makes sure that session actions are correctly reported as FAILED in the cases of:
* `envEnter` — failure caused by /usr/bin/false and command not found or no executable bit
* `taskRun `— failure caused by /usr/bin/false and command not found or no executable bit
* `envExit` — failure caused by /usr/bin/false and command not found or no executable bit 

Quality for the Worker Agent code that reports failed session actions as FAILED  correctly is kept high and making sure that there will be no regressions.
`hatch run linux-e2e-test`
no
no
----

### DIFF
--- a/test/e2e/linux/test_job_submissions.py
+++ b/test/e2e/linux/test_job_submissions.py
@@ -4,21 +4,15 @@ This test module contains tests that verify the Worker agent's behavior by submi
 Deadline Cloud service and checking that the result/output of the jobs is as we expect it.
 """
 
+from typing import Any, Dict
 import boto3
 import botocore.client
 import botocore.config
 import botocore.exceptions
 import pytest
-
 import logging
 
-from deadline_test_fixtures import (
-    Job,
-    TaskStatus,
-    DeadlineClient,
-    PosixSessionUser,
-)
-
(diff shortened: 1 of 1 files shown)


test/e2e/linux/test_job_submissions.py
```diff
@@ -117,3 +111,108 @@ def test_job_run_as_user(
...
+            session_actions = deadline_client.list_session_actions(
+                farmId=job.farm.id,
+                queueId=job.queue.id,
+                jobId=job.id,
+                sessionId=session["sessionId"],
+            ).get("sessionActions")
+
+            for session_action in session_actions:
+
+                # Session action should be failed IFF it's the expected action to fail
```
leongdl, line 212: IFF ?

YutongLi291: If and only if?

benl-2023 (approved):

crowecawcaw (approved):

leongdl (approved):

PR #365, opened 2024-07-26
test: Add test for CANCELED session actions and make FAILED action test cross OS

YutongLi291:
We have a core feature of worker agent in which the session actions should be CANCELED properly, when a job is set to CANCELED for its taskRunStatus.
We should keep this part of the code quality high for the future and prevent and possible regressions through dependencies or code.
Add a test that tests that each of `envEnter`, `taskRun`, and `envExit` are CANCELED properly when job is updated to be CANCELED taskRunStatus.

Also refactored an existing test so that it runs on both Linux and Windows. No logical change.
Better testing and code quality for code that CANCELs session actions
`hatch run linux-e2e-test`
`hatch run  cross-os-e2e-test ` with windows and linux fleets specified.
No
No
----


test/e2e/cross_os/test_job_submissions.py
```diff
@@ -0,0 +1,281 @@
...
+        def is_job_started(current_job: Job) -> bool:
+            current_job.refresh_job_info(client=deadline_client)
+            LOG.info(f"Waiting for job {current_job.id} to be created")
+            return current_job.lifecycle_status != "CREATE_IN_PROGRESS"
+
+        retries_for_job_start: int = 10
+        for i in range(retries_for_job_start):
+            if is_job_started(job):
+                break
+            sleep(10)
```
benl-2023: This could be simplified if we wrapped the logic to call `is_job_started` in an inner function and used the `backoff` libraries annotations to manage the retrying logic

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -0,0 +1,281 @@
...
+
+        retries_for_session_start = 10
+        for i in range(retries_for_session_start):
+            sessions = deadline_client.list_sessions(
+                farmId=job.farm.id, queueId=job.queue.id, jobId=job.id
+            ).get("sessions")
+
+            if len(sessions) > 0:
+                break
+            sleep(10)
```
benl-2023: Same comment as above

benl-2023 (approved):

PR #368, opened 2024-07-31
test: Add test to verify that worker and session logs are streamed to Cloudwatch

YutongLi291:
When jobs are ran, worker agent streams the worker and session logs to Cloudwatch logs, so that users can debug any issues or logs with the worker and/or session.

We should make sure that there are no regressions regarding this part of the worker functionality, as it is essential for user to be able to debug their jobs.
Add an E2E test that submits a job, and confirms that the corresponding worker and session logs exist in CloudWatch.
Upholding high quality code for the worker code that streams logs to cloudwatch.
`hatch run cross-os-test`
no
no
----

### DIFF
--- a/test/e2e/cross_os/test_job_submissions.py
+++ b/test/e2e/cross_os/test_job_submissions.py
@@ -6,13 +6,15 @@ Deadline Cloud service and checking that the result/output of the jobs is as we 
 from typing import Any, Dict, List
 import pytest
 import logging
-from deadline_test_fixtures import Job, DeadlineClient, TaskStatus
+from deadline_test_fixtures import Job, DeadlineClient, TaskStatus, EC2InstanceWorker
+from e2e.conftest import DeadlineResources
 from utils import get_operating_system_name
 import backoff
 import boto3
 import botocore.client
 import botocore.config
 import botocore.exceptions
+import time
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


AlexTranAmz (dismissed):

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -370,3 +372,57 @@ def test_worker_run_with_number_of_environments(
...
+            full_log.count("HelloWorld") == 1
+        ), "Expected number of HelloWorld statements not found in job logs."
+
+        # Retrieve worker logs and verify that it's not empty
+        worker_log_group_name: str = (
+            f"/aws/deadline/{deadline_resources.farm.id}/{deadline_resources.fleet.id}"
+        )
+        worker_id = worker.worker_id
+
+        worker_logs = logs_client.get_log_events(
```
leongdl, line 426: Since this is a recycled log stream from a persisted resource, is this 100% correct?

Should we verify the logs exist AFTER the jobs are created?

YutongLi291: You're right. I missed that part. I'll change it so that we verify logs exist after the job is created

leongdl (approved):

AlexTranAmz (approved):

PR #370, opened 2024-08-01
test: Refactor conftest to make worker run with session and function scope

AlexTranAmz:
We create an EC2 as a worker for the entire test fixture which limit some test cases where we need to terminate/cancel worker
Refactor and scope down so we have worker fixture for session and function
We can now have tests that could tear down or cancel worker
`hatch run cross-os-e2e-test`
`hatch run linux-e2e-test`
`hatch run windows-e2e-test`
Yes
No

----


test/e2e/conftest.py
```diff
@@ -213,78 +213,97 @@ def worker(
...
+            stop_worker(request, worker)
             raise
 
+        all_created_workers.append(worker)
+        return worker
+
+    yield generate_worker
+
+    for worker in all_created_workers:
+        stop_worker(request, worker)
```
YutongLi291: This doesn't seem too correct. the request would be consistent across the session. So if any tests fail across the session, we don't shut down any of the workers?

AlexTranAmz: Changed in the new commit to make sure function worker are killed once the tests finish rather then wait until the end and got deleted with the session worker.

test/e2e/conftest.py
```diff
@@ -213,78 +213,101 @@ def worker(
...
-            worker.stop()
+            worker.start()
         except Exception as e:
-            LOG.exception(f"Error while stopping worker: {e}")
-            LOG.error(
-                "Failed to stop worker. Resources may be left over that need to be cleaned up manually."
-            )
+            LOG.exception(f"Failed to start worker: {e}")
+            LOG.info("Stopping worker because it failed to start")
+            stop_worker(request, worker)
```
YutongLi291, line 315: Why are we still stopping the worker here using the `request` from this fixture? Should we not be using the `request` from the `function` and `session` fixtures?

AlexTranAmz: Fixed

YutongLi291 (dismissed):

test/e2e/conftest.py
```diff
@@ -213,78 +212,101 @@ def worker(
...
+    request: pytest.FixtureRequest, worker: Callable[[pytest.FixtureRequest], DeadlineWorker]
+) -> Generator[DeadlineWorker, None, None]:
+    session_worker: DeadlineWorker = worker(request)
+    yield session_worker
+
+    stop_worker(request, session_worker)
+
+
+@pytest.fixture(scope="function")
+def function_worker(
```
jericht: This could be an issue with job submissions. The test fixtures are currently assuming that there is only one worker deployed at a time. If a test suite uses both `session_worker` and `function_worker`, it's possible that there can be two workers running at the same time, so tests that rely on job submissions could fail because the job went to the wrong worker, unless each worker has its own dedicated queue.

Have we tested that this does not happen? or made sure that it does not affect our tests?

AlexTranAmz: I've test and make sure test that use mixed of session and function work at the same time. All the test are e2e and have job submission

jericht: Discussed offline, we will defer making changes to these fixtures so that job submission tests go to the worker they are meant to go to, rather than another one running at the same time, which can happen with both the session and function workers.

jericht (dismissed):

YutongLi291 (approved):

jericht (approved):

PR #373, opened 2024-08-07
test: Add test to verify WA job attachment settings

YutongLi291:
We want to verify that worker agent  is able to follow queues job attachment settings and correctly identify input/output directories to take actions on files.

We should add testing regarding this functionality.
Add a test that verifies that WA is able to execute a job that respects the job attachment settings, and verifies that the job output is as we expected.
Better verification on the functionality to use job attachments with jobs, as well as with Worker Agent.
hatch run linux-e2e-tests
no
no
----


test/e2e/linux/test_job_submissions.py
```diff
@@ -43,7 +52,7 @@ def test_success(
 
         # THEN
         LOG.info(f"Waiting for job {job.id} to complete")
-        job.wait_until_complete(client=deadline_client)
+        job.wait_until_complete(client=deadline_client, max_retries=20)
```
ddneilson, line 56: Would you mind looking into adding a non-`None` default to the [`wait_until_complete` implementation](https://github.com/aws-deadline/deadline-cloud-test-fixtures/blob/1f27578da0f2fe9cc241334acd439d9e1d741a1d/src/deadline_test_fixtures/deadline/resources.py#L577-L583)? The default, right now, is infinite retries forever. That's dangerous in that an unexpected malfunction would mean that automated tests will literally run forever, or until we take notice and nuke the test runner. It can be a really big default, but there should still be a default, imo.

test/e2e/linux/test_job_submissions.py
```diff
@@ -109,3 +118,95 @@ def test_job_run_as_user(
...
+
+        job_bundle_path: str = os.path.join(
+            os.path.dirname(__file__),
+            "..",
+            "..",
+            "..",
+            "scripts",
+            "submit_jobs",
+            "job_attachment_job",
+        )
```
ddneilson: I'd suggest putting the test job beside this file, rather than back up in the scripts directory. Reason being that the scripts directory is intended to be a fairly random collection of useful scripts and files for manual testing. Putting a part of the e2e tests in there risks someone changing the bundle in some way that's unexpected by the test. Yes, the test should catch that, but it's a better idea to have the testing files all within the testing directories, imo.

test/e2e/linux/test_job_submissions.py
```diff
@@ -109,3 +118,95 @@ def test_job_run_as_user(
...
+            open(
+                os.path.join(job_bundle_path, "outputs", test_run_uuid, "output_file.txt"), "r"
+            ) as output_file,
+        ):
+            input_file_content: str = input_file.read()
+            output_file_content = output_file.read()
+
+            assert output_file_content == (input_file_content + test_run_uuid)
+
+        if os.path.exists(
```
ddneilson: Rather than this, I'd suggest downloading the output files into [a temporary directory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). Reason being that a failed test will leave files behind right now -- e.g. if the `output_file.txt` doesn't get generated, or downloaded, for some reason then the `open` call above will raise an exception and abort the test.

If you use a `tempfile.TemporaryDirectory` as a context manager, instead, then the tempdir will get deleted regardless of whether the test exists successfully or not.

Alternatively to that, there's also the `tmp_path` fixture built-in to pytest for getting a temporary directory for the duration of a test: https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture

ddneilson (changes requested):

test/e2e/linux/test_job_submissions.py
YutongLi291: Agreed, I'll add it to the implementation.

YutongLi291: https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/134

test/e2e/linux/test_job_submissions.py
YutongLi291: I've changed it to use a temp directory in the revision. Thanks for pointing this out

test/e2e/linux/test_job_submissions.py
YutongLi291: Done, I've put the job bundle under the tests path

ddneilson (approved):
Thanks for adding this test!

AlexTranAmz (approved):

PR #375, opened 2024-08-08
test:Add test to verify install give shutdown permission to WA user

AlexTranAmz:
WA missing some tests to verify its functionality.

In this case, we missing test to verify if the installer give shutdown permission to WA user
Add test case to verify if the installer give shutdown permission to WA user
We cover more test cases and reduce test gap

`hatch run cross-os-e2e-test`

Yes

No

----

### DIFF
--- a/test/e2e/cross_os/test_installer.py
+++ b/test/e2e/cross_os/test_installer.py
@@ -1,0 +1,27 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""
+This test module contains tests that verify the Installer's behavior by submitting commands to the
+Deadline Cloud worker and checking that the result/output of the worker agent is as we expect it.
+"""
+import pytest
+import logging
+from deadline_test_fixtures import EC2InstanceWorker
+from utils import get_operating_system_name
+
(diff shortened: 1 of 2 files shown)


YutongLi291 (dismissed):

test/e2e/cross_os/test_installer.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+import pytest
+import logging
+from deadline_test_fixtures import EC2InstanceWorker
+from utils import get_operating_system_name
+
+LOG = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize("operating_system", [get_operating_system_name()], indirect=True)
```
AWS-Samuel: If I'm reading this right, it looks like we change which operating system the worker is to be the same operating system of the machine we're submitting from. 

Looking at the test case, it looks like the test only works for linux. 

Have you tested submitting this to a windows worker?

AlexTranAmz: Changed to only run on Linux

AWS-Samuel (approved):

YutongLi291 (approved):

PR #376, opened 2024-08-08
test: Add test that worker status is what we expect

YutongLi291:
The worker agent has statuses that are exposed to users of Deadline Cloud. We should make sure that  the worker status returned by the API corresponds to the status that we expect the worker to be in, when we update the worker to the corresponding status.

We should make sure that code quality for this logic is kept high, so as to ensure a good customer experience in regards to the worker.
Add testing regarding the worker.

Also moved some files so that another test for a test job with job Attachments is ran on both windows and Linux

Better code quality
`hatch run cross-os-e2e-tests`, `hatch run linux-e2e-tests`
no
no
----


test/e2e/cross_os/test_job_submissions.py
```diff
@@ -430,3 +439,83 @@ def test_worker_streams_logs_to_cloudwatch(
         )
 
         assert len(worker_logs["events"]) > 0
+
+    def test_worker_uses_job_attachment_configuration(
```
YutongLi291, line 443: This is moved from the linux test file, so that it will run on both linux and windows

AlexTranAmz (approved):

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -430,3 +439,83 @@ def test_worker_streams_logs_to_cloudwatch(
...
+        job_id: Optional[str] = api.create_job_from_job_bundle(
+            job_bundle_path,
+            job_parameters,
+            priority=99,
+            config=config,
+            queue_parameter_definitions=[],
+        )
+        assert job_id is not None
+
+        job_details = Job.get_job_details(
```
jairaws, line 474: nit: type

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -430,3 +439,83 @@ def test_worker_streams_logs_to_cloudwatch(
...
+        )
+        job = Job(
+            farm=deadline_resources.farm,
+            queue=deadline_resources.queue_a,
+            template={},
+            **job_details,
+        )
+        job.wait_until_complete(client=deadline_client, max_retries=20)
+
+        job_attachment_settings = get_queue(
```
jairaws, line 488: nit: type

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -430,3 +439,83 @@ def test_worker_streams_logs_to_cloudwatch(
...
+            farm_id=deadline_resources.farm.id,
+            queue_id=deadline_resources.queue_a.id,
+        ).jobAttachmentSettings
+
+        assert job_attachment_settings is not None
+
+        job_output_downloader = download.OutputDownloader(
```
jairaws, line 495: nit: type

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -430,3 +439,83 @@ def test_worker_streams_logs_to_cloudwatch(
...
+            s3_settings=job_attachment_settings,
+            farm_id=deadline_resources.farm.id,
+            queue_id=deadline_resources.queue_a.id,
+            job_id=job.id,
+            step_id=None,
+            task_id=None,
+        )
+
+        output_paths_by_root = job_output_downloader.get_output_paths_by_root()
```
jairaws, line 504: nit: type

test/e2e/cross_os/test_worker_status.py
```diff
@@ -0,0 +1,72 @@
...
+from deadline_worker_agent.api_models import (
+    WorkerStatus,
+)
+import backoff
+from deadline_test_fixtures import DeadlineClient, EC2InstanceWorker
+from utils import get_operating_system_name
+import pytest
+
+
+@pytest.mark.parametrize("operating_system", [get_operating_system_name()], indirect=True)
```
jairaws, line 17: are we planning on adding more parameters later on? Also, I'm not familiar with using parameters at test class level. What's the purpose of this? Could it be a fixture instead?

jairaws (approved):

PR #379, opened 2024-08-11
test: put worker job attachment test template in test file and add OPERATING_SYSTEM env variable

YutongLi291:
We should put the test template for the job attachment test in the test file itself, as there is no need for it to live inside a folder.

Furthermore, we need to run the test on windows, so we need to create a windows script so the test works properly on windows fleets. 

Move the template to the test class that creates  the temporary template, based off whether the OS is Windows or Linux that we want to test.

Also added an operating system variable so we can easier specify the OS we want to run the tests on

Better file readability in the package file structure.
 
Windows OS test for job attachments test should be running properly after. 

Easier to specify OS we want to run the tests on 
`hatch run cross-os-e2e-test` 
No
No
----

### DIFF
--- a/test/e2e/cross_os/test_job_submissions.py
+++ b/test/e2e/cross_os/test_job_submissions.py
@@ -3,12 +3,12 @@
 This test module contains tests that verify the Worker agent's behavior by submitting jobs to the
 Deadline Cloud service and checking that the result/output of the jobs is as we expect it.
 """
+import json
 from typing import Any, Dict, List, Optional
 import pytest
 import logging
 from deadline_test_fixtures import Job, DeadlineClient, TaskStatus, EC2InstanceWorker
 from e2e.conftest import DeadlineResources
-from utils import get_operating_system_name
 import backoff
 import boto3
 import botocore.client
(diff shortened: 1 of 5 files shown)


AlexTranAmz (dismissed):

AlexTranAmz (approved):

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -457,19 +468,63 @@ def test_worker_uses_job_attachment_configuration(
...
+                                }
+                            ],
+                        }
+                    )
+                )
+
+            config = configparser.ConfigParser()
+
+            set_setting("defaults.farm_id", deadline_resources.farm.id, config)
+            set_setting("defaults.queue_id", deadline_resources.queue_a.id, config)
```
godobyte, line 515: Curious - How is the infra setup done for windows and linux separation, are linux and windows jobs submitted to different queues? AFAIR queue for linux worker has conda environment which doesn't work with queue for windows worker.

godobyte: Synced offline for the test setup. TLDR - windows and linux jobs are submitted to different queues.

godobyte (approved):

PR #387, opened 2024-08-14
test: change test worker instance type to t3.medium

YutongLi291:
The current worker instance type that the tests launch with is t3.micro, which is very small and makes the worker tests run quite slowly.

* Change the instance type for the workers to be t3.medium. 

* Also added some minor documentation strings, as well as cleaned up an unused template file. this template is now created in [this](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/85d032504a58e63bdf21579e81c39b8f38b4a089/test/e2e/cross_os/test_job_submissions.py#L469) file
E2E tests should run faster, due to having a faster worker instance.
Tested by `hatch build` on my local, and running both the linux and windows tests using `hatch run cross-os-e2e-test` , `hatch run linux-e2e-test`, and `hatch run windows-e2e-test` with `OPERATING_SYSTEM` set accordingly
no
no
----


scripts/deploy_e2e_testing_infrastructure.sh
```diff
@@ -16,6 +16,7 @@ fi
 
 if ! aws cloudformation describe-stacks --stack-name DeadlineCloudAgentE2EInfrastructure 2>&1 > /dev/null; then
   OP=create-stack
+  echo "Creating the DeadlineCloudAgentE2EInfrastructure stack as it does not already exist"
```
YutongLi291, line 19: This is because currently it prints an error when not finding the stack, which may lead to thinking that the script had an error.

test/e2e/conftest.py
```diff
@@ -278,7 +278,7 @@ def __exit__(self, exc_type, exc_value, traceback):
         ami_id = os.getenv("AMI_ID")
         subnet_id = os.getenv("SUBNET_ID")
         security_group_id = os.getenv("SECURITY_GROUP_ID")
-        instance_type = os.getenv("WORKER_INSTANCE_TYPE", default="t3.micro")
```
AWS-Samuel, line 281: What was the performance increase with this change?

test/e2e/cross_os/job_attachment_bundle/template.json
```diff
@@ -1,40 +0,0 @@
...
-                        "name": "runScript",
-                        "type": "TEXT",
-                        "runnable": true,
-                        "data": "#!/usr/bin/env bash\n\n  echo -n $(cat {{Param.DataDir}}/files/test_input_file){{Param.StringToAppend}} > {{Param.DataDir}}/output_file.txt\n"
-                    }
-                ]
-            }
-        }
-    ]
-}
```
jusiskin, line 40: This file is referenced here:

https://github.com/YutongLi291/deadline-cloud-worker-agent-fork/blob/a01275aedf44f84d700ee57d29e6c72e3390b531/test/e2e/cross_os/test_job_submissions.py#L479-L482

Was this deleted unintentionally?

jusiskin (dismissed):

test/e2e/cross_os/job_attachment_bundle/template.json
YutongLi291: That's the job bundle location, which includes the input file. however we create the template file at runtime https://github.com/YutongLi291/deadline-cloud-worker-agent-fork/blob/a01275aedf44f84d700ee57d29e6c72e3390b531/test/e2e/cross_os/test_job_submissions.py#L488 here, based on whether it's windows or linux.

AlexTranAmz (approved):

test/e2e/conftest.py
YutongLi291: It used to take around 15 minutes on the Codebuild to run the cross-os-e2e-test on linux, but after this change I found that it takes around 10 minutes on my local.

leongdl (approved):

PR #390, opened 2024-08-15
test: Add test to verify JA sync progress

AlexTranAmz:
We do not have e2e test to verify JA sync attachment progress to see whether the progress percentage is correct or not

Add a e2e test that sync multiple file and assert the progress percentage to see if it's increase as expected
More e2e coverage
`hatch run cross-os-e2e-test`
Yes
No

----

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -12,15 +12,11 @@ import tempfile
 from dataclasses import dataclass, field, InitVar
 from typing import Generator, Type
 from contextlib import contextmanager
-from deadline.job_attachments._aws.deadline import get_queue
-from deadline.job_attachments import download
 
 from deadline_test_fixtures import (
     DeadlineWorker,
     DeadlineWorkerConfiguration,
     DockerContainerWorker,
-    DeadlineClient,
-    Job,
     Farm,
     Fleet,
     Queue,
@@ -373,29 +369,3 @@ def operating_system(request) -> OperatingSystem:
         return OperatingSystem(name="AL2023")
     else:
         return OperatingSystem(name="WIN2022")
-
-
-def get_job_output(
-    job: Job, deadline_client: DeadlineClient, deadline_resources: DeadlineResources
-) -> dict[str, list[str]]:
-    job.wait_until_complete(client=deadline_client, max_retries=20)
-
-    job_attachment_settings = get_queue(
-        farm_id=deadline_resources.farm.id,
-        queue_id=deadline_resources.queue_a.id,
-    ).jobAttachmentSettings
-
-    assert job_attachment_settings is not None
-
-    job_output_downloader = download.OutputDownloader(
-        s3_settings=job_attachment_settings,
-        farm_id=deadline_resources.farm.id,
-        queue_id=deadline_resources.queue_a.id,
-        job_id=job.id,
-        step_id=None,
-        task_id=None,
-    )
-    output_paths_by_root = job_output_downloader.get_output_paths_by_root()
-    job_output_downloader.download_job_output()
-
-    return output_paths_by_root
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -12,11 +12,15 @@ import tempfile
 from dataclasses import dataclass, field, InitVar
 from typing import Generator, Type
 from contextlib import contextmanager
+from deadline.job_attachments._aws.deadline import get_queue
+from deadline.job_attachments import download
 
 from deadline_test_fixtures import (
     DeadlineWorker,
     DeadlineWorkerConfiguration,
     DockerContainerWorker,
+    DeadlineClient,
+    Job,
     Farm,
     Fleet,
     Queue,
@@ -369,3 +373,29 @@ def operating_system(request) -> OperatingSystem:
         return OperatingSystem(name="AL2023")
     else:
         return OperatingSystem(name="WIN2022")
+
+
+def get_job_output(
+    job: Job, deadline_client: DeadlineClient, deadline_resources: DeadlineResources
+) -> dict[str, list[str]]:
+    job.wait_until_complete(client=deadline_client, max_retries=20)
+
+    job_attachment_settings = get_queue(
+        farm_id=deadline_resources.farm.id,
+        queue_id=deadline_resources.queue_a.id,
+    ).jobAttachmentSettings
+
+    assert job_attachment_settings is not None
+
+    job_output_downloader = download.OutputDownloader(
+        s3_settings=job_attachment_settings,
+        farm_id=deadline_resources.farm.id,
+        queue_id=deadline_resources.queue_a.id,
+        job_id=job.id,
+        step_id=None,
+        task_id=None,
+    )
+    output_paths_by_root = job_output_downloader.get_output_paths_by_root()
+    job_output_downloader.download_job_output()
+
+    return output_paths_by_root
--- a/test/e2e/cross_os/test_job_submissions.py
+++ b/test/e2e/cross_os/test_job_submissions.py
@@ -3,7 +3,9 @@
 This test module contains tests that verify the Worker agent's behavior by submitting jobs to the
 Deadline Cloud service and checking that the result/output of the jobs is as we expect it.
 """
+import hashlib
 import json
+import shutil
 from typing import Any, Dict, List, Optional
 import pytest
 import logging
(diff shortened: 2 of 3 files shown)


test/e2e/conftest.py
```diff
@@ -369,3 +373,29 @@ def operating_system(request) -> OperatingSystem:
         return OperatingSystem(name="AL2023")
     else:
         return OperatingSystem(name="WIN2022")
+
+
+def get_job_output(
```
YutongLi291: I don't think this function belongs in the conftest file, it belongs more in some util file.

AlexTranAmz: Moved in the new commit

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -632,47 +630,195 @@ def test_worker_uses_job_attachment_configuration(
...
+            os.path.dirname(__file__),
+            "job_attachment_bundle_large",
+        )
+        file_path: str = os.path.join(job_bundle_path, "files")
+
+        os.mkdir(job_bundle_path)
+        os.mkdir(file_path)
+
+        # Create 2500 very small files to transfer
+        for i in range(2500):
```
YutongLi291, line 709: Where are you deleting these files?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -632,47 +630,195 @@ def test_worker_uses_job_attachment_configuration(
...
+        job = Job(
+            farm=deadline_resources.farm,
+            queue=deadline_resources.queue_a,
+            template={},
+            **job_details,
+        )
+
+        # Query the session to check for progress percentage
+        complete_percentage: float = 0
+        while complete_percentage < 100:
```
YutongLi291: We should make sure this while loop doesn't run forever by adding some sort of max iterations count. Also I think we need to backoff somewhere to avoid this loop running too fast

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -632,47 +630,195 @@ def test_worker_uses_job_attachment_configuration(
...
+        )
+
+        with (
+            open(os.path.join(list(output_path.keys())[0], "output_file.txt"), "r") as output_file,
+        ):
+            output_file_content = output_file.read()
+            # Verify that the hash is the same
+            assert output_file_content == combined_hash
+
+        shutil.rmtree(job_bundle_path)
```
YutongLi291: This is bad, it will only run when the test is finished successfully. We want to cleanup the job bundle no matter what, so we should either use a try catch or temporary directories.

YutongLi291 (changes requested):

test/e2e/cross_os/test_job_submissions.py
YutongLi291: I see it at the bottom of the test, but that will not always run, if the test fails

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -632,47 +632,207 @@
...
+
+        # Query the session to check for progress percentage
+        complete_percentage: float = 0
+
+        @backoff.on_predicate(
+            wait_gen=backoff.constant,
+            max_time=30,
+            interval=2,
+        )
+        def check_percentage(complete_percentage):
```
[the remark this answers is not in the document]
YutongLi291: @AlexTranAmz Is this an issue?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -632,47 +632,207 @@ def test_worker_uses_job_attachment_configuration(
...
+
+                for session_action in session_actions:
+                    if "syncInputJobAttachments" in session_action["definition"]:
+                        assert complete_percentage <= session_action["progressPercent"]
+                        complete_percentage = session_action["progressPercent"]
+                        return complete_percentage == 100
+
+        assert check_percentage(complete_percentage)
+
+        shutil.rmtree(job_bundle_path)
```
YutongLi291: This is not safe, this code may never get ran due to a test failure before this line, and then the job bundle path will never get deleted. i think we should either use a temporary directory like https://docs.python.org/3/library/tempfile.html or use a `finally` clause to clean up the directory.

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -632,47 +632,207 @@ def test_worker_uses_job_attachment_configuration(
...
+
+        with (
+            open(os.path.join(list(output_path.keys())[0], "output_file.txt"), "r") as output_file,
+        ):
+            output_file_content = output_file.read()
+            # Verify that the hash is the same
+            assert output_file_content == combined_hash
+
+        # Clean up the output file
+        os.remove(os.path.join(list(output_path.keys())[0], "output_file.txt"))
```
YutongLi291: Again, this is not safe. We should be saving the output file into a temporary directory instead.

YutongLi291 (changes requested):

test/e2e/cross_os/test_job_submissions.py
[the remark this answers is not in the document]
AlexTranAmz: Nope from my testing

leongdl:
typo in your description `wether` -> `whether`

test/e2e/utils.py
```diff
@@ -0,0 +1,36 @@
...
+from deadline.job_attachments import download
+from deadline_test_fixtures import (
+    DeadlineClient,
+    Job,
+)
+
+from e2e.conftest import DeadlineResources
+
+
+def get_job_output(
```
leongdl: Can we add some function documentation, and also change the name? Right now it is not clear this function WAITS for the job to be completed!

test/e2e/utils.py
```diff
@@ -0,0 +1,36 @@
...
+    job_output_downloader = download.OutputDownloader(
+        s3_settings=job_attachment_settings,
+        farm_id=deadline_resources.farm.id,
+        queue_id=deadline_resources.queue_a.id,
+        job_id=job.id,
+        step_id=None,
+        task_id=None,
+    )
+    output_paths_by_root = job_output_downloader.get_output_paths_by_root()
+    job_output_downloader.download_job_output()
```
leongdl, line 35: This downloads the outputs but is not used anywhere? Maybe this line does not belong in the shared function?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -676,3 +654,192 @@ def test_worker_uses_job_attachment_configuration(
...
                 # Verify that the output file content is the input file content plus the uuid we appended in the job
                 assert output_file_content == (input_file_content + test_run_uuid)
+        finally:
+            os.remove(os.path.join(list(output_path.keys())[0], "output_file"))
+
+    @pytest.mark.parametrize(
+        "hash_string_script",
+        [
+            (
+                '#!/usr/bin/env bash\n\nfolder_path={{Param.DataDir}}/files\ncombined_contents=""\nfor file in "$folder_path"/*; do\n   if [ -f "$file" ]; then\n      combined_contents+="$(cat "$file" | tr -d \'\\n\')"\n   fi\ndone\nsha256_hash=$(echo -n "$combined_contents" | sha256sum | awk \'{ print $1 }\')\necho -n "$sha256_hash" > {{Param.DataDir}}/output_file.txt'
```
leongdl: Can we format this code as string :) Its very hard to understand.

test/e2e/cross_os/test_job_submissions.py
leongdl: Maybe better to have a fixture to create these files and cleanup?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -676,3 +654,192 @@ def test_worker_uses_job_attachment_configuration(
...
+                json.dumps(
+                    {
+                        "specificationVersion": "jobtemplate-2023-09",
+                        "name": "AssetsSync",
+                        "parameterDefinitions": [
+                            {
+                                "name": "DataDir",
+                                "type": "PATH",
+                                "dataFlow": "INOUT",
+                                "userInterface": {
```
leongdl, line 742: nit, not useful.

test/e2e/utils.py
AlexTranAmz: It download the file to the path and we pass the path back to the function to read the content and assert

test/e2e/cross_os/test_job_submissions.py
AlexTranAmz: I believe we need this to pass in the folder path to sync job attachment?

test/e2e/cross_os/test_job_submissions.py
AlexTranAmz: Since these file is changed depend on each test req, I don't think having it in a fixture is needed.

ddneilson: I'd suggest using this: https://docs.pytest.org/en/stable/how-to/tmp_path.html

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -678,3 +656,203 @@ def test_worker_uses_job_attachment_configuration(
...
+            else:
+                return False
+
+        assert check_percentage(complete_percentage)
+
+        output_path: dict[str, list[str]] = wait_for_job_output(
+            job=job, deadline_client=deadline_client, deadline_resources=deadline_resources
+        )
+        with (
+            open(os.path.join(list(output_path.keys())[0], "output_file.txt"), "r") as output_file,
```
YutongLi291, line 854: Where is the output file deleted?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -678,3 +656,203 @@ def test_worker_uses_job_attachment_configuration(
...
+        # Verify that the worker sync job attachment correctly and report the progress correctly as well
+
+        job_bundle_path: str = os.path.join(
+            tmp_path,
+            "job_attachment_bundle_large",
+        )
+        file_path: str = os.path.join(job_bundle_path, "files")
+
+        os.mkdir(job_bundle_path)
+        os.mkdir(file_path)
```
YutongLi291, line 706: Where are these directories deleted?

YutongLi291 (changes requested):

test/e2e/cross_os/test_job_submissions.py
YutongLi291: Nvm, you are using the pytest tmp_path fixture

test/e2e/cross_os/test_job_submissions.py
YutongLi291: Nvm, you are using the pytest tmp_path fixture

YutongLi291 (approved):

test/e2e/cross_os/test_job_submissions.py
ddneilson: The `userInterface` property is only useful if using a GUI to submit the job. Job attachments only uses the `dataFlow` property of a parameter.

leongdl: Yeah, what Daniel's comment suggests, we can delete the user interface block.

leongdl (approved):

PR #391, opened 2024-08-16
test: all 0 retries for test that purposely failed/canceled

AlexTranAmz:
Test that purposely failed/canceled job will be retries up to 5 times per the default config
Make sure when submitting job, `max_retries` is set at 0
Job that purposely failed will not retries to save resource
`hatch run cross-os-e2e-test`
No
No

----

### DIFF
--- a/test/e2e/cross_os/test_job_submissions.py
+++ b/test/e2e/cross_os/test_job_submissions.py
@@ -92,6 +92,7 @@ class TestJobSubmission:
             farm=deadline_resources.farm,
             queue=deadline_resources.queue_a,
             priority=98,
+            max_retries_per_task=0,
             template={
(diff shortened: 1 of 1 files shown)


test/e2e/cross_os/test_job_submissions.py
```diff
@@ -612,6 +614,7 @@ def test_worker_uses_job_attachment_configuration(
                 job_bundle_path,
                 job_parameters,
                 priority=99,
+                max_retries_per_task=0,
```
YutongLi291: We aren't trying to fail this job, so having retries is preferable

leongdl (approved):
Can you post the test run results?

YutongLi291 (approved):

PR #392, opened 2024-08-16
test: refactor test for worker status to use ssm command

YutongLi291:
The existing test to test that worker status is correctly updated uses the deadline API, and not directly interacting with the worker.

We want to make sure that when the worker is STOPPED correctly when the worker service is shut down.

Change the test to shut down the worker using an SSM command to shut down the service.

Also fixed an issue with the default deadline worker configuration not automatically starting the worker service. 
Better testing for the worker service in regards to the worker status presented to users.
Using `DEVELOPMENT.md`

# Linux
source .e2e_linux_infra.sh
hatch run linux-e2e-test

hatch run cross-os-e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run windows-e2e-test

hatch run cross-os-e2e-test
No
No
----

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -196,6 +196,7 @@ def worker_config(
             service_model_path=dst_path,
             file_mappings=file_mappings or None,
             windows_job_users=windows_job_users,
+            start_service=True,
         )
 
 
--- a/test/e2e/cross_os/test_worker_status.py
+++ b/test/e2e/cross_os/test_worker_status.py
@@ -3,15 +3,15 @@
 This test module contains tests that verify the Worker agent's behavior by starting/stopping the Worker,
 and making sure that the status of the Worker is that of what we expect.
 """
+import logging
 import os
 import pytest
 from typing import Any, Dict
-from deadline_worker_agent.api_models import (
-    WorkerStatus,
-)
 import backoff
 from deadline_test_fixtures import DeadlineClient, EC2InstanceWorker
 import pytest
+
+LOG = logging.getLogger(__name__)
 
 
 @pytest.mark.parametrize("operating_system", [os.environ["OPERATING_SYSTEM"]], indirect=True)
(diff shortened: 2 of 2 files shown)


test/e2e/conftest.py
```diff
@@ -196,6 +196,7 @@ def worker_config(
             service_model_path=dst_path,
             file_mappings=file_mappings or None,
             windows_job_users=windows_job_users,
+            start_service=True,
```
ddneilson, line 199: The PR description indicates that you only ran the cross-os tests, but this change will impact the os-specific tests as well. Can you please run those and verify that the change does not impact them negatively?

test/e2e/cross_os/test_worker_status.py
```diff
@@ -47,19 +47,19 @@ def is_worker_started() -> bool:
 
         assert is_worker_started()
 
-        deadline_client.update_worker(
-            farmId=deadline_resources.farm.id,
-            fleetId=deadline_resources.fleet.id,
-            workerId=function_worker.worker_id,
-            status=WorkerStatus.STOPPED,
+        function_worker.send_command(
```
ddneilson: There's a helper function for this. You shouldn't need to craft the ssm commands directly.

[Windows](https://github.com/aws-deadline/deadline-cloud-test-fixtures/blob/afbc5fb88f981976d60635cfb84d409eda65ffb8/src/deadline_test_fixtures/deadline/worker.py#L489-L493) & [Linux](https://github.com/aws-deadline/deadline-cloud-test-fixtures/blob/afbc5fb88f981976d60635cfb84d409eda65ffb8/src/deadline_test_fixtures/deadline/worker.py#L691-L695)

Please use those instead. That way the test doesn't need updating if there are ever changes to how we run the worker agent.

ddneilson (changes requested):

test/e2e/cross_os/test_worker_status.py
YutongLi291: Refactored and published the change!

test/e2e/conftest.py
YutongLi291: Just ran them, they passed so it did not impact them

AlexTranAmz (approved):

ddneilson (approved):

PR #393, opened 2024-08-20
chore: add custom attribute to test and cloud formation template

AlexTranAmz:
We want to run e2e test in parallel to save time, especially when spinning up a new worker. 
This however is not easy with current setup and could lead to race condition
Add custom capabilities for each test to make sure test do not grab worker that's not in scope.
Also make some change to the cloud formation template to reflect the newly added custom capabilities.
We can now research on running test in parallel
`hatch run cross-os-e2e-test`
`hatch run windows-e2e-test`
`hatch run linux-e2e-test`
All using the cloud formation template deployment
Yes
No

----


test/e2e/conftest.py
```diff
@@ -170,43 +225,43 @@
         LOG.info(f"Using Worker agent package {worker_agent_requirement_specifier}")
 
     # Path map the service model
-    with tempfile.TemporaryDirectory() as tmpdir:
-        src_path = pathlib.Path(tmpdir) / f"{service_model.service_name}-service-2.json"
+    tmpdir = tmpdir
```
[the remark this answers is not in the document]
AlexTranAmz: fixed

YutongLi291 (approved):

test/e2e/conftest.py
```diff
@@ -215,10 +268,10 @@ def session_worker(
 @pytest.fixture(scope="class")
 def class_worker(
     request: pytest.FixtureRequest,
-    worker_config: DeadlineWorkerConfiguration,
+    worker_config_session: DeadlineWorkerConfiguration,
     ec2_worker_type: Type[EC2InstanceWorker],
 ) -> Generator[DeadlineWorker, None, None]:
-    with create_worker(worker_config, ec2_worker_type, request) as worker:
+    with create_worker(worker_config_session, ec2_worker_type, request) as worker:
         yield worker
```
jericht, line 275: We have a class-scoped worker fixture? Is this actually being used anywhere? If so, we'll need a fleet for this as well

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -103,7 +102,11 @@ def test_job_reports_failed_session_action(
                                 {
                                     "name": "attr.worker.os.family",
                                     "allOf": [os.environ["OPERATING_SYSTEM"]],
-                                }
+                                },
+                                {
+                                    "name": "attr.TestFleetType",
+                                    "anyOf": ["FleetforSession"],
+                                },
```
jericht, line 110: Instead of hardcoding the attribute into the job templates everywhere we need it, how about placing this in a common function/variable and just referencing that instead?

PR #394, opened 2024-08-21
test: add e2e tests for worker agent cancel action and environment exit

YutongLi291:
Worker agent is the core functionality of processing and interacting with a job, including canceling job actions while they are underway, as well as running job environments.

We should make sure that these components of worker agent are thoroughly tested, to ensure high code quality standards in this package.
Add tests that test these functionalities, including
* Test that environment exit on an environment that has been entered, is always ran, regardless of the result of the job, and whether it's been canceled or failed
* Test that canceled session actions, which have a trap to intercept INT signals, are almost immediately canceled and move to `CANCELED` status
* Test that when syncing job attachment inputs, if the job is canceled, then the job attachment input sync is canceled accordingly

Better test coverage of the worker agent code, to ensure better Worker Agent code quality, especially regarding canceling jobs while they are running, as well as relating to environment exiting properly.
```
source .e2e_linux_infra.sh
hatch run linux-e2e-test
hatch run cross-os-e2e-test


source .e2e_windows_infra.sh
hatch run windows-e2e-test
hatch run cross-os-e2e-test

```

All the tests passed
no
no
----

### DIFF
--- a/test/e2e/cross_os/test_job_submissions.py
+++ b/test/e2e/cross_os/test_job_submissions.py
@@ -4,6 +4,7 @@ This test module contains tests that verify the Worker agent's behavior by submi
 Deadline Cloud service and checking that the result/output of the jobs is as we expect it.
 """
 import hashlib
+from flaky import flaky
 import json
 import pathlib
 from typing import Any, Dict, List, Optional
@@ -211,7 +212,6 @@ class TestJobSubmission:
             farm=deadline_resources.farm,
             queue=deadline_resources.queue_a,
             priority=98,
-            max_retries_per_task=0,
             template={
                 "specificationVersion": "jobtemplate-2023-09",
                 "name": f"jobactioncancel-{expected_canceled_action}",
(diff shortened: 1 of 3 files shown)


AlexTranAmz (dismissed):

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -312,6 +313,498 @@ def is_expected_session_action_canceled(sessions: List[Dict[str, Any]]) -> bool:
...
+                    }
+                )
+            )
+        # Create the input files to make sync inputs take a relatively long time
+        files_path = os.path.join(tmp_path, "files")
+        os.mkdir(files_path)
+        for i in range(2000):
+            file_name: str = os.path.join(files_path, f"input_file_{i+1}.txt")
+            with open(file_name, "w+") as input_file:
+                input_file.write(f"{i}")
```
benl-2023, line 605: >  Flaky as sync input sometimes completes before expected.

Would randomizing and creating a little more data in these input files help with the flakiness of the test?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -312,6 +313,498 @@ def is_expected_session_action_canceled(sessions: List[Dict[str, Any]]) -> bool:
...
+                ).get("sessionActions")
+                logging.info(f"Session actions: {session_actions}")
+                for session_action in session_actions:
+                    # Session action should be canceled if it's the action we expect to be canceled
+                    if "syncInputJobAttachments" in session_action["definition"]:
+                        if session_action["status"] == "CANCELED":
+                            found_canceled_sync_input_action = True
+                    else:
+                        assert (
+                            session_action["status"] != "CANCELED"
```
benl-2023: Given the comment below, should the assertion here then be:

```
session_action["status"] == "SUCCEEDED" or session_action["status"] == "NEVER_ATTEMPTED" 
```

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -312,6 +313,498 @@ def is_expected_session_action_canceled(sessions: List[Dict[str, Any]]) -> bool:
...
+                jobId=job.id,
+                sessionId=session["sessionId"],
+            ).get("sessionActions")
+            logging.info(f"Session actions: {session_actions}")
+            for session_action in session_actions:
+                definition = session_action["definition"]
+                if "envEnter" in definition:
+                    if successful_environment_name in definition["envEnter"]["environmentId"]:
+                        assert session_action["status"] == "SUCCEEDED"
+                        found_successful_env_enter = True
```
benl-2023: nit, since we're asserting the intersection of all of these values, I think we can set these "found_(un)successful..." variables to the expression.

ie, `found_successful_env_enter = session_action["status"] == "SUCCEEDED"`

test/e2e/cross_os/test_job_submissions.py
YutongLi291: Doesn't seem like it. I've tried a few approaches and none of them seem to help. If I make the input files even bigger, it also times out, likely due to the sheer number. It's also the case that the sync input never uses the already uploaded attachments, if that's what you meant by randomizing.

test/e2e/cross_os/test_job_submissions.py
YutongLi291: I guess the issue here is that I'm not waiting for the job to be completed, which means that other actions could be in some other state, such as RUNNING.

I think it makes sense that I wait for the job to be completed

test/e2e/cross_os/test_job_submissions.py
YutongLi291: Makes sense, will revise

THIRD-PARTY-LICENSES.txt
```diff
@@ -1704,4 +1704,219 @@ ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
 LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
 ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
-SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
\ No newline at end of file
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+
+** flaky; version 3.8 -- https://pypi.org/project/flaky/
```
ddneilson: The third party license file is only for direct dependencies of the release product. We don't need entries for the packages that are only used in tests.

YutongLi291: Ah, ok

AlexTranAmz (dismissed):

THIRD-PARTY-LICENSES.txt
YutongLi291: Fixed in second revision

AlexTranAmz (dismissed):

benl-2023 (approved):

AlexTranAmz (approved):

PR #397, opened 2024-08-23
fix: allow retries to tests

AlexTranAmz:
Test start to fail after strictly do not allow retries. Need to enable it back to work
Remove `max_retries_per_task` when create job
Tests will now have more retries to pass
`hatch run windows-e2e-test` 
`hatch run linux-e2e-test`
`hatch run cross-os-e2e-test`
No
No

----


YutongLi291 (approved):

AWS-Samuel (approved):

AWS-Samuel:
Nit, we should title test changes with `test: <commit message>` or else they show up in the changelog when we release a new version. Not a big deal so merging anyway to expedite fixing the tests

PR #401, opened 2024-08-26
test: add test for session action failing by TIMEOUT

YutongLi291:
When tasks take more time than the defined timeout in the job template, the worker agent should FAIL the corresponding session action with progressMessage containing a "TIMEOUT" string.
We should add testing to verify this functionality.
Add a test to verify this functionality with session action TIMEOUTs.

Also increase backoff timeout for the job attachment sync test, as sometimes it takes longer than the backoff we defined for the worker to pickup the job.
Better worker agent code quality 
```
# Linux
source .e2e_linux_infra.sh
hatch run linux-e2e-test
hatch run cross-os-e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run windows-e2e-test
hatch run cross-os-e2e-test
```
No
No
----

### DIFF
--- a/test/e2e/cross_os/test_job_submissions.py
+++ b/test/e2e/cross_os/test_job_submissions.py
@@ -32,6 +32,7 @@ LOG = logging.getLogger(__name__)
 @pytest.mark.usefixtures("session_worker")
 @pytest.mark.parametrize("operating_system", [os.environ["OPERATING_SYSTEM"]], indirect=True)
 class TestJobSubmission:
+
     @pytest.mark.parametrize(
         "run_actions,environment_actions, expected_failed_action",
         [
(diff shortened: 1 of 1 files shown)


test/e2e/cross_os/test_job_submissions.py
```diff
@@ -32,6 +32,335 @@
 @pytest.mark.usefixtures("session_worker")
 @pytest.mark.parametrize("operating_system", [os.environ["OPERATING_SYSTEM"]], indirect=True)
 class TestJobSubmission:
+
+    @pytest.mark.parametrize(
+        "append_string_script",
+        [
+            (
+                "#!/usr/bin/env bash\n\n  echo -n $(cat {{Param.DataDir}}/files/test_input_file){{Param.StringToAppend}} > {{Param.DataDir}}/output_file\n"
```
leongdl: Why the different `"` vs `'''` Is it because of the `"` in the windows branch?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -32,6 +32,335 @@
...
+                    json.dumps(
+                        {
+                            "specificationVersion": "jobtemplate-2023-09",
+                            "name": "AppendStringJob",
+                            "parameterDefinitions": [
+                                {
+                                    "name": "DataDir",
+                                    "type": "PATH",
+                                    "dataFlow": "INOUT",
+                                },
```
leongdl: nit formatting

test/e2e/cross_os/test_job_submissions.py
YutongLi291: Yes, it's mostly because of that. By the way this test (and the one after) was just moved to the beginning of this file for this PR. It's been already merged a few weeks ago.

leongdl (commented):
The PR's name vs content is very different. The PR has many more test cases than the title.

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -845,504 +1174,268 @@ def test_worker_always_runs_env_exit_despite_failure(
...
-            wait_gen=backoff.constant,
-            max_time=60,
-            interval=2,
+    def test_worker_fails_session_action_timeout(
+        self,
+        deadline_resources: DeadlineResources,
+        deadline_client: DeadlineClient,
+    ) -> None:
+        # Test that if a task takes longer than the timeout defined, the session action goes to FAILED status
+        job = Job.submit(
```
ddneilson: [`max_retries_per_task`](https://github.com/aws-deadline/deadline-cloud-test-fixtures/blob/adac342fc6dff6c6174ae39783448b1020cc9a61/src/deadline_test_fixtures/deadline/resources.py#L367C9-L367C29) should also probably be set to 1, rather than the default of 5.

YutongLi291:
> The PR's name vs content is very different. The PR has many more test cases than the title.

As I've put in the description, I've moved around some of the tests.

test/e2e/cross_os/test_job_submissions.py
YutongLi291: I already ran `hatch run fmt`, so not sure the issue?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -1332,17 +1424,16 @@ def check_percentage(complete_percentage):
...
                         return complete_percentage == 100
 
-            else:
-                return False
+            return False
 
         assert check_percentage(complete_percentage)
 
-        output_path: dict[str, list[str]] = wait_for_job_output(
+        output_path: Dict[str, List[str]] = wait_for_job_output(
```
ddneilson: Why the change to Python 3.7 syntax for the typehints? The agent requires at least Python 3.9, so the previous typehints are just fine.

ddneilson (changes requested):

test/e2e/cross_os/test_job_submissions.py
YutongLi291: Ah, this is just the syntax that I was used to, in the other code, so I wanted to be consistent. I could change it back if needed

AlexTranAmz (approved):

ddneilson (approved):

PR #403, opened 2024-08-27
test: add test that worker reports task progress and message

YutongLi291:
The worker is able to report the progressPercent of a task, as well as the progressMessage, which should both be returned in the `GetSessionActions` API.

We should confirm that this functionality is working properly, to prevent future regressions.
Add a test to make sure that the progressPercent and progressMessage fields are reported as expected.
Better code quality in the future regarding the `progressPercent` and `progressMessage` reporting logic in the worker.
```
# Linux
source .e2e_linux_infra.sh
hatch run linux-e2e-test
hatch run cross-os-e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run windows-e2e-test
hatch run cross-os-e2e-test
```
No
No
----

### DIFF
--- a/test/e2e/cross_os/test_job_submissions.py
+++ b/test/e2e/cross_os/test_job_submissions.py
@@ -1437,3 +1437,166 @@ class TestJobSubmission:
             output_file_content: str = output_file.read()
             # Verify that the hash is the same
             assert output_file_content == combined_hash
+
+    def test_worker_reports_task_progress_and_status_message(
+        self,
+        deadline_resources: DeadlineResources,
+        deadline_client: DeadlineClient,
+    ) -> None:
+
+        # Make sure that worker reports task progress, as well as the status message
+
+        # Submit a job with a task that sleeps for 60 seconds , which is more than the UpdateWorkerSchedule interval of 30 seconds
+
+        test_run_status_message: str = "Sleep job is running!"
+        sleep_script: str = (
+            f"""
+            #!/usr/bin/env bash
+            percent=0
+            
+            while [ $percent -le 100 ] 
+            do
+                echo "openjd_progress: $percent"
+                echo "openjd_status: {test_run_status_message}"
+                ((percent+=10))
+                sleep 6
+            done
+            """
+            if os.environ["OPERATING_SYSTEM"] == "linux"
+            else f"""
+            $percent = 0
+            while ($percent -le 100) {{
+                Write-Output "openjd_progress: $percent"
+                Write-Output "openjd_status: {test_run_status_message}"
+                $percent += 10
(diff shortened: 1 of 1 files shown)


AlexTranAmz (dismissed):

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -1437,3 +1437,160 @@ def check_percentage(complete_percentage) -> bool:
...
+                            "embeddedFiles": [
+                                {
+                                    "name": "runScript",
+                                    "type": "TEXT",
+                                    "runnable": True,
+                                    "data": sleep_script,
+                                    **(
+                                        {"filename": "SleepScript.ps1"}
+                                        if os.environ["OPERATING_SYSTEM"] == "windows"
+                                        else {}
```
jericht, line 1516: Shouldn't this be a filename like `SleepScript.sh` or something as well?

test/e2e/cross_os/test_job_submissions.py
```diff
@@ -1437,3 +1437,160 @@ def check_percentage(complete_percentage) -> bool:
...
+                session_actions: list[dict[str, Any]] = deadline_client.list_session_actions(
+                    farmId=job.farm.id,
+                    queueId=job.queue.id,
+                    jobId=job.id,
+                    sessionId=session["sessionId"],
+                ).get("sessionActions")
+                # There should be at most 1 sessionAction as there is only one task
+                assert len(session_actions) <= 1
+                if session_actions:
+                    return session_actions[0]["sessionActionId"]
```
jericht: There should also only be one session right? Can we remove the `for` loop and replace with an assertion that only 1 session was created? Otherwise, we are just returning the first session action ID from one of potentially many sessions

YutongLi291: Okay, that totally makes sense I missed that. Will change

test/e2e/cross_os/test_job_submissions.py
YutongLi291: Only windows need a file to run. Linux does not

jericht: This filename is an optional parameter. We use it here on Windows so the script runs as powershell and not a batch script.

https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#61-embeddedfiletext

> filename — The filename for the written file. This must strictly be the basename of the filename, and not contain any directory pathing. (i.e. foo.txt not dir/foo.txt). Defaults to a random filename if not provided.

jericht (approved):

AlexTranAmz (approved):

PR #408, opened 2024-09-03
fix: Windows workers become NOT_RESPONDING on host shutdown

ttblanchard:
When the Windows Service initiates the Worker shutdown procedure, the scheduler does not drain the remaining work, and the Worker status is never updated before the instance is Shutdown.

If we run the worker as a service on EC2 - and then terminate the instance from the EC2 console, we can observe the final output from the Worker:

```
{"level": "DEBUG", "message": "interval = 15"}
{"level": "INFO", "message": "Windows Service is being stopped"}
{"level": "DEBUG", "message": "EC2 shutdown monitoring thread exited"}
{"level": "DEBUG", "message": "monitor ec2 shutdown future complete"}
{"level": "DEBUG", "message": "Waiting for threads to join..."}
```

If we query the service after this, the workers status was never updated. Eventually it will be marked as STOPPED/UNHEALTHY by the heartbeat check. 

The main worker event loop has 2 threads (futures). The first is the [thread which monitors for EC2 shutdown events](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/worker.py#L218), such as Auto-scaling scale in events, or Spot instance interruption events. The loop in this thread runs with a 1 second timeout. In the second [thread we have the scheduler](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/scheduler/scheduler.py#L279), this thread runs with a 15 second timeout. 

Since the `monitor_ec2_shutdown` thread completed because the `stop` event was set, [it returns `None`, and the main event loop never tells the scheduler to stop](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/worker.py#L243).

The solution is simply to wake up the scheduler thread in this case. Allowing it to detect that the `stop` event was set and perform it's own shutdown naturally. 

With this change we can see the scheduler shuts down and the status is updated from the worker logs:

```
{"level": "DEBUG", "message": "interval = 15"}
{"level": "INFO", "message": "Windows Service is being stopped"}
{"level": "DEBUG", "message": "EC2 shutdown monitoring thread exited"}
{"level": "DEBUG", "message": "monitor ec2 shutdown future complete"}
{"level": "DEBUG", "message": "Waiting for threads to join..."}
{"level": "INFO", "message": "Main event loop exited."}
{"level": "INFO", "message": "Draining any remaining Sessions."}
{"level": "INFO", "message": "Worker shutdown complete"}
{"level": "INFO", "message": "Setting Worker state to STOPPED."}
{"level": "INFO", "ti": "📤", "type": "API", "subtype": "Req", "operation": "deadline:UpdateWorker", "params": {"status": "STOPPED"}, "request_url": "redacted", "resource": {"farm-id": "redacted", "fleet-id": "redacted", "worker-id": "redacted"}}
{"level": "INFO", "ti": "📥", "type": "API", "subtype": "Resp", "operation": "deadline:UpdateWorker", "status_code": 200, "params": {}, "request_id": "redacted"}	
{"level": "INFO", "ti": "💻", "type": "Worker", "subtype": "Status", "message": "Status set to STOPPED.", "farm_id": "redacted", "fleet_id": "redacted", "worker_id": "redacted"}
```

When running as a service on Windows and the instance is shutdown the worker will now complete it's shutdown behavior, marking itself as STOPPED with the service and draining it's scheduler, freeing up other workers to pick up tasks. 

I built and deployed a worker instance with this change to a CMF, I scheduled a bunch of work and once the worker started picking up tasks I terminated the instance and observed the logs. 

No.

No.
----

### DIFF
--- a/src/deadline_worker_agent/worker.py
+++ b/src/deadline_worker_agent/worker.py
@@ -246,6 +246,12 @@ class Worker:
                                 grace_time=worker_shutdown.grace_time,
                                 fail_message=worker_shutdown.fail_message,
                             )
+                        else:
+                            # If we are here, it's because self._stop.set() was set causing the monitor_ec2_shutdown thread to join.
+                            # The scheduler thread has a longer wait, so let's wake it up so it can join as well.
+                            self._scheduler.shutdown(
+                                fail_message="The Worker received a shutdown event locally from the host machine."
+                            )
                     elif future is scheduler_future:
                         logger.debug("scheduler future complete")
                         try:


src/deadline_worker_agent/worker.py
```diff
@@ -246,6 +246,10 @@ def run(self) -> None:
                                 grace_time=worker_shutdown.grace_time,
                                 fail_message=worker_shutdown.fail_message,
                             )
+                        else:
+                            # If we are here, it's because self._stop.set() was set causing the monitor_ec2_shutdown thread to join.
+                            # The scheduler thread has a longer wait, so let's wake it up so it can join as well.
+                            self._scheduler._wakeup.set()
```
jusiskin: Should we be using:

```
self._scheduler.shutdown(...)
```

instead? It is marked as a public API and allows providing a reason to the tasks being interrupted.

jusiskin (changes requested):

src/deadline_worker_agent/worker.py
AWS-Samuel: While this does work, calling the private `_wakeup` event doesn't feel ideal. 

It looks like calling `self._scheduler.shutdown()` [also sets the `_wakeup` event](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/scheduler/scheduler.py#L1483)

I wonder if we should instead modify [`_monitor_ec2_shutdown`](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/worker.py#L364)  to return a [`WorkerShutdown`](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/worker.py#L44). This would give the benefit of letting the scheduler handle any edge cases it might need to in the future if the implementation ever changes

jusiskin: To me, it doesn't seem like this is the responsibility of the `_monitor_ec2_shutdown()` method. That method should only be responsible for handling EC2-based shutdown events. Instead, it would be ideal if the code that responds to the event:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/353d9a25448462484fa8a263318520f63a8015e2/src/deadline_worker_agent/windows/win_service.py#L43-L45

could somehow inform the shutdown grace period and provide the interruption status message when calling `self._scheduler.shutdown(...)`.

ttblanchard: Updated and re-tested!

AWS-Samuel (approved):

src/deadline_worker_agent/worker.py
ttblanchard: I considered adding to the monitor_ec2_shutdown function. However, this doesn't just include EC2 shutdowns, it could be a local host. This could also occur when the service is forcefully stopped, but the host is still running. 

I decided to let the scheduler just use the default grace period, rather than define some arbitrary value since we won't have any idea what a good grace period is for this case. We are just making a best effort here to let the worker shut down cleanly.

AWS-Samuel: Fair, we could also rename `_monitor_ec2_shutdown()` to something like `_monitor_shutdown` since in this case it is required for us to detect any shutdown at all. 

Regardless, what's in this PR works and now that we're calling `_scheduler_.shutdown` directly I think this is good enough to merge

AWS-Samuel (dismissed):

src/deadline_worker_agent/worker.py
```diff
@@ -246,6 +246,12 @@ def run(self) -> None:
                                 grace_time=worker_shutdown.grace_time,
                                 fail_message=worker_shutdown.fail_message,
                             )
+                        else:
+                            # If we are here, it's because self._stop.set() was set causing the monitor_ec2_shutdown thread to join.
+                            # The scheduler thread has a longer wait, so let's wake it up so it can join as well.
+                            self._scheduler.shutdown(
+                                fail_message="The Worker recieved a shutdown event."
```
jusiskin: Two suggestions:

1.  nit: type `recieved` &rarr; `received`
2.  Can I make a small suggestion here to indicate that the shutdown event was received locally on the host? Because the worker can also receive shutdown events from the service API responses.

Putting these together, how about:

```suggestion
                                fail_message="The Worker received a shutdown event locally from the host machine."
```

jusiskin (changes requested):

AWS-Samuel (approved):

AWS-Samuel (approved):

jusiskin (approved):

Issue #406, opened 2024-08-30
Bug: Object of type WindowsPath is not JSON serializable when running on Windows


AWS-Samuel:
### Describe Behaviour

When exiting a session action the following bug was observed:

```
[2024-08-29 22:29:24,927][INFO    ] 💾 FileSystem.Write 💾 Saving profile updates. [C:\ProgramData\Amazon\Deadline\Cache\queues\queue-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\config]
[2024-08-29 22:29:24,928][INFO    ] 💾 FileSystem.Write 💾 Saving profile updates. [C:\ProgramData\Amazon\Deadline\Cache\queues\queue-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\credentials]
[2024-08-29 22:29:24,929][ERROR   ] Exception in WorkerScheduler
Traceback (most recent call last):
  File "C:\Program Files\Python312\Lib\site-packages\deadline_worker_agent\scheduler\scheduler.py", line 291, in run
    interval = self._sync(interruptable=True)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python312\Lib\site-packages\deadline_worker_agent\scheduler\scheduler.py", line 454, in _sync
    self._update_sessions(response=response)
  File "C:\Program Files\Python312\Lib\site-packages\deadline_worker_agent\scheduler\scheduler.py", line 612, in _update_sessions
    self._cleanup_queue_aws_credentials(assigned_sessions=assigned_sessions)
  File "C:\Program Files\Python312\Lib\site-packages\deadline_worker_agent\scheduler\scheduler.py", line 1211, in _cleanup_queue_aws_credentials
    credentials_dataclass.session.cleanup()
  File "C:\Program Files\Python312\Lib\site-packages\deadline_worker_agent\aws_credentials\queue_boto3_session.py", line 236, in cleanup
    self._delete_credentials_directory()
  File "C:\Program Files\Python312\Lib\site-packages\deadline_worker_agent\aws_credentials\queue_boto3_session.py", line 482, in _delete_credentials_directory
    _logger.warning(
  File "C:\Program Files\Python312\Lib\logging\__init__.py", line 1551, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "C:\Program Files\Python312\Lib\logging\__init__.py", line 1684, in _log
[...]

PR #412, opened 2024-09-05
fix: WindowsPath is not JSON serializable during session cleanup

AWS-Samuel:
There is a bug: `Object of type WindowsPath is not JSON serializable` which can crash the worker agent when running on Windows. See https://github.com/aws-deadline/deadline-cloud-worker-agent/issues/406

I was able to recreate this by changing the ACL of the queues cache directory (which stores queue credentials) to prevent deletion. 

Convert `Path` types to `str` before logging. Doing this in the `FilesystemLogEvent` class since it will catch all present and future cases of logging a `Path` object.

Fixes a worker agent crash: https://github.com/aws-deadline/deadline-cloud-worker-agent/issues/406

Reproduced https://github.com/aws-deadline/deadline-cloud-worker-agent/issues/406, then added this change in and confirmed the worker agent instead logged the message properly and resumed 

```
2024/09/05 16:07:53-05:00 {"level": "WARNING", "ti": "💾", "type": "FileSystem", "subtype": "Delete", "message": "Failed to delete.", "filepath": "C:\\ProgramData\\Amazon\\Deadline\\Cache\\queues\\queue-2bb9d04f40da43f1b9bd204ef2684e64"}
```

It is here!

No

----

### DIFF
--- a/src/deadline_worker_agent/log_messages.py
+++ b/src/deadline_worker_agent/log_messages.py
@@ -1,4 +1,5 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+from __future__ import annotations
 
 import sys
 from enum import Enum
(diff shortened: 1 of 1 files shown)


src/deadline_worker_agent/log_messages.py
```diff
@@ -1,4 +1,5 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+from __future__ import annotations
```
AWS-Samuel, line 2: We need this to maintain python 3.9 support

jusiskin (approved):

gmchale79 (approved):

PR #413, opened 2024-09-10
fix: vague error message when no AWS region specified

AWS-Samuel:

When running the worker agent from a CLI, a frequently encountered issue is the worker agent failing with a message like:

```
[CRITICAL] You must specify a region.
[INFO    ] 🚪 Worker Agent exiting
```

This is a bit vague and is hard to debug if you aren't familiar with the reason for that error message.

Catch `NoRegionError`s from boto, and log a warning with guidance to check the worker agent documentation on how to set the AWS region.

Also update the README with information for running the worker agent outside of a OS service context
 
Hopefully more clear messaging on a common situation of starting the worker agent without a region specified, which often happens when running from the command line (since the WA service installation sets the region).

Tested on a machine and removed the region configuration in the system installation. Then started the worker agent and observed the following lines:

```
[WARNING ] The Worker Agent was started with no AWS region specified. Refer to the Deadline Cloud Worker Agent documentation for guidance.
[CRITICAL] You must specify a region.
[INFO    ] 🚪 Worker Agent exiting
```

Yes, through the commit message/changelog

No
----

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,23 @@ REM stop the service
 sc.exe stop DeadlineWorker
 ```
 
+### Running Outside of an Operating System Service
+
+---
+**NOTE:**
+It is highly recommended to run the worker agent through an OS service in production environments. The OS service automatically handles:
+- Restarting the worker agent process if the agent process crashes.
+- Starting the worker agent when the host machine starts up.
+- Attempting graceful shutdown when the host machine is shutting down.
+----
+The worker agent can also be started outside of a service context if required. Run `deadline-worker-agent --help` to see a list of supported command line arguments.
+
+ **NOTE:** You must have an [AWS region](https://docs.aws.amazon.com/sdkref/latest/guide/feature-region.html) specified in order to run the worker agent from the command line. This can be configured through:
+  - The `AWS_DEFAULT_REGION` [environment variable](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html#envvars-list).
+  - The [AWS Configuration File](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where)
+    - Configured [manually](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-format), or through the [command line](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-methods).
+
+
 ## Logging
 
 [See logging](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/release/docs/logging.md)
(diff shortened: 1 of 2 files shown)


moorec-aws (dismissed):

README.md
```diff
@@ -175,6 +175,17 @@ REM stop the service
 sc.exe stop DeadlineWorker
 ```
 
+### Running Outside of an Operating System Service
+
+The worker agent can also be started outside of a service context if required. Run `deadline-worker-agent --help` to see a list of supported command line arguments.
```
jusiskin, line 187: Can we add a warning indicate that using the OS service is the preferred approach for production use. The benefits of the OS service are:

1.  it restarts the worker agent process if it crashes
2. handles starting the agent when the host machine boots
3. handles attempts to gracefully shutting down the agent when the host machine is shutting down

ttblanchard (dismissed):

README.md
```diff
@@ -175,6 +175,17 @@ REM stop the service
...
+
+**NOTE:** You must have an AWS region specified in order to run the worker agent from the command line. This can be configured through:
+  - Environment Variables:
+    - `AWS_REGION` and `AWS_DEFAULT_REGION`
+  - The AWS CLI:
+    - Run `aws configure` and specify the region.
+  - The [AWS CLI Configuration File](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html).
```
jusiskin: These are kind of related. The CLI just writes to the config file. Perhaps we can make the AWS CLI a sub-bullet of the config file. What do you think?

src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -82,7 +83,13 @@ def entrypoint(cli_args: Optional[list[str]] = None, *, stop: Optional[Event] =
         config.log()
 
         # Register the Worker
-        worker_bootstrap = bootstrap_worker(config=config)
+        try:
+            worker_bootstrap = bootstrap_worker(config=config)
+        except NoRegionError:
+            _logger.warn(
+                "The Worker Agent was started with no AWS region specified. Refer to the Deadline Cloud Worker Agent documentation for guidance."
+            )
```
jusiskin: Should we include the URL to the README? I think we'd want

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/release/README.md#running-outside-of-an-operating-system-service

jusiskin (changes requested):

README.md
AWS-Samuel: Added a warning with this information above

README.md
AWS-Samuel: Amended these bullets, let me know what you think

src/deadline_worker_agent/startup/entrypoint.py
AWS-Samuel: Added, thanks

jusiskin (approved):

moorec-aws (approved):

PR #416, opened 2024-09-11
test: add tests relating to job attachment failures and empty output

YutongLi291:
For when the customer is syncing their job attachments, either while syncing the inputs, or the outputs, we need to make sure that the sync works as expected.

In particular, the functionalities that:
* If an input file is not found, but the customer specified that all input paths should exist, we should fail the job submission, as this indicates that the customer likely forgot to save an input file 
* If there are no output files found in the output directory, the job should still pass.  This path should not have any regressions in the line
* Add two tests that test these functionalities
* Also added a lot more typing to the tests file
Better code coverage of the current features in worker-agent.
```
# Linux
source .e2e_linux_infra.sh
hatch run linux-e2e-test
hatch run cross-os-e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run windows-e2e-test
hatch run cross-os-e2e-test
```
No
No

----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -133,10 +133,6 @@ class TestJobSubmission:
             ),
         ],
     )
-    @pytest.mark.skipif(
-        os.environ["OPERATING_SYSTEM"] == "linux",
-        reason="Linux test is flaky. Disabling until further investigation.",
-    )
     def test_job_reports_failed_session_action(
         self,
         deadline_resources: DeadlineResources,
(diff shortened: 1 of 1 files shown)


test/e2e/test_job_submissions.py
```diff
@@ -1289,6 +1290,163 @@ def test_worker_uses_job_attachment_configuration(
...
+                            ],
+                        }
+                    )
+                )
+
+            config = configparser.ConfigParser()
+
+            set_setting("defaults.farm_id", deadline_resources.farm.id, config)
+            set_setting("defaults.queue_id", deadline_resources.queue_a.id, config)
+            with pytest.raises(Exception):
```
moorec-aws: Isn't this just testing the deadline cloud API? What are we testing for the worker here?

AlexTranAmz: same question

YutongLi291: There's no testing for the worker in this case. It tests the job attachments instead so I think I will remove this test. Thanks for calling this out!

YutongLi291:
SonarCloud failing due to the typing I added.

moorec-aws (dismissed):

AlexTranAmz (approved):

jairaws (approved):

PR #423, opened 2024-09-25
test: add test to verify worker enter stopping state while draining

AlexTranAmz:
We having a gap in e2e test to make sure worker report stopping/stopped when we kill worker
Implement a test to make sure when we submit a command to shutdown worker, it will report correctly as stopping/stopped

Increase test coverage and improve confident when release worker agent
`hatch run e2e-test`
Both on Windows and Linux
Yes
No

----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -25,7 +25,7 @@ import uuid
 import os
 import configparser
 import tempfile
-from e2e.utils import wait_for_job_output, submit_sleep_job
+from e2e.utils import wait_for_job_output, submit_sleep_job, submit_custom_job
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


test/e2e/test_job_submissions.py
```diff
@@ -1727,3 +1685,59 @@ def session_action_has_expected_progress(session_action_id) -> bool:
...
+            """
+            #!/usr/bin/env bash
+            while true
+            do
+                sleep 1
+            done
+            """
+            if os.environ["OPERATING_SYSTEM"] == "linux"
+            else """
+            while ($true) {{
```
YutongLi291: Seems like this could be a bit dangerous if the worker doesn't stop properly. THe job will run forever. Can we make this not a WHile true loop? Maybe have it timeout after 10 minutes for example.

YutongLi291 (changes requested):

YutongLi291 (approved):

test/e2e/test_job_submissions.py
```diff
@@ -1727,3 +1685,54 @@ def session_action_has_expected_progress(session_action_id) -> bool:
...
+        sleep_script: str = (
+            """
+            #!/usr/bin/env bash
+            sleep 600
+            """
+            if os.environ["OPERATING_SYSTEM"] == "linux"
+            else """
+            Start-Sleep -Seconds 600
+            """
+        ),
```
jericht, line 1703: Does this need to be a function parameter? Can we just define it as a variable in the test?

test/e2e/test_job_submissions.py
```diff
@@ -1727,3 +1685,54 @@ def session_action_has_expected_progress(session_action_id) -> bool:
...
+    ):
+
+        submit_custom_job(
+            job_name="Infinite Sleep Job",
```
jericht, line 1707: nit: Not infinite, just long :) Can we say 10 minute sleep job?

test/e2e/test_job_submissions.py
```diff
@@ -1727,3 +1685,54 @@ def session_action_has_expected_progress(session_action_id) -> bool:
...
+            cmd_result = function_worker.send_command("sc.exe stop DeadlineWorker")
+
+        assert cmd_result.exit_code == 0
+
+        @backoff.on_predicate(
+            wait_gen=backoff.constant,
+            max_time=120,
+            interval=10,
+        )
+        def worker_stop(worker: EC2InstanceWorker) -> bool:
```
jericht, line 1726:
```suggestion
        def is_worker_stopped(worker: EC2InstanceWorker) -> bool:
```

nit: more descriptive function name

jericht (approved):
Just left some tiny nits, feel free to address now or later

PR #424, opened 2024-10-02
test: add test for submitting a job sync job to a queue with nonvalid role

YutongLi291:
We have code logic that makes sure if the queue role is nonvalid, i.e. it does not have the right permissions to get objects from the S3 Job Attachments bucket, the syncInputJobAttachments step will fail.
We need to make sure that there are no regressions in this code.

Add a queue that has a queue role without `Get` access to the Job attachments S3 bucket. 
Add a test that confirms that when submitting a job with a job attachments to this queue, the job will fail after submission with the faiilure being the `syncInputJobAttachments` step
Prevention of regressions in this part of the code for `syncInputJobAttachments` when the queue role lacks permissions.

Deployed the new queue using `./scripts/deploy_e2e_testing_infrastructure.sh`
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test

```
No
No
----

### CODE REVIEW

File: scripts/e2e_testing_infrastructure.yaml [hunk 1]
```diff
@@ -561,6 +564,37 @@ Resources:
...
+              - Effect: Allow
+                Action:
+                  - s3:PutObject
+                  - s3:ListBucket
+                Resource:
+                  - !GetAtt JobAttachmentsBucket.Arn
+                  - !Join
+                    - ''
+                    - - !GetAtt JobAttachmentsBucket.Arn
+                      - /*
```
jericht, line 593: nit: Can we add a comment here explaining why this in nonvalid? What permission is missing?

jericht (approved):

AlexTranAmz (approved):
LGTM

PR #436, opened 2024-10-09
test: add test that verifies job fails if attachment is accidentally deleted from bucket

YutongLi291:
If in the case that an user accidentally deletes a job attachment from the job attachments bucket (or clears the job attachments bucket completely),  if a job is running that depends on any of the deleted job attachments, the job should fail at the `syncInputJobAttachments` stage since the job attachment is not able to be synced from the bucket to the worker.

We should test that this is indeed what happens, and that the worker is still able to process other jobs after this failure.

Add a test that 
1. Submits a job with input job attachments in SUSPENDED status, this will ensure that the input job attachments are uploaded to the Job Attachments bucket but not synced to the worker yet, as the job has yet to start.
2. Using the manifest, find any of the hashed input job attachment files in the job attachments bucket, and delete it
3. Set the job status to READY, ensuring that the worker will start working on the job
4. Wait for the job to finish, in which case it should be in FAILED status since the `syncInputJobAttachments` step failed
5. Check that `syncInputJobAttachments` session action failed using `ListSessionActions`
6. Submit another test job that sleeps, ensuring it passes without fail to verify that the worker is still running as expected. 
Better testing for worker agent code and handling failures
```
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test

```
No
No
----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -1730,6 +1730,7 @@ class TestJobSubmission:
         self,
         deadline_resources: DeadlineResources,
         deadline_client: DeadlineClient,
+        session_worker: EC2InstanceWorker,
         tmp_path: pathlib.Path,
     ) -> None:
         # Test that submits a job that has step step dependencies and confirm that the final output is as we expect
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -27,6 +27,7 @@ import configparser
 import tempfile
 from e2e.utils import wait_for_job_output, submit_sleep_job, submit_custom_job
 
+
 LOG = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


AlexTranAmz (approved):

test/e2e/test_job_submissions.py
```diff
@@ -2015,6 +2016,215 @@ def session_action_has_expected_progress(session_action_id) -> bool:
...
+        with open(
+            os.path.join(job_bundle_path, "parameter_values.json"), "w+"
+        ) as parameter_values_file:
+            # Make sure the job is submitted in SUSPENDED state so we have time to delete an input job attachment in the bucket
+            parameter_values_file.write(
+                json.dumps(
+                    {
+                        "parameterValues": [
+                            {"name": "deadline:targetTaskRunStatus", "value": "SUSPENDED"},
+                        ]
```
jericht, line 2052: Is this duplicated above in `job_parameters` variable?

test/e2e/test_job_submissions.py
```diff
@@ -2015,6 +2016,215 @@ def session_action_has_expected_progress(session_action_id) -> bool:
...
+            get_manifest_object_result["Body"].read()
+        )
+
+        # Get the Job Attachment bucket file paths of the input files
+        input_file_paths: list[dict[str, Any]] = get_object_result_body["paths"]
+        first_input_file_hash = input_file_paths[0]["hash"]
+
+        # Delete one of the input files from the Job Attachments bucket after confirming that it exists
+
+        s3_client.get_object(
```
jericht, line 2147: [HeadObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html) is a lighter API call if we just want to check for existence, but if the role doesn't have permission to call it already, don't worry about it.

test/e2e/test_job_submissions.py
```diff
@@ -2015,6 +2016,215 @@ def session_action_has_expected_progress(session_action_id) -> bool:
...
+        found_failed_session_action = False
+        for session in sessions:
+            session_actions = deadline_client.list_session_actions(
+                farmId=job.farm.id,
+                queueId=job.queue.id,
+                jobId=job.id,
+                sessionId=session["sessionId"],
+            ).get("sessionActions")
+
+            logging.info(f"Session actions: {session_actions}")
```
jericht, line 2189:
```suggestion
            LOG.info(f"Session actions: {session_actions}")
```

Should use same logger as above for consistency. I believe this method goes to [root logger](https://docs.python.org/3/library/logging.html#logging.info)

jericht (approved):

test/e2e/test_job_submissions.py
YutongLi291: Will address in next PR

PR #437, opened 2024-10-10
test: add test that verifies subsequent tasks are NEVER_ATTEMPTED when previous action is canceled

YutongLi291:
The scheduler doesn’t pipeline more than one task run in a session until the first task run in that session has completed.

We should verify that this behaviour is as expected, and that if the task run is not completed, other tasks should not have been attempted.

Add an E2E test that verifies that when a task is canceled during its run, the subsequent tasks are in `NEVER_ATTEMPTED` status.

Also edited an existing test that verifies `CANCELED` actions that other subsequent actions are `NEVER_ATTEMPTED`

Better testing to verify worker behaviour is as expected when task is canceled, that following tasks are NEVER_ATTEMPTED
```
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test

```
No
No
----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -159,7 +159,7 @@ class TestJobSubmission:
                     sessionId=session["sessionId"],
                 ).get("sessionActions")
 
-                logging.info(f"Session actions: {session_actions}")
+                LOG.info(f"Session actions: {session_actions}")
                 for session_action in session_actions:
                     # Session action should be failed IFF it's the expected action to fail
                     if expected_failed_action in session_action["definition"]:
@@ -246,7 +246,7 @@ class TestJobSubmission:
                 sessionId=session["sessionId"],
             ).get("sessionActions")
 
-            logging.info(f"Session Actions: {session_actions}")
+            LOG.info(f"Session Actions: {session_actions}")
             for session_action in session_actions:
                 # taskRun session action should be failed
                 if "taskRun" in session_action["definition"]:
@@ -428,10 +428,16 @@ class TestJobSubmission:
                     if expected_canceled_action in session_action["definition"]:
                         if session_action["status"] == "CANCELED":
                             found_canceled_session_action = True
+                    elif "envExit" in session_action["definition"]:
+                        # envExit should always run no matter what
+                        if session_action["status"] != "SUCCEEDED":
+                            return False
                     else:
-                        assert (
-                            session_action["status"] != "CANCELED"
-                        )  # This should not happen at all, so we fast exit
+                        if expected_canceled_action == "envEnter":
+                            # If we canceled the envEnter, everything else should have been NEVER_ATTEMPTED
+                            assert session_action["status"] == "NEVER_ATTEMPTED"
+                        else:
+                            assert session_action["status"] == "SUCCEEDED"
             return found_canceled_session_action
 
         sessions: list[dict[str, Any]] = deadline_client.list_sessions(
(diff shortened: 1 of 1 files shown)


test/e2e/test_job_submissions.py
```diff
@@ -427,10 +430,15 @@
...
+                    elif "envExit" in session_action["definition"]:
+                        # envExit should always run no matter what
+                        assert session_action["status"] == "SUCCEEDED"
                     else:
-                        assert (
-                            session_action["status"] != "CANCELED"
-                        )  # This should not happen at all, so we fast exit
+                        if expected_canceled_action == "envEnter":
+                            # If we canceled the envEnter, everything else should have been NEVER_ATTEMPTED
+                            session_action["status"] == "NEVER_ATTEMPTED"
```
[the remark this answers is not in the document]
YutongLi291: Resolved

test/e2e/test_job_submissions.py
```diff
@@ -837,6 +845,221 @@ def sync_input_actions_are_canceled(sessions: List[Dict[str, Any]]) -> bool:
...
+                                "onRun": {"command": "whoami"},
+                            }
+                        },
+                    },
+                ],
+            },
+        )
+
+        # Wait for the job to start
+
```
leongdl, line 964: nit extra line

test/e2e/test_job_submissions.py
```diff
@@ -842,6 +850,221 @@ def sync_input_actions_are_canceled(sessions: List[Dict[str, Any]]) -> bool:
...
+                farmId=job.farm.id, queueId=job.queue.id, jobId=job.id
+            ).get("sessions")
+            if sessions and len(sessions) > 0:
+                return True
+            return False
+
+        assert is_job_started_with_sessions(job)
+
+        # Wait some time for the second step (which sleeps for 2 minutes) to start
+        time.sleep(20)
```
hindleym: Instead of sleeping here, it may be worth just waiting on the second step to switch to `RUNNING`? That way if it starts sooner, the test can complete faster.

test/e2e/test_job_submissions.py
```diff
@@ -406,6 +406,9 @@ def sessions_exist(current_job: Job) -> bool:
 
         LOG.info(f"Job result: {job}")
 
+        # Wait until the envExit runs as well
+        time.sleep(10)
```
hindleym: Is anything actually happening on exit that we need to sleep this long? I see that it waits for the job to complete above. Should that function not already make sure that the job has finished it's calls before returning?

hindleym (commented):
I would avoid forcing tests to sleep as much as possible.

test/e2e/test_job_submissions.py
YutongLi291: The job is considered completed even without running the envExit and envExit is not required to finish for the job to finish.

test/e2e/test_job_submissions.py
YutongLi291: That's true, but that's also more API calls. There's a tradeoff, and I believe that 20 seconds is not a big time wait to justify the extraneous API calls.

hindleym: I see the `sleep` command more as a last resort. It guarantees to slow down all testing, especially when testing new changes before PR. Between these two sleeps, you're adding thirty seconds to every full test attempt, which already takes an extremely long time to complete. If I'm not mistaken, most people make sure their changes work with the full suite more than once before creating a PR, that's at least a minute to every future change people try to merge.

hindleym (approved):

leongdl (approved):

PR #439, opened 2024-10-16
test: add E2E test that OS service restarts worker agent when it exits

YutongLi291:

On Linux, the worker agent creates a systemd service when running `install-deadline-worker` (without the `--no-service` argument). One responsibility of this systemd service is to restart the worker agent process if it exits (e.g. a program crash).

We require automated end-to-end testing to ensure this functionality works as expected.

Add a test that verifies that when linux worker shuts down by a SIGKILL signal, the worker service restarts the process.
Better code quality for the code that is responsible for restarting the worker.
```
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test
```
No
No

----

### DIFF
--- a/test/e2e/test_worker_status.py
+++ b/test/e2e/test_worker_status.py
@@ -3,12 +3,14 @@
 This test module contains tests that verify the Worker agent's behavior by starting/stopping the Worker,
 and making sure that the status of the Worker is that of what we expect.
 """
+from datetime import datetime, timezone
 import logging
 import os
 import pytest
 from deadline_test_fixtures import DeadlineClient, EC2InstanceWorker
 import pytest
 from e2e.utils import is_worker_started, is_worker_stopped
+import backoff
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


test/e2e/test_worker_status.py
```diff
@@ -40,3 +42,83 @@ def test_worker_lifecycle_status_is_expected(
...
+
+    def test_linux_worker_restarts_process(
+        self,
+        deadline_resources,
+        deadline_client: DeadlineClient,
+        function_worker: EC2InstanceWorker,
+    ) -> None:
+        # Verifies that Linux Worker service restarts the process when we start/stop worker process
+
+        assert function_worker.worker_id is not None  # To fix linter type mismatch
```
leongdl: to fix?

leongdl (commented):
Is the "to fix" something we must fix?

test/e2e/test_worker_status.py
jericht: I think this is needed because the linter will complain without it since `worker_id` is `Optional` IIRC

test/e2e/test_worker_status.py
```diff
@@ -40,3 +42,83 @@ def test_worker_lifecycle_status_is_expected(
...
+        pkill_command_result = function_worker.send_command(
+            f"sudo pkill -9 --full -u {function_worker.configuration.agent_user} deadline-worker-agent"
+        )
+        assert (
+            pkill_command_result.exit_code == 0
+        ), f"Failed to kill the worker agent process: {pkill_command_result}"
+
+        # Wait for the process to be restarted by the service
+
+        time.sleep(10)
```
jericht: nit: I'm pretty sure 10 seconds is more than enough for this test, but I think a better pattern would be to build in retries with a timeout (10 seconds is plenty, you can probably check after a second or two even) so we at least don't give up on the first failure for this check. Magic sleeps like this tend to lead to flakiness in the tests.

jericht (dismissed):

test/e2e/test_worker_status.py
YutongLi291: As in this line fixes the linter type mismatch 😆

test/e2e/test_worker_status.py
YutongLi291: That's true, I'll create something like that

YutongLi291:
> Is the "to fix" something we must fix?

Nope, will change the comment so that's more clear.

jericht (approved):

AlexTranAmz (approved):

PR #441, opened 2024-10-16
fix: increase default instance size

YutongLi291:
We have some flaky worker agent tests in the past week, in which Windows workers took too long to start, causing multiple tests to fail 

We should fix these flakiness so the tests are trustworthy.

Increase default EC2 size so the workers boot up faster.
Less flakiness in tests lead to better reliability and consistency of the code
```
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test

```
All the tests passed, and the worker did boot up faster. (Around 5-6 minutes for windows as opposed to around 7 minutes before)
No
No
----

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -291,7 +291,7 @@ def create_worker(
         ami_id = os.getenv("AMI_ID")
         subnet_id = os.getenv("SUBNET_ID")
         security_group_id = os.getenv("SECURITY_GROUP_ID")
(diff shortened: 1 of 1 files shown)


moorec-aws (approved):

leongdl (approved):
Faster instances don't always boot faster right? We should measure how flaky before and after is to analyze.

YutongLi291:
> Faster instances don't always boot faster right? We should measure how flaky before and after is to analyze.

I have tested and it boots around 1-2 minutes faster.

PR #444, opened 2024-10-16
test: include large files in syncInputJobAttachments cancelation tests to allow more time for cancelation

YutongLi291:
`test_worker_reports_canceled_sync_input_actions_as_canceled` was failing flakily because sometimes the syncInputAttachments goes too fast so we don't have time to cancel it.
Create bigger files in between the small files, so that  the syncInputJobAttachments is much slower (due to the addition of the big files), but it does not time out due to the perceived transfer rate being too slow (which happens when too many small files are attempted to be uploaded).

The small files are actually what's slowing down the syncInputJobAttachments action, so we still need them to be able to cancel the action.
More robust `test_worker_reports_canceled_sync_input_actions_as_canceled` test
```
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test

```
Ran 5 times each, passed every time.
No
No
----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -735,10 +735,14 @@ class TestJobSubmission:
         # Create the input files to make sync inputs take a relatively long time
         files_path: str = os.path.join(tmp_path, "files")
         os.mkdir(files_path)
-        for i in range(2000):
+        for i in range(6000):
             file_name: str = os.path.join(files_path, f"input_file_{i+1}.txt")
             with open(file_name, "w+") as input_file:
-                input_file.write(f"{i}")
+                if i % 1000 == 0:
+                    # Create some big files (1GB each) so the syncInputAttachments don't fail due to low transfer rates
+                    input_file.write("A" * 1000000000)
+                else:
+                    input_file.write(f"{i}")
         config = configparser.ConfigParser()
 
         set_setting("defaults.farm_id", deadline_resources.farm.id, config)


test/e2e/test_job_submissions.py
```diff
@@ -735,10 +735,14 @@ def test_worker_reports_canceled_sync_input_actions_as_canceled(
...
         files_path: str = os.path.join(tmp_path, "files")
         os.mkdir(files_path)
-        for i in range(2000):
+        for i in range(6000):
             file_name: str = os.path.join(files_path, f"input_file_{i+1}.txt")
             with open(file_name, "w+") as input_file:
-                input_file.write(f"{i}")
+                if i % 1000 == 0:
+                    # Create some big files (1GB each) so the syncInputAttachments don't fail due to low transfer rates
+                    input_file.write("A" * 1000000000)
```
jusiskin, line 743: Is this actually a functional bug that needs to be fixed? Do jobs with large numbers of small files fail during the syncInputJobAttachments due to low transfer rates?

If so, let's create an issue in deadline-cloud: https://github.com/aws-deadline/deadline-cloud/issues/new/choose

jusiskin:
> ### What was the problem/requirement? (What/Why)
>
> `test_worker_reports_canceled_sync_input_actions_as_canceled` was failing flakily because sometimes the syncInputAttachments goes too fast so we don't have time to cancel it.
>
> ### What was the solution? (How)
>
> Create bigger files in between the small files, so that the syncInputAttachments is much slower, but it does not time out due to the perceived transfer rate being too slow.

Two things:

1.  nit: it should be `syncInputAttachments` &rarr; `synInputJobAttachments`
2. These seem to be pointing to different root-causes for the test failures. One says the transfer happens so fast we don't have time to cancel it. The other says that the syncInputAttachments fails because of perceived transfer rates being low.

Which is true? Is it both?

test/e2e/test_job_submissions.py
```diff
@@ -735,10 +735,14 @@ def test_worker_reports_canceled_sync_input_actions_as_canceled(
         # Create the input files to make sync inputs take a relatively long time
         files_path: str = os.path.join(tmp_path, "files")
         os.mkdir(files_path)
-        for i in range(2000):
+        for i in range(6000):
             file_name: str = os.path.join(files_path, f"input_file_{i+1}.txt")
             with open(file_name, "w+") as input_file:
-                input_file.write(f"{i}")
+                if i % 1000 == 0:
```
leongdl, line 741: https://unix.stackexchange.com/questions/33629/how-can-i-populate-a-file-with-random-data

This seems to be an easier solution?

YutongLi291:
> > ### What was the problem/requirement? (What/Why)
> > `test_worker_reports_canceled_sync_input_actions_as_canceled` was failing flakily because sometimes the syncInputAttachments goes too fast so we don't have time to cancel it.
> > ### What was the solution? (How)
> > Create bigger files in between the small files, so that the syncInputAttachments is much slower, but it does not time out due to the perceived transfer rate being too slow.
> 
> Two things:
> 
>     1. nit: it should be `syncInputAttachments` → `synInputJobAttachments`
> 
>     2. These seem to be pointing to different root-causes for the test failures. One says the transfer happens so fast we don't have time to cancel it. The other says that the syncInputAttachments fails because of perceived transfer rates being low.
> 
> 
> Which is true? Is it both?

For 1, I will change that.

For 2, the flaky failures are because the action completes too quickly. However, we tried before with around 5000 small files but that fails the action because the transfer rate is too slow and times out the job attachments code (which the latter part refers to, pardon the lack of context). So the middle ground is to create small files with large files in between, so that it still takes a long time, but average transfer rate is brought up by the larger files.

test/e2e/test_job_submissions.py
YutongLi291: We don't need random data, we need a lot of small files (to ensure the syncInputJobAttachments takes a long time) and a few large files (to ensure the syncInputJobAttachments does not fail because of the perceived low transfer rate from job attachments side)

jusiskin: We also cannot assume that the tests are running on a machine with `/dev/random` and `dd` available. If the host running the tests is a Windows machine without WSL, it will not have those.

jusiskin (approved):

godobyte (approved):

PR #451, opened 2024-10-24
test: add windows test that worker service restarts process

YutongLi291:

On windows, the worker agent creates a service when running `install-deadline-worker` (without the `--no-service` argument). One responsibility of this service is to restart the worker agent process if it exits (e.g. a program crash).

We require automated end-to-end testing to ensure this functionality works as expected.

Add a test that verifies that when windows worker shuts down by a SIGKILL signal, the worker service restarts the process.

Also moved around the other test in this class, so that we do not have to create 2 workers in this class, rather than one.
Better code quality for the code that is responsible for restarting the worker.
```
source .e2e_windows_infra.sh
hatch run e2e-test
```
No
No

----

### DIFF
--- a/test/e2e/test_worker_status.py
+++ b/test/e2e/test_worker_status.py
@@ -17,31 +17,6 @@ LOG = logging.getLogger(__name__)
 
 @pytest.mark.parametrize("operating_system", [os.environ["OPERATING_SYSTEM"]], indirect=True)
 class TestWorkerStatus:
-    def test_worker_lifecycle_status_is_expected(
-        self,
-        deadline_resources,
-        deadline_client: DeadlineClient,
-        function_worker: EC2InstanceWorker,
-    ) -> None:
-        # Verifies that Worker Status returned by the GetWorker API is as expected when we start/stop workers
-
-        assert function_worker.worker_id is not None  # To fix linter type mismatch
-
-        assert is_worker_started(
-            deadline_client=deadline_client,
-            farm_id=deadline_resources.farm.id,
-            fleet_id=deadline_resources.fleet.id,
-            worker_id=function_worker.worker_id,
-        )
-
-        function_worker.stop_worker_service()
-
-        assert is_worker_stopped(
-            deadline_client=deadline_client,
-            farm_id=deadline_resources.farm.id,
-            fleet_id=deadline_resources.fleet.id,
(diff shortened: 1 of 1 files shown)


test/e2e/test_worker_status.py
```diff
@@ -130,3 +103,92 @@ def check_worker_processes_exist() -> None:
...
+            ), f"DeadlineWorker service is in unexpected status {service_check_result.stdout}"
+
+        check_service_is_running()
+
+        # Check that the worker process is running
+
+        def check_worker_processes_exist() -> None:
+            process_check_result = class_worker.send_command("Get-Process pythonservice")
+
+            assert process_check_result.exit_code == 0, "Worker agent process is not running"
```
hindleym, line 155: Should the quote for the error start with an `f`?

YutongLi291: Nope, f indicates f string which can be populated with variables. We do not need a variable inside of the string in this case.

test/e2e/test_worker_status.py
```diff
@@ -17,31 +17,6 @@
 
 @pytest.mark.parametrize("operating_system", [os.environ["OPERATING_SYSTEM"]], indirect=True)
 class TestWorkerStatus:
-    def test_worker_lifecycle_status_is_expected(
```
leongdl, line 20: This seems not related to the PR?

test/e2e/test_worker_status.py
```diff
@@ -130,3 +103,92 @@ def check_worker_processes_exist() -> None:
 
         # Check that there are worker processes running
         check_worker_processes_exist()
+
+    @pytest.mark.skipif(
+        os.environ["OPERATING_SYSTEM"] == "linux",
```
leongdl, line 108: This would fail if we were compiling on a MAC?

hindleym: We only have Linux and Windows worker agents.

test/e2e/test_worker_status.py
hindleym: Was noted with the PR that this is moved to reduce the number of workers needed.

hindleym (approved):

AlexTranAmz (approved):

PR #471, opened 2024-11-07
test: add test that verifies worker local session logs can be turned off

YutongLi291:
When specified, local worker session logs on the worker machine can be turned off with the environment variable "DEADLINE_WORKER_LOCAL_SESSION_LOGS" set to false.

We should make sure that this behaviour is as expected, that the logs are no longer written to the worker disk when specified.
Add a test that verifies that the local session logs will not exist when a job is created on a worker with this option off
Better testing for behaviour of local session logs writing of the worker.
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test
No
No

----

### DIFF
--- a/test/e2e/test_worker_config.py
+++ b/test/e2e/test_worker_config.py
@@ -1,0 +1,78 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""
+This test module contains tests that verify the Worker agent's behavior by creating the Worker with non-default configuration settings,
+and making sure that the behavior and outputs of the Worker is that of what we expect.
+"""
+import logging
+import os
+from typing import Any, Callable
+import pytest
+from deadline_test_fixtures import DeadlineClient, EC2InstanceWorker, DeadlineWorkerConfiguration
+import pytest
+import dataclasses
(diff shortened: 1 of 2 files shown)


test/e2e/test_worker_config.py
```diff
@@ -0,0 +1,81 @@
...
+                else:
+                    # Windows worker
+                    check_log_exists_result = worker_with_local_session_logs_off.send_command(
+                        command=f'Test-Path -Path "{session_logs_file_path}" -PathType leaf -Credential $Cred'
+                    )
+                    assert (
+                        "false" in check_log_exists_result.stdout.lower()
+                    ), f"Checking that local session logs do not exist returned unexpected response: {check_log_exists_result}"
+
+        check_session_logs_do_not_exist()
```
AlexTranAmz: Could you clarify why create a function and call it here? It seem to serve no purpose

YutongLi291: You're right, I initially had it backoff. I will fix this

AlexTranAmz (approved):

jericht (approved):

PR #476, opened 2024-11-13
feat: integrate with job attachment download cli as a openjd action run

godobyte:
Worker Agent runs sync_input and sync_output callbacks to interact with customer s3 as worker agent user. This brings the following problem
1. job user is a more suitable user to run the sync process
2. sync_input and sync_output handling in session polluted the purpose of the class

Worker Agent is in the process of changing Job Attachment sync input and sync output to use attachment and manifest commands running as job user.

This is still in development and is under a feature flag.

Locally set the feature flag, build and install whl files, run `hatch deadline-worker-agent`

It's currently only documented in the code docstring.

No

----


src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+            # =========================== TO BE DELETED ===========================
+            path_mapping_file_path: str = os.path.join(
+                session._session.working_directory, "path_mapping"
+            )
+            for rule in job_attachment_path_mappings:
+                rule["source_path"] = rule["destination_path"]
+
+            with open(path_mapping_file_path, "w", encoding="utf8") as f:
+                f.write(json.dumps([rule for rule in job_attachment_path_mappings]))
+            # =========================== TO BE DELETED ===========================
```
godobyte: Keeping it for now, pending changes and testing with attachment upload cli action integration.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+                        dir_mode=WindowsPermissionEnum.WRITE,
+                        file_mode=WindowsPermissionEnum.WRITE,
+                    )
+
+        if (
+            attachments.fileSystem == JobAttachmentsFileSystem.VIRTUAL.value
+            and sys.platform != "win32"
+            and fs_permission_settings is not None
+            and os.environ is not None
+            and "AWS_PROFILE" in os.environ
```
jusiskin, line 324: Is this condition necessary? I'm not sure if this is a requirement to launch the VFS.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+            ),
+            embeddedFiles=[
+                EmbeddedFileText_2023_09(
+                    name="AttachmentDownload",
+                    type=EmbeddedFileTypes_2023_09.TEXT,
+                    runnable=True,
+                    data=script,
+                )
+            ],
+        )
```
jusiskin, line 119: I may be missing something, but this looks like it is a single command being run in this script. Do we need the embedded file at all? Or can we just invoke the command directly in the `onRun` action?

Removing `bash` from the solution removes a hidden dependency and command injection issues.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+        super(AttachmentDownloadAction, self).__init__(
+            id=id,
+            action_log_kind=(
+                SessionActionLogKind.JA_SYNC
+                if step_details is None
+                else SessionActionLogKind.JA_DEP_SYNC
+            ),
+            step_id=step_details.step_id if step_details is not None else None,
+        )
+        self._cancel = Event()
```
jusiskin: This appears to be unused

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+
+        Parameters
+        ----------
+        session : Session
+            The Session that is the target of the action
+        executor : Executor
+            An executor for running futures
+        """
+
+        self._logger.info(f"Syncing inputs using session {session}")
```
jusiskin: The `openjd-sessions` library includes an `extra` in each log statement to classify the log record, for example:

https://github.com/OpenJobDescription/openjd-sessions-for-python/blob/fa8aa605c10222197961a7142a1fc73abe1e17b6/src/openjd/sessions/_action_filter.py#L159

This allows consumers of the logger to add conditional handling to the log records. I'm not certain, but we might want to do the same here to ensure the worker agent's log handling code doesn't omit these when deciding what to forward to CloudWatch.

Same feedback applies to the other logging statements below.

jusiskin (changes requested):

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+class SyncCanceled(Exception):
+    """Exception indicating the synchronization was canceled"""
+
+    pass
+
+
+class AttachmentDownloadAction(OpenjdAction):
+    """Action to synchronize input job attachments for a AWS Deadline Cloud job
```
leongdl: nit, for Deadline Cloud Session? Since a session has job and step-step dependency.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+        )
+        self._job_attachment_details = job_attachment_details
+        self._step_details = step_details
+        self._logger = LoggerAdapter(OPENJD_LOG, extra={"session_id": session_id})
+
+    def set_step_script(self, manifests, path_mapping, s3_settings) -> None:
+        profile = os.environ.get("AWS_PROFILE")
+        deadline_path = os.path.join(Path(sysconfig.get_path("scripts")), "deadline")
+
+        script = "#!/usr/bin/env bash\n\n{} attachment download -m {} --path-mapping-rules {} --s3-root-uri {}  --profile {}".format(
```
leongdl, line 99: nit, wdyt if the script line was a formatted multi-line string?
```
script = """
#!/usr/bin/env bash {} \
    attachment download \
    -m {} \
    --path-mapping-rules {} \
    --s3-root-uri {} \
    --profile {}"""
```

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+                queue_id=session._queue_id,
+                job_id=session._queue._job_id,
+                attachments=attachments,
+                step_dependencies=step_dependencies,
+                dynamic_mapping_rules=dynamic_mapping_rules,
+                storage_profiles_path_mapping_rules=storage_profiles_path_mapping_rules_dict,
+            )
+        )
+
+        vfs_handled = self._vfs_handling(
```
leongdl: nit, is this starting VFS? If yes, maybe `self._start_vfs(...`

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
leongdl: I actually think writing out a path mapping file is super useful for users? They can easily cat this file and use `jq` Or are we depending on the environment variable for path mapping for customers?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+                if not isinstance(session._os_user, PosixSessionUser):
+                    raise ValueError(f"The user must be a posix-user. Got {type(session._os_user)}")
+                fs_permission_settings = PosixFileSystemPermissionSettings(
+                    os_user=session._os_user.user,
+                    os_group=session._os_user.group,
+                    dir_mode=0o20,
+                    file_mode=0o20,
+                )
+            else:
+                if not isinstance(session._os_user, WindowsSessionUser):
```
leongdl, line 308: Maybe else block of code is not needed since VFS doesn't work on windows anyways?

Maybe an early exit if `sys.platform == win32 return False` ?

What about mac, would this cause any issues in this function?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Openjd session already writes the path mapping files, it's in a slightly different format tho, that's why I was thinking whether still keep this path mapping file or reuse the path written by openjd. Openjd would need to change to return the file path, and the attachment CLI would need slight change to read the following file.

```
{
    "version": "pathmapping-1.0",
    "path_mapping_rules": [{
        "source_path_format": "POSIX",
        "source_path": "/local/bea/bealine-clients/deadline-cloud-worker-agent/scripts/submit_jobs/asset_example",
        "destination_path": "/sessions/session-39d74575600f47eb80d31d440721a27b7mrqf9ip/assetroot-a7714e87e776d9f1c179"
    }]
}
```

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: This is a move of existing code, I'm trying to keep it as-is since this part is not in scope of the change. I can probably remove some of the conditions, but that should be tested out with the VFS setup later.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Good catch, removed in the rev.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: I'd like to keep it as-is for now and convert it to python as embedded files later. That should come after the attachment upload PR and along with other potential windows change.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: This is a move of existing code, I'm trying to keep it as-is since this part is not in scope of the change. I was thinking about improving it, but that would be safer with the VFS setup and tests later.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: I tried it, and the test started to fail. I wouldn't worry about it too much since it will be replaced by a python embedded file.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Updated to add the extra info log content, the banner probably needs more work to take in the title from client, but that'd be for another day.

src/deadline_worker_agent/api_models.py
```diff
@@ -46,6 +46,7 @@
 EnvironmentActionType = Literal["ENV_ENTER", "ENV_EXIT"]
 StepActionType = Literal["TASK_RUN"]  # noqa
 SyncInputJobAttachmentsActionType = Literal["SYNC_INPUT_JOB_ATTACHMENTS"]  # noqa
+AttachmentDownloadActionType = Literal["SYNC_INPUT_JOB_ATTACHMENTS"]  # noqa
```
jusiskin, line 49: nit: these are redundant.

This file is meant to provide type-hinting for our service model API schema. The service model API schema has not changed, so IMO we do not need `AttachmentDownloadActionType`. Instead we can just use `SyncInputJobAttachmentsActionType`.

src/deadline_worker_agent/api_models.py
```diff
@@ -87,6 +88,12 @@ class SyncInputJobAttachmentsAction(TypedDict):
     stepId: NotRequired[str]
 
 
+class AttachmentDownloadAction(TypedDict):
+    sessionActionId: str
+    actionType: AttachmentDownloadActionType
+    stepId: NotRequired[str]
```
jusiskin, line 94: Same feedback as `AttachmentDownloadActionType`. We can just use `SyncInputJobAttachmentsAction`.

src/deadline_worker_agent/boto/shim.py
```diff
@@ -187,14 +206,21 @@ def parse_sync_input_job_attachments_action(
                 environmentId=action["environmentId"],
             ),
             "taskRun": parse_task_run_action,
-            "syncInputJobAttachments": parse_sync_input_job_attachments_action,
+            "syncInputJobAttachments": (
+                parse_sync_input_job_attachments_action
+                if not ASSET_SYNC_JOB_USER_FEATURE
```
jusiskin, line 211: `ASSET_SYNC_JOB_USER_FEATURE` is currently defined in `src/deadline_worker_agent/feature_flag.py` as:

```py
ASSET_SYNC_JOB_USER_FEATURE = os.environ.get("ASSET_SYNC_JOB_USER_FEATURE")
```

This will have a `str` value. Based on how Python evaluates truth values of strings ([docs](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)), any non-empty string will be evaluated to true. Is that what we want? For example:

```sh
$ export ASSET_SYNC_JOB_USER_FEATURE=false
# Will treat the above as having the feature turned on
$ deadline-worker-agent ...
```

Do we care about this? Should there be an allow-listed set of values? Or perhaps a case-insensitive string matching "true"?

src/deadline_worker_agent/feature_flag.py
```diff
@@ -0,0 +1,5 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import os
+
+ASSET_SYNC_JOB_USER_FEATURE = os.environ.get("ASSET_SYNC_JOB_USER_FEATURE")
```
jusiskin: Should we document these feature flags? We should at least include a code comment here, but maybe we can link to this file from `DEVELOPMENT.md` for contributors who want to discover or build their own feature flags. Maybe we can also indicate the level of stability for the feature. This feature is still in-progress (e.g. untested on Windows).

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,336 @@
...
+        else:
+            section_title = "Job Attachments Download for Job"
+
+        # Banner mimicing the one printed by the openjd-sessions runtime
+        self._logger.info(
+            "==============================================",
+            extra={"openjd_log_content": LogContent.BANNER},
+        )
+        self._logger.info(
+            f"--------- AttachmentDownloadAction  {section_title}",
```
jusiskin, line 157: Are we planning to remove the `AttachmentDownloadAction` in the banner at a later time? I suspect this is added here to help developers debug and confirm that the feature flag is working as expected.

If we're planning to remove later, can we add a TODO comment?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,336 @@
...
+        # Validate that job attachment details have been provided before syncing step dependencies.
+        if session._job_attachment_details is None:
+            raise RuntimeError(
+                "Job attachments must be synchronized before downloading Step dependencies."
+            )
+        step_dependencies = self._step_details.dependencies if self._step_details else []
+
+        assert job_attachment_settings.s3_bucket_name is not None
+        assert job_attachment_settings.root_prefix is not None
+        assert session._asset_sync is not None
```
jusiskin, line 180: Two things to discuss:

1.  We are assigning to another class' private attributes which breaks object oriented encapsulation.
2.  This session action accepts a `job_attachment_settings` in its constructor but also tries to obtain it from the session

I wonder if the simpler approach would be for the `SessionActionQueue` to always provide the job attachment settings in the constructor arguments to `AttachmentDownloadAction` (make it a required argument). There is less complexity and we avoid setting private attributes of the `Session` class.

Under the hood, the `SessionActionQueue` is constructed with a reference to a `JobEntities` class which handles caching of job entities, so we can achieve the same performance optimization of storing the job attachment settings as an attribute of the `Session` class and simplify this code.

What do you think?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,336 @@
...
+        )
+
+        vfs_handled = self._start_vfs(
+            session=session,
+            attachments=attachments,
+            merged_manifests_by_root=merged_manifests_by_root,
+            s3_settings=s3_settings,
+        )
+
+        if not vfs_handled:
```
jusiskin: nit: this is somewhat stylistic, but in this case I'd suggest an early return here. In the current form, there is a lot of indented code in the `if not vfs_handled` block.

What do you think of switching to:

```py
if self._start_vfs(...):
    return

# Code for COPIED mode
```

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: The path to the file can be provided by command-line argument by using the `Session.PathMappingRulesFile` string interpolation expression ([docs](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#73-format-strings)) in the step script.

It does require that the job attachments CLI accept this command-line argument and understand the file format ([docs](https://github.com/OpenJobDescription/openjd-specifications/wiki/How-Jobs-Are-Run#path-mapping))

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: Checked and the existing code has this condition as well. We can keep this out-of-scope.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: One alternative to using an embedded file is to bundle the python script in the worker agent code package. We can then specify the absolute path to this script in the step script.

That avoids maintaining the Python script inside a Python string literal in our code. What do you think?

test/unit/sessions/actions/test_run_attachment_download.py
```diff
@@ -0,0 +1,182 @@
...
+        Tests that AttachmentDownloadAction.start() calls AssetSync functions to prepare input
+        for constructing step script to run openjd action
+        """
+        # WHEN
+        action.start(session=session, executor=executor)
+
+        # THEN
+        assert job_details.job_attachment_settings is not None
+        assert job_details.job_attachment_settings.s3_bucket_name is not None
+        assert job_details.job_attachment_settings.root_prefix is not None
```
jusiskin, line 133: nit: these could be moved up to the `GIVEN` block

test/unit/sessions/actions/test_run_attachment_download.py
```diff
@@ -0,0 +1,182 @@
...
+            s3_settings=JobAttachmentS3Settings(
+                s3BucketName=job_details.job_attachment_settings.s3_bucket_name,
+                rootPrefix=job_details.job_attachment_settings.root_prefix,
+            ),
+            queue_id=TestStart.QUEUE_ID,
+            job_id=TestStart.JOB_ID,
+            attachments=ANY,
+            step_dependencies=[],
+            dynamic_mapping_rules=ANY,
+            storage_profiles_path_mapping_rules={},
```
jusiskin, line 149: Do we need to parametrize these to ensure that the behavior changes depending on the inputs? As it stands right now, these parameters don't seem to be adequately covered.

Same applies to the expected calls below.

jusiskin (changes requested):
Great work here! I have a few suggested improvements. Happy to have a discussion on which ones we want to adopt.

src/deadline_worker_agent/api_models.py
godobyte: I'd like a clean cut between the `SyncInputJobAttachmentsAction` and `AttachmentDownloadAction`. Once the new feature is released, all `SyncInputJobAttachmentsAction` related code will be cleaned up. This `AttachmentDownloadAction` is also corresponding to the  upcoming`AttachmentUploadAction`.

src/deadline_worker_agent/api_models.py
godobyte: Same as the above, IMO introducing this new action would save effort for cutover and cleanup.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: > I wonder if the simpler approach would be for the SessionActionQueue to always provide the job attachment settings in the constructor arguments to AttachmentDownloadAction (make it a required argument).

That's exactly why creating a new `AttachmentDownloadAction` is a better approach than using the existing `SyncInputJobAttachmentsAction` ;). It'd give more flexibility.

I'll follow up offline on this one, we need to make sure step-dependency handling works if going with passing from action queue approach.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: I think that's a good idea if it fits the data flow, need to make sure the `Session.PathMappingRulesFile` can be used for both attachment download and attachment upload. Let's chat more offline.

src/deadline_worker_agent/boto/shim.py
godobyte: I didn't think we care about this much since it's supposed to be short-lived. With that said, this is a easy enough improvement and I can quickly incorporate. Addressed in `src/deadline_worker_agent/feature_flag.py` below.

src/deadline_worker_agent/feature_flag.py
godobyte: I'll document in this file for now and add to `DEVELOPMENT.md` later if needed.

test/unit/sessions/actions/test_run_attachment_download.py
godobyte: Synced offline, we will revisit it after addressing few other priorities.

src/deadline_worker_agent/api_models.py
jusiskin: So after looking again, I see some deviation in `api_models.py` from our service model schema For example, we have `SYNC_INPUT_JOB_ATTACHMENTS` here vs. `syncInputJobAttachments` in our API.

I think this may be an artifact from earlier revisions of our service model that have not been updated in the code.
Longer term, I'd like to consolidate these types and have them model our service model API shape, but I'm okay with leaving this out-of-scope for this PR.

test/unit/sessions/actions/test_run_attachment_download.py
jusiskin: Discussed offline and we will postpone this work

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Synced offline, encountered file not found issue trying to directly invoke the python script, and having the python file as embedded file worked. In following PRs, will use a static file for dev/build tooling and then read it at runtime and inject it into an embedded file.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: Discussed offline. We decided to address this cleanup in a future PR and merge this as-is for now.

jusiskin (approved):

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
leongdl: I see, ok as long as this is debuggable.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -0,0 +1,338 @@
...
+                OpenjdPathMapping.from_dict(r) for r in job_attachment_path_mappings
+            ]
+
+        # Open Job Description Sessions sort the path mapping rules based on length of the parts make
+        # rules that are subsets of each other behave in a predictable manner. We must
+        # sort here since we're modifying that internal list appending to the list.
+        session._session._path_mapping_rules.sort(key=lambda rule: -len(rule.source_path.parts))
+
+        # =========================== TO BE DELETED ===========================
+        path_mapping_file_path: str = os.path.join(
```
leongdl, line 263: nit this will be deleted later right?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
leongdl: ok fair enough as a follow up.

leongdl (approved):

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Yes. In order to delete this, I'd need to do some additional changes, but that'd be part of this feature completion.

PR #477, opened 2024-11-14
test: add linux job user override tests

YutongLi291:
Users are able to boot up the linux worker agent with custom job override users rather than using the default job-user. This will make sure that jobs are ran with the custom overridden user.

We should make sure this  functionality works as expected and there are no regressions.

Add WA E2E test cases that verify that the job user can be overridden through the config or the env variable on Linux

Prevention of regressions in functionality allowing customers to override the job user.

```
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test
```

`31 passed, 6 skipped, 5 warnings in 3099.93s (0:51:39)`
Tested over 10 times to verify that it is not flaky.
No
No
----


test/e2e/conftest.py
```diff
@@ -385,6 +387,25 @@ def job_run_as_user() -> PosixSessionUser:
...
+        ),
+        PosixSessionUser(
+            user="config-override",
+            group="job-override-group",
+        ),
+        PosixSessionUser(
+            user="env-override",
+            group="job-override-group",
+        ),
+    ]
```
jericht: I think it makes more sense to split this up. We should be able to select each user without having to memorize/hardcode the `user`/`group` of that user in the test code.

test/e2e/test_override_job_user.py
```diff
@@ -234,3 +237,209 @@ def test_env_var_user_override(
...
+            )
+
+            assert worker_status_cmd_response.exit_code != 0
+            assert worker_status_cmd_response.stdout != "active"
+
+        check_worker_service_stopped()
+
+        cmd_result = class_worker.send_command(
+            'sed -i \'s/# posix_job_user = "user:group"/posix_job_user = "config-override:job-override-group"/g\' /etc/amazon/deadline/worker.toml'
+        )
```
jericht: Follow-up from my comment about the `posix_job_users` fixture, for example this code could use the user fixture directly rather than assuming the user/group exists on the worker
```suggestion
            f'sed -i \'s/# posix_job_user = "user:group"/posix_job_user = "{override_user.user}:{override_user.group}"/g\' /etc/amazon/deadline/worker.toml'
        )
```

test/e2e/test_override_job_user.py
```diff
@@ -234,3 +237,209 @@ def test_env_var_user_override(
...
+                config=botocore.config.Config(retries={"max_attempts": 10, "mode": "adaptive"}),
+            ),
+            expected_pattern="I am: config-override",
+        )
+
+        assert job.task_run_status == TaskStatus.SUCCEEDED
+
+        cmd_result = class_worker.send_command(
+            'sed -i \'s/posix_job_user = "config-override:job-override-group"/# posix_job_user = "user:group"/g\' /etc/amazon/deadline/worker.toml'
+        )
```
jericht: nit: why not just delete the line?

test/e2e/test_override_job_user.py
```diff
@@ -234,3 +237,209 @@ def test_env_var_user_override(
...
+
+        assert job.task_run_status == TaskStatus.SUCCEEDED
+
+        cmd_result = class_worker.send_command(
+            "sed -i '/Environment=DEADLINE_WORKER_POSIX_JOB_USER=env-override/d' /etc/systemd/system/deadline-worker.service.d/config.conf"
+        )
+        assert (
+            cmd_result.exit_code == 0
+        ), f"Resetting the job user override via CLI failed: {cmd_result}"
+        class_worker.send_command("sudo systemctl daemon-reload")
```
jericht: We need to do this because this worker is reused in future tests right? Maybe this should be in a `finally` block in case something goes wrong in the test

test/e2e/conftest.py
YutongLi291: What do you mean split this up? Make each user into a different fixture?

test/e2e/test_override_job_user.py
YutongLi291: This worker would be used if any future tests are added to this class. I can make it into a finally block.

jericht (approved):

AlexTranAmz (approved):

PR #479, opened 2024-11-15
feat: more robust Linux cross-user cancelation under low memory conditions

jusiskin:

Under low memory situations on Linux, the worker agent can fail to cancel a running session action.

On Linux, processes can only send signals to other processes if at least one of the following are true:

*   the target process is owned by the same OS user as the signalling process OR
*   if the process sending the signal
    *   is `root`
    *   has the `CAP_KILL` capability (see[capabilities(7) man page](https://man7.org/linux/man-pages/man7/capabilities.7.html))

To overcome this limitation, `openjd-sessions` used `sudo` to create a `bash` process as the target user which then sends the OS signal. When the system has low memory, there may not be sufficient memory to create the subprocesses successfully.

OpenJobDescription/openjd-sessions-for-python#196 is a work-in-progress enhancement to leverage `CAP_KILL` and send signals to cross-user processes without creating subprocesses.

This PR builds on this support for `CAP_KILL` so the worker agent directly signals cross-user processes.

1.  The code for `install-deadline-worker` was modified to configure the worker agent systemd unit with the `CAP_KILL` ambient capability. This has the effect of adding `CAP_KILL` to the worker agent process' permitted/effective/inheritable capability sets.
2. The worker agent now drops `CAP_KILL` from the inheritable capability set early in the program startup so that other threads and subprocesses will not inherit this privileged capability.
3. The `openjd-session` change takes care of the rest and handles direct signalling for cross-user processes.

This change allows the worker agent to cancel session actions more robustly in low-memory situations.

The cross-user direct signal cancelation tests live in `openjd-sessions`, but a security end-to-end test was added to ensure that session actions are not able to signal other processes.

No

No

----

### DIFF
--- a/src/deadline_worker_agent/installer/install.sh
+++ b/src/deadline_worker_agent/installer/install.sh
@@ -456,15 +456,28 @@ Environment=AWS_REGION=$region AWS_DEFAULT_REGION=$region
 EOF
     fi
     
+    ###############################################################
+    ###############    NOTE FOR CODE REVIEWERS    #################
+    ###############################################################
+    #    Review changes to AmbientCapabilities below carefully    #
+    ###############################################################
+
     cat >> /etc/systemd/system/deadline-worker.service <<EOF   
 ExecStart=$worker_agent_program
 Restart=on-failure
 StandardOutput=null
 StandardError=null
+AmbientCapabilities=CAP_KILL
 
 [Install]
 WantedBy=multi-user.target
 EOF
+    ###############################################################
+    ###############    NOTE FOR CODE REVIEWERS    #################
+    ###############################################################
+    #    Review changes to AmbientCapabilities above carefully    #
+    ###############################################################
+
     chown root:root /etc/systemd/system/deadline-worker.service
     chmod 600 /etc/systemd/system/deadline-worker.service
     echo "Done installing systemd service"
--- a/src/deadline_worker_agent/linux/capabilities.py
+++ b/src/deadline_worker_agent/linux/capabilities.py
@@ -1,0 +1,236 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""This module contains code for interacting with Linux capabilities.
+
+See https://man7.org/linux/man-pages/man7/capabilities.7.html for details on this Linux kernel
+feature.
+"""
+
+import ctypes
+import os
+import sys
+from logging import getLogger
+from ctypes.util import find_library
+from functools import cache
+from typing import Any, Optional, Tuple, TYPE_CHECKING
+
+
+logger = getLogger(__name__)
+
+
+# Capability sets
+# See https://ddnet.org/codebrowser/include/sys/capability.h.html#cap_flag_t
+CAP_EFFECTIVE = 0
+CAP_PERMITTED = 1
+CAP_INHERITABLE = 2
+
+# Capability bit numbers
+# See https://github.com/torvalds/linux/blob/28eb75e178d389d325f1666e422bc13bbbb9804c/include/uapi/linux/capability.h#L147
+CAP_KILL = 5
+
+# Values for cap_flag_value_t arguments
+# See https://ddnet.org/codebrowser/include/sys/capability.h.html#cap_flag_value_t
+CAP_CLEAR = 0
+CAP_SET = 1
+
+cap_flag_t = ctypes.c_int
+cap_flag_value_t = ctypes.c_int
+cap_value_t = ctypes.c_int
+
+
+class UserCapHeader(ctypes.Structure):
+    _fields_ = [
+        ("version", ctypes.c_uint32),
+        ("pid", ctypes.c_int),
+    ]
+
+
+class UserCapData(ctypes.Structure):
+    _fields_ = [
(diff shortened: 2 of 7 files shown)


src/deadline_worker_agent/installer/install.sh
```diff
@@ -456,15 +456,28 @@ Environment=AWS_REGION=$region AWS_DEFAULT_REGION=$region
 EOF
     fi
     
+    ###############################################################
+    ###############    NOTE FOR CODE REVIEWERS    #################
+    ###############################################################
+    #    Review changes to AmbientCapabilities below carefully    #
+    ###############################################################
+
```
ddneilson, line 464: Nice touch. I like it!

src/deadline_worker_agent/linux/capabilities.py
```diff
@@ -0,0 +1,215 @@
...
+        raise OSError(errno, os.strerror(errno))
+    return result
+
+
+@cache
+def _get_libcap() -> ctypes.CDLL:
+    if not sys.platform.startswith("linux"):
+        raise OSError(f"libcap is only available on Linux, but found platform: {sys.platform}")
+
+    libcap = ctypes.CDLL(find_library("cap"), use_errno=True)
```
ddneilson: Same issue as the `openjd-sessions` PR. We're assuming that the lib is always present. We modified the other change to account for it not being present.

test/e2e/test_cap_kill.py
```diff
@@ -0,0 +1,113 @@
...
+    # GIVEN
+    # Send SSM command to write and run a bash script
+    # The script creates a detached sleep process and outputs that process' PID
+    # This sleep process will run as the ssm-user which is different from the job user
+    result = session_worker.send_command(
+        " ; ".join(
+            [
+                "echo '#!/bin/bash' > script.sh",
+                "echo 'set -euo pipefail' >> script.sh",
+                "echo 'nohup sleep 120 < /dev/null 2> /dev/null > /dev/null &' >> script.sh",
```
ddneilson: Two questions:
1. Is 120s sufficient? I'm thinking about potential test timing. e.g. if it, for some reason, takes longer than expected for the worker to pick up and run the submitted job.
2. Should we have an ssm command at the end of the test that explicitly kills the script, just to make sure that we're not leaving anything behind? There's *probably* no harm, right now, in the test leaving a script running for a short time, but if this becomes a pattern that is replicated then the replicant may clash with this test in random unexpected ways...

ddneilson (changes requested):

ddneilson:
Oh, also, the README should probably be updated to list the (optional) dependency on `libcap.so` on linux

src/deadline_worker_agent/linux/capabilities.py
jusiskin: I've added error handling for when `libcap` cannot be found. On Amazon Linux, it appears that systemd depends on libcap:

```sh
$ dnf repoquery -q --installed --whatrequires libcap
chrony-0:4.3-1.amzn2023.0.4.x86_64
coreutils-0:8.32-30.amzn2023.0.3.x86_64
iproute-0:6.10.0-319.amzn2023.0.1.x86_64
iputils-0:20210202-2.amzn2023.0.4.x86_64
kernel-tools-0:6.1.115-126.197.amzn2023.x86_64
libcap-devel-0:2.48-2.amzn2023.0.3.x86_64
nfs-utils-1:2.5.4-2.rc3.amzn2023.0.3.x86_64
rpm-build-libs-0:4.16.1.3-29.amzn2023.0.6.x86_64
rpm-libs-0:4.16.1.3-29.amzn2023.0.6.x86_64
rpm-plugin-selinux-0:4.16.1.3-29.amzn2023.0.6.x86_64
rpm-plugin-systemd-inhibit-0:4.16.1.3-29.amzn2023.0.6.x86_64
rpm-sign-libs-0:4.16.1.3-29.amzn2023.0.6.x86_64
systemd-0:252.23-2.amzn2023.x86_64
systemd-libs-0:252.23-2.amzn2023.x86_64
systemd-pam-0:252.23-2.amzn2023.x86_64
```

In the ["Worker host" section Deadline Cloud user guide](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/worker-host.html), we recommend using Amazon Linux:

>  To complete the tasks in the following sections, you need the following:
>   *   A Linux-based Amazon Elastic Compute Cloud (Amazon EC2) instance. We recommend Amazon Linux 2023.

systemd is used to launch the worker agent on boot and restart on a crash. We also configure the worker agent to be given `CAP_KILL` in the systemd unit generated by `install-deadline-worker`.

On this basis, we decided to log a warning when the library is not found and move on. Those running the worker agent on Linux with `CAP_KILL` but without `libcap` available is not officially supported.

test/e2e/test_cap_kill.py
jusiskin: I've bumped the sleep to 240s which should be sufficient. I've also moved the background sleep job to a pytest fixture which handles cleanup of the process

AWS-Samuel (approved):

ddneilson (approved):
Thanks for taking this on, Josh. Looks great!

PR #480, opened 2024-11-15
chore!: Remove offical support for python 3.13

amarsjac:
Dependencies required for the project do not exist for python 3.13
put in version cap into pyproject.toml file to prevent pip installing on python 3.13
Users will no longer be able to pip install the library if python version is >=3.13
```
hatch build
```
no
Yes , prevents python 3.13 from pip installing.
----


leongdl (approved):

jusiskin:
Closing this PR. I suspect the dependency that had not supported Python 3.13 was pywin32, however that now supports Python 3.13.

PR #481, opened 2024-11-16
test: add test for worker that requires no instance profile

YutongLi291:

The worker agent is able to be installed with the option `disallow-instance-profile`, in which case if the worker instance is booted with an instance profile, the worker agent will shut itself down and not run jobs.

We should verify that this functionality works as expected and that there are no regressions

Add a test to verify that the worker will not pick up any jobs in the case that the worker is configured to not have instance profiles, but the instance has a profile anyway.

Also upgraded test-fixtures version from 0.16 to 0.17 to allow for the option to disallow instance profile.
Better verification of worker agent installer code and behaviour.
```
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test

```
No
No

----


AlexTranAmz (approved):

jericht (approved):

test/e2e/test_worker_config.py
```diff
@@ -21,6 +22,64 @@
...
+        worker_config: DeadlineWorkerConfiguration,
+        function_worker_factory: Callable[[DeadlineWorkerConfiguration], EC2InstanceWorker],
+    ) -> None:
+
+        # Create a EC2 worker with disallow-instance-profiles option for the worker agent
+        # Note that the EC2 instance is created with an instance profile, so no job will ever be picked up by this worker
+        function_worker_factory(
+            dataclasses.replace(
+                worker_config,
+                disallow_instance_profile="True",
```
godobyte, line 38: Is it expected to cause lint error?

```
test/e2e/test_worker_config.py:33: error: Unexpected keyword argument
"disallow_instance_profile" for "replace" of "DeadlineWorkerConfiguration"  [call-arg]
                dataclasses.replace(
                ^
Found 1 error in 1 file (checked 145 source files)
```

godobyte: I see it's due to an dependent change

https://github.com/aws-deadline/deadline-cloud-test-fixtures/blob/mainline/src/deadline_test_fixtures/deadline/worker.py#L96

PR #486, opened 2024-11-18
test: re-enable linux job user override tests

YutongLi291:
The linux job user override tests were failing in the mainline canaries as the canaries had separate workers that were not cleaned up properly in the fleet. This caused the job user override tests to fail as the jobs were submitted to workers that did not have the override.

These tests were disabled for the meanwhile to investigate. We can re-enable them after the leftover worker instances have been terminated.

Re-enable these tests
The tests should be re-enabled and work properly.
```

source .e2e_linux_infra.sh
hatch run e2e-test

```
No
No
----

### DIFF
--- a/test/e2e/test_override_job_user.py
+++ b/test/e2e/test_override_job_user.py
@@ -318,9 +318,6 @@ class TestLinuxJobUserOverride:
 
         assert job.task_run_status == TaskStatus.SUCCEEDED
 
-    @pytest.mark.skip(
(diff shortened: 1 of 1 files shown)


leongdl (approved):

jericht (approved):
Approving for now, but...

Could we do a more thorough cleanup? e.g. keep track of all workers deployed in the test session then at the end make sure all are terminated, if we're not doing that already

PR #487, opened 2024-11-19
test: add test to verify queue credentials are locked down

YutongLi291:
Queue credentials on the worker should only be able to be able to be accessed by the queue user.

We should verify this to prevent security concerns and regressions in the future.

Add a test that verifies this behaviour that other users on the worker instance cannot access the queue credentials.

Also verify that a different queue also cannot access another queue's credentials.
Better verification that other users cannot access queue credentials.
```
# Linux
source .e2e_linux_infra.sh
hatch run e2e-test

# Windows
source .e2e_windows_infra.sh
hatch run e2e-test

```

Multiple times to verify not flaky.
No
No
----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -63,6 +63,107 @@ class TestJobSubmission:
 
     @pytest.mark.skipif(
         os.environ["OPERATING_SYSTEM"] == "windows",
+        reason="Linux specific queue crendentials test",
+    )
+    def test_queue_credentials_file_is_secure(
+        self,
+        deadline_resources,
+        session_worker: EC2InstanceWorker,
+        posix_job_user: PosixSessionUser,
+        generic_non_queue_job_user: PosixSessionUser,
+        deadline_client: DeadlineClient,
+    ) -> None:
+        # Test to verify that the queue credentials can never be accessed by a different queue or user
+
+        job = submit_custom_job(
+            "Test Sleep",
+            deadline_client,
+            deadline_resources.farm,
+            deadline_resources.queue_a,
+            """
+            #!/usr/bin/env bash
+            sleep 90
+            """,
+        )
+
+        try:
+
+            @backoff.on_predicate(
+                wait_gen=backoff.constant,
+                max_time=120,
+                interval=10,
+            )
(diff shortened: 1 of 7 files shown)


AlexTranAmz (dismissed):

AlexTranAmz (dismissed):

test/e2e/test_job_submissions.py
```diff
@@ -61,6 +61,107 @@ def test_success(
...
+        # Test to verify that the queue credentials can never be accessed by a different queue or user
+
+        job = submit_custom_job(
+            "Test Sleep",
+            deadline_client,
+            deadline_resources.farm,
+            deadline_resources.queue_a,
+            """
+            #!/usr/bin/env bash
+            sleep 90
```
jericht, line 85: Is 90 seconds long enough? Can we have the job just wait forever?

test/e2e/test_job_submissions.py
```diff
@@ -61,6 +61,107 @@ def test_success(
...
+                assert second_queue_job.task_run_status == TaskStatus.FAILED
+
+            finally:
+                deadline_client.update_job(
+                    farmId=second_queue_job.farm.id,
+                    queueId=second_queue_job.queue.id,
+                    jobId=second_queue_job.id,
+                    targetTaskRunStatus="CANCELED",
+                )
+                second_queue_job.wait_until_complete(client=deadline_client)
```
jericht, line 207: Can this be made into a separate test case? These tests take a long time to run so it would be great if our assertions/test cases are more fine grained, especially if we need to submit another job to run this test case anyway

test/e2e/test_job_submissions.py
YutongLi291: Yeah it's long enough for these checks. I opted not to have it wait forever as there's always the possibility that the job cancel fails and the job runs forever, which is not ideal.

test/e2e/test_job_submissions.py
YutongLi291: Makes sense, okay, I'll add a separate test case.

jericht (approved):

AlexTranAmz (approved):

PR #492, opened 2024-11-21
test: simplify test fixtures

jusiskin:
Depends on release of aws-deadline/deadline-cloud-test-fixtures#164

1.  The worker agent end-to-end tests had overridden the `worker_config` fixture from `deadline-cloud-test-fixtures` to work around bugs and enhance its functionality. This bug has been fixed upstream (see aws-deadline/deadline-cloud-test-fixtures#164)

2.  The worker agent end-to-end tests had used a combination of:

    ```py
    pytest.mark.parametrize("operating_system", ["linux"], indirect=True)
    pytest.mark.skipif(os.environ["OPERATING_SYSTEM"] == "windows")
    ```

    These are redundant. We should not need to use both of the above for each test case.
    
1. Modify the `worker_config` fixture override to inherit the value from `deadline-cloud-test-fixture` but modify only the parts necessary
2. Modify the `operating_system` fixture to take its value from the `OPERATING_SYSTEM` environment variable

1.  There is less duplication of the `worker_config` fixture from the upstream `deadline-cloud-test-fixtures` package and we do not have to maintain synchronization of their logic
2.  Tests are simpler to write and do not need a combination of pytest parametrization and conditional skipping. Test authors only need to use test skipping

1.  Setup the environment with `hatch env create`
2. Installed the `mainline` version of `deadline-cloud-test-fixtures` (includes aws-deadline/deadline-cloud-test-fixtures#164) into the hatch env:

    ```sh
    $ hatch shell
    (deadline-cloud-worker-agent) $ pip install git+https://github.com/aws-deadline/deadline-cloud-test-fixtures.git
    ```
4.  Ran the end-to-end tests.
    *   Results pending...

Yes, updated `test/e2e/README.md` with updated documentation for writing os-specific tests.

No

----

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -1,23 +1,26 @@
 # Cross OS E2E Tests
 
-Tests designed to work with both Linux and Windows workers should use the `operating_system` parameter to differentiate differences between operating systems.
+Tests designed to cover Linux and Windows workers should use the `operating_system` fixture to for conditional testing logic between operating systems. The fixture uses the value of the `OPERATING_SYSTEM` environment variable and its value should be either `linux` or `windows`.
 
-Export the `OPERATING_SYSTEM` environment variable to either "linux" or "windows", depending on the OS of the worker that you would like to test.
+The `OPERATING_SYSTEM` environment variable is setup both in CI configuration and developer scripting (see [Running Worker Agent E2E Tests](../../DEVELOPMENT.md#running-worker-agent-e2e-tests)).
 
 # OS Specific Tests
-Use the `mark.skipif` parameter to differentiate between operating systems for os specific tests.
-```
+Use the `mark.skipif` parameter to differentiate between operating systems for os specific tests. Decorators are evaluated at import time and cannot use fixtures for their conditions.Instead, use the `OPERATING_SYSTEM` environment variable:
+
+```py
+# Linux-specific test
 @pytest.mark.skipif(
-        os.environ["OPERATING_SYSTEM"] == "windows",
-        reason="Linux specific test",
-    )
+    os.environ["OPERATING_SYSTEM"] != "linux",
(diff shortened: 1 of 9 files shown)


moorec-aws (dismissed):

test/e2e/README.md
```diff
@@ -1,23 +1,26 @@
...
-```
+Use the `mark.skipif` parameter to differentiate between operating systems for os specific tests. Decorators are evaluated at import time and cannot use fixtures for their conditions.Instead, use the `OPERATING_SYSTEM` environment variable:
+
+```py
+# Linux-specific test
 @pytest.mark.skipif(
-        os.environ["OPERATING_SYSTEM"] == "windows",
-        reason="Linux specific test",
-    )
+    os.environ["OPERATING_SYSTEM"] == "windows",
```
AWS-Samuel:
```suggestion
    os.environ["OPERATING_SYSTEM"] != "linux",
```

Nit, would a negative check be better here if we, say, one day supported another OS like mac?

Applies below to windows as well

test/e2e/conftest.py
```diff
@@ -146,74 +137,11 @@ def worker_config(
...
-            file_mappings=file_mappings or None,
-            job_users=[posix_job_user, posix_config_override_job_user, posix_env_override_job_user],
-            windows_job_users=windows_job_users,
-            start_service=True,
-        )
+    return dataclasses.replace(
+        worker_config,
+        job_users=[posix_job_user, posix_config_override_job_user, posix_env_override_job_user],
+        windows_job_users=windows_job_users,
+    )
```
AWS-Samuel, line 144: nice!

AWS-Samuel (dismissed):

AWS-Samuel (approved):

moorec-aws (approved):

PR #495, opened 2024-11-27
feat: run job attachment output upload as job user

godobyte:
Worker Agent runs sync_input and sync_output callbacks to interact with customer s3 as worker agent user. This brings the following problem

1.  job user is a more suitable user to run the sync process
2. sync_input and sync_output handling in session polluted the purpose of the class

Worker Agent is in the process of changing Job Attachment sync input and sync output to use attachment and manifest commands running as job user.

A sequel to https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/476

This is still in development and is under a feature flag.

1. unit tests
2. Locally set the feature flag, build and install whl files, run `hatch deadline-worker-agent`

It's currently only documented in the code docstring.

Not yet

----


test/unit/sessions/actions/test_run_attachment_download.py
```diff
@@ -12,6 +12,7 @@
 import pytest
 
 from deadline_worker_agent.sessions.actions import AttachmentDownloadAction
+import deadline_worker_agent.sessions.actions as actions_module
```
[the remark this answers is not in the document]
godobyte: Will address in the next revision.

src/deadline_worker_agent/scheduler/session_queue.py
```diff
@@ -287,6 +293,37 @@ def cancel_all(
...
+        self,
+        *,
+        action: AttachmentUploadActionApiModel,
+    ) -> None:
+        """Inserts an attachmen upload action at the front of the queue
+
+        Parameters
+        ----------
+        action : SessionActionQueueEntry
+            The action to be inserted
```
jusiskin: nit: the type documented here mismatches the type hint of the method

src/deadline_worker_agent/scheduler/session_queue.py
```diff
@@ -287,6 +293,37 @@ def cancel_all(
...
+        """
+        action_type = action["actionType"]
+        action_id = action["sessionActionId"]
+        logger.info(
+            "Inserting attachment upload to the front of queue: %s action: %s",
+            action_type,
+            action_id,
+        )
```
jusiskin: I wonder if we need both or any of these log statements at `INFO` level. 

While not consistently applied in the worker agent, here are some guidelines for logging.

*   **Log before the action** &mdash; if it is expected to be cpu/memory-intensive. This is to provide visibility into what is happening if there is no logging for a significant period or if the program crashes (e.g. due to being out of memory).
*   **Log after the action** &mdash; to communicate the outcome (e.g. success vs. failure) to readers of the logs.

With regards to the logging levels:

| Level | When to use |
| --- | --- |
| `DEBUG` | If the log event is useful for development or debugging |
| `INFO` | If the log event is useful for regular usage |
| `WARNING` | If the log event communicates a potential issue  |
| `ERROR` | If an important operation was unsuccessful |
| `CRITICAL` | If a critical operation was unsuccessful. Normally resulting in a program shutdown or crash |

I'd propose removing this log statement entirely, and changing the one below to `DEBUG` level. Currently, this will emit two log lines for each `taskRun` in the worker log and I don't expect these logs to be that useful for typical worker log use-cases. What do you think?

src/deadline_worker_agent/api_models.py
```diff
@@ -94,6 +95,13 @@ class AttachmentDownloadAction(TypedDict):
     stepId: NotRequired[str]
 
 
+class AttachmentUploadAction(TypedDict):
+    sessionActionId: str
+    actionType: AttachmentUploadActionType
+    stepId: str
+    taskId: str
```
jusiskin, line 103: This is not part of the API. Does this belong in the `api_models.py` file?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -91,32 +90,43 @@ def __init__(
...
-                    name="AttachmentDownload",
-                    type=EmbeddedFileTypes_2023_09.TEXT,
-                    runnable=True,
-                    data=script,
-                )
-            ],
-        )
+            "-m",
+        ]
+        args.extend([quote(p) for p in manifests])
```
jusiskin: We shouldn't need to use `quote(p)` here since OpenJD step script commands and arguments are not treated as shell commands.

```suggestion
        args.extend(manifests)
```

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
```diff
@@ -0,0 +1,190 @@
...
+from deadline.job_attachments.models import ManifestSnapshot
+
+from ..session import Session
+from ...log_messages import SessionActionLogKind
+
+
+from .openjd_action import OpenjdAction
+
+if TYPE_CHECKING:
+    from ..session import Session
```
jusiskin, line 29: We already import this up higher

```suggestion
```

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
```diff
@@ -0,0 +1,190 @@
...
+        assert job_attachment_settings.s3_bucket_name is not None
+        assert job_attachment_settings.root_prefix is not None
+
+        s3_settings = JobAttachmentS3Settings(
+            s3BucketName=job_attachment_settings.s3_bucket_name,
+            rootPrefix=job_attachment_settings.root_prefix,
+        )
+
+        manifest_paths_by_root = session._manifest_paths_by_root
+        output_path = os.path.join(session._session.working_directory, "diff")
```
jusiskin: These look like they are both accessing private APIs of `session`. Should we make public APIs for these use-cases?

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
```diff
@@ -0,0 +1,190 @@
...
+            self._logger.info(
+                f"Snapshooting manifest {path} for local root {root}",
+                extra={"openjd_log_content": LogContent.PARAMETER_INFO},
+            )
+            manifest: Optional[ManifestSnapshot] = _manifest_snapshot(
+                root=root,
+                destination=str(output_path),
+                # `output` is used for job download to discover output manifests
+                # manifest file name need to contain the hash of root path for attachment CLI path mapping
+                name=f"output-{os.path.basename(path)}",
```
jusiskin: So it is the responsibility of the caller to know and use this `output-` prefix? This feels like this is something that should be abstracted in the `deadline.job_attachments` API.

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
```diff
@@ -0,0 +1,190 @@
...
+                diff=path,
+            )
+            if manifest:
+                manifests.append(manifest.manifest)
```
jusiskin: So `_manifest_snapshot` returns `None` in some cases? Is there a situation where there are no manifests in `manifests`? I wonder if we can skip the remaining code for such cases.

src/deadline_worker_agent/sessions/actions/scripts/attachment_download.py
```diff
@@ -0,0 +1,53 @@
...
+import time
+import os
+import boto3
+
+from deadline.job_attachments import api
+
+"""
+A small script to download job output. Can provide just the Job ID to download all outputs
+for a Job, optionally include the Step ID to get all outputs for the Job's Step, or optionally
+include the Job, Step, and Task ID to get the outputs for a specific Task.
```
jusiskin: nit: this description looks inaccurate. Maybe a copy/paste leftover?

src/deadline_worker_agent/sessions/actions/scripts/attachment_download.py
```diff
@@ -0,0 +1,53 @@
...
+    parser.add_argument("-s3", "--s3-uri", type=str, help="", required=True)
+    parser.add_argument("-m", "--manifests", nargs="*", type=str, help="", required=True)
+
+    args = parser.parse_args()
+    path_mapping = args.path_mapping
+    manifests = args.manifests
+    s3_uri = args.s3_uri
+
+    print("\nStarting download...")
+    download(s3_root_uri=s3_uri, path_mapping_rules=path_mapping, manifests=manifests)
```
jusiskin, line 48: Does this handle progress reporting ([OpenJD docs](https://github.com/OpenJobDescription/openjd-specifications/wiki/How-Jobs-Are-Run#stdoutstderr-messages)) and graceful cancelation ([OpenJD docs](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#532-cancelationmethodnotifythenterminate))?

Or are we planning to add that in the future?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -91,32 +90,43 @@ def __init__(
...
+
+        with open(Path(__file__).parent / "scripts" / "attachment_download.py", "r") as f:
+            self._step_script = StepScript_2023_09(
+                actions=StepActions_2023_09(
+                    onRun=Action_2023_09(
+                        command=os.path.join(Path(sysconfig.get_path("scripts")), "python"),
```
jusiskin: This differs from how `openjd-sessions` gets the Python interpreter path:

(see https://github.com/OpenJobDescription/openjd-sessions-for-python/blob/84008be79e80cdd9b06095933ea0c58baee89c92/src/openjd/sessions/_subprocess.py#L606)

On Windows with Python 3.12 installed, this path doesn't seem to exist:

image

Same feedback applies to `run_attachment_upload.py`

src/deadline_worker_agent/sessions/actions/scripts/attachment_download.py
```diff
@@ -0,0 +1,53 @@
...
+    -m /sessions/session-f63c206fb5f04c04aa17821001aa3847fajfm5x4/manifests/0bb7eb91fdf8780c4a7e6174de6dfc5e_manifest
+"""
+
+
+def download(s3_root_uri: str, path_mapping_rules: str, manifests: list[str]) -> None:
+
+    api.attachment_download(
+        manifests=manifests,
+        s3_root_uri=s3_root_uri,
+        boto3_session=boto3.session.Session(profile_name=os.environ.get("AWS_PROFILE")),
```
jusiskin: nit: there is no need to pass `profile_name` here. Boto3 uses the `AWS_PROFILE` environment variable by default ([docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables))

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+
+if __name__ == "__main__":
+    start_time = time.perf_counter()
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-pm", "--path-mapping", type=str, help="", required=True)
+    parser.add_argument("-s3", "--s3-uri", type=str, help="", required=True)
+    parser.add_argument("-m", "--manifests", nargs="*", type=str, help="", required=True)
```
jusiskin: Do we need at least one manifest? Should [`nargs`](https://docs.python.org/3/library/argparse.html#nargs) be `+`?

Same applies to `attachment_download.py`?

src/deadline_worker_agent/sessions/actions/scripts/attachment_download.py
```diff
@@ -0,0 +1,53 @@
...
+        path_mapping_rules=path_mapping_rules,
+    )
+
+
+if __name__ == "__main__":
```
jusiskin, line 34: Should we add graceful error handling to this? The default behavior of Python is to output a stack trace and this may not be the most customer-friendly way to present errors.

Same suggestion applies to `attachment_upload.py`.

src/deadline_worker_agent/sessions/session.py
```diff
@@ -242,6 +247,13 @@ def os_user(self) -> Optional[SessionUser]:
         """The session user"""
         return self._os_user
 
+    @property
+    def manifest_paths_by_root(self) -> dict[str, str]:
+        return self._manifest_paths_by_root
+
+    def set_manifest_paths_by_root(self, manifest_paths_by_root) -> None:
+        self._manifest_paths_by_root = manifest_paths_by_root
```
jusiskin: Two things:

1.  nit: we could use a [property setter](https://docs.python.org/3/library/functions.html#property) here
2. we should probably make a copy rather than setting a reference

Putting this together:

```suggestion
    @manifest_paths_by_root.setter
    def manifest_paths_by_root(self, value: dict[str, str]) -> None:
        self._manifest_paths_by_root = dict(value)
```

src/deadline_worker_agent/sessions/session.py
```diff
@@ -299,6 +311,7 @@ def run(self) -> None:
             # service. If an action was running at the time of this exception, its failure is
             # reported immediately in the call to Session._cleanup() below.
             self._stop_fail_message = f"Worker encountered an unexpected error: {e}"
+            self.logger.info(self._stop_fail_message)
```
jusiskin: this should be logged as an error:


```suggestion
            self.logger.error(self._stop_fail_message)
```

src/deadline_worker_agent/sessions/session.py
```diff
@@ -643,6 +658,9 @@ def _start_action(self) -> None:
                 executor=self._executor,
             )
         except Exception as e:
+            if self._output_sync_target_action:
+                action_definition = self._output_sync_target_action.definition
```
jusiskin, line 672: Do we also need to clear `self._output_sync_target_action` here? This exception block cancels remaining actions except for `envExit`s. So handling of those actions may be impacted.

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1084,33 +1102,63 @@ def _action_updated_impl(
...
+
+            if OPENJD_ACTION_STATE_TO_DEADLINE_COMPLETED_STATUS.get(action_status.state, None):
+                task_run_action = self._output_sync_target_action
+                self._output_sync_target_action = None
+                logger.info(
+                    f"task run is {task_run_action}, _output_sync_target_action is {self._output_sync_target_action}"
+                )
+                self._handle_action_update(is_unsuccessful, action_status, task_run_action, now)
+            else:
+                logger.info("AttachmentUploadAction is still running")
```
jusiskin: Should this be logged at `INFO` level? In the future when output upload support progress reporting, this will get logged every time there is a progress update.

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1208,16 +1256,17 @@ def _handle_action_update(
...
+            self._report_action_update(
+                SessionActionStatus(
+                    id=current_action.definition.id,
+                    status=action_status,
+                    start_time=current_action.start_time,
+                    end_time=now if action_status.state != ActionState.RUNNING else None,
+                    update_time=now if action_status.state == ActionState.RUNNING else None,
+                    completed_status=completed_status,
+                )
             )
```
jusiskin, line 1283: I understand this is because we don't want to complete the action

This will mean that progress updates for output upload will not be reported back in `UpdateWorkerSchedule` API requests. I don't believe progress reporting is supported by the output upload yet, but maybe this should be added as a comment for when we build that in.

jusiskin (changes requested):

src/deadline_worker_agent/scheduler/session_queue.py
godobyte: Good catch, already updating it in the next commit.

src/deadline_worker_agent/api_models.py
godobyte: Synced offline, keep it here for all action models to be in the same place. Added a comment to explain.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Good catch, updated.

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
godobyte: I think it's actually the higher one (L28) needs to be removed?

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
godobyte: Synced offline, we'll keep it as-is for now. Ideally we wouldn't need to require certain things in file name with upcoming changes. However if we do, JA library could provide an API to encapsulate this implementation detail to caller.

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
godobyte: `_manifest_snapshot` returns `None` where there are no different from the provided base manifest (via `diff`) after comparison. However we still need to run the remaining code for OpenJD task with the current design, which should be a very light overhead.

src/deadline_worker_agent/sessions/actions/scripts/attachment_download.py
godobyte: Yes updating in a commit.

src/deadline_worker_agent/sessions/actions/scripts/attachment_download.py
godobyte: Not yet, progress reporting and graceful cancelation are planned for future PRs.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: This is related to earlier snapshot is None case (https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/495/files#r1862456177). We still need to support zero manifest at this point.

src/deadline_worker_agent/sessions/session.py
godobyte: Good catch, will incorporate, thanks!

src/deadline_worker_agent/sessions/session.py
godobyte: Good question. I thought it's not necessary assuming new session will be created if needed after the cancellation if any other actions got assigned to Worker afterwards. 

Nevertheless, it's always safe to clear. I'll update this part.

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
leongdl: No - because "output" is our own designation for output manifests. For customers, they can use "snapshot" to capture the state of any directory, and can be input or output. 

I won't abstract it into `_manifest_snapshot` directory. If need be, we can make a helper function like `format_output_manfiest_name`

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -0,0 +1,52 @@
...
+python attachment_upload.py \
+    -pm /sessions/session-f63c206fb5f04c04aa17821001aa3847fajfm5x4/path_mapping.json \
+    -s3 s3://test-job-attachment/DeadlineCloud \
+    -m /sessions/session-f63c206fb5f04c04aa17821001aa3847fajfm5x4/manifests/0bb7eb91fdf8780c4a7e6174de6dfc5e_manifest \
+    -m /sessions/session-f63c206fb5f04c04aa17821001aa3847fajfm5x4/manifests/0bb7eb91fdf8780c4a7e6174de6dfc5e_manifest
+"""
+
+
+def upload(s3_root_uri: str, path_mapping_rules: str, manifests: list[str]) -> None:
+    s3_path = f"{os.environ.get("DEADLINE_FARM_ID")}/{os.environ.get("DEADLINE_QUEUE_ID")}/{os.environ.get("DEADLINE_JOB_ID")}/{os.environ.get("DEADLINE_STEP_ID")}/{os.environ.get("DEADLINE_TASK_ID")}/{os.environ.get("DEADLINE_SESSIONACTION_ID")}"
```
leongdl: Super nit, maybe a nice refactoring later to make this a helper function in Job Attachment APIs.

src/deadline_worker_agent/sessions/session.py
godobyte: I'll change to `DEBUG`.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: Noted!

godobyte:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=aws-deadline_deadline-cloud-worker-agent&pullRequest=495) **Quality Gate failed**
> 
> Failed conditions  [8.1% Duplication on New Code](https://sonarcloud.io/component_measures?id=aws-deadline_deadline-cloud-worker-agent&pullRequest=495&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=aws-deadline_deadline-cloud-worker-agent&pullRequest=495)

This is because some tests added are under feature flag. The duplicate will be removed along with feature flag cleanup.

src/deadline_worker_agent/sessions/actions/scripts/attachment_download.py
godobyte: I'd like to take this in a follow-up PR, there are other error handling cases on my list that I'd like to address together.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -91,33 +87,49 @@ def __init__(
         self._step_details = step_details
         self._logger = LoggerAdapter(OPENJD_LOG, extra={"session_id": session_id})
 
-    def set_step_script(self, manifests, path_mapping, s3_settings) -> None:
-        # TODO - update to run python as embedded file
-        profile = os.environ.get("AWS_PROFILE")
-        deadline_path = os.path.join(Path(sysconfig.get_path("scripts")), "deadline")
+    def set_step_script(self, manifests, s3_settings) -> None:
```
jusiskin: Can we add type hints for the arguments here? It's not documented well, but the guideline is to add type hints anywhere type information cannot be inferred by an IDE, which includes function arguments.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -245,39 +258,31 @@ def start(
...
         # rules that are subsets of each other behave in a predictable manner. We must
         # sort here since we're modifying that internal list appending to the list.
-        session._session._path_mapping_rules.sort(key=lambda rule: -len(rule.source_path.parts))
-
-        # =========================== TO BE DELETED ===========================
-        path_mapping_file_path: str = os.path.join(
-            session._session.working_directory, "path_mapping"
+        session.openjd_session._path_mapping_rules.sort(
+            key=lambda rule: -len(rule.source_path.parts)
         )
```
jusiskin, line 275: Should we encapsulate this in a `Session.add_path_mapping_rules(...)` method? Reasons:

1. we can keep the code here smaller 
2. it seems like code that belongs in the session class
3. we can avoid making `openjd_session` a public attribute of the `Session` class. The less public API the better.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -245,39 +258,31 @@ def start(
...
-        with open(path_mapping_file_path, "w", encoding="utf8") as f:
-            f.write(json.dumps([rule for rule in job_attachment_path_mappings]))
-        # =========================== TO BE DELETED ===========================
 
-        manifest_paths = session._asset_sync._check_and_write_local_manifests(
+        manifest_paths_by_root = session._asset_sync._check_and_write_local_manifests(
             merged_manifests_by_root=merged_manifests_by_root,
-            manifest_write_dir=str(session._session.working_directory),
+            manifest_write_dir=str(session.openjd_session.working_directory),
         )
```
jusiskin: To make this cleaner, we could have a `Session.check_and_write_local_manifests(merged_manifests_by_root)` that is a light wrapper over `session._asset_sync._check_and_write_local_manifests()` and removes the need for exposing `session.openjd_session` as a public attribute.

What do you think?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -245,39 +258,31 @@ def start(
...
+        # TODO: remove type: ignore for manifest_paths_by_root after deadline-cloud release
+        session.manifest_paths_by_root = manifest_paths_by_root  # type: ignore
 
         self.set_step_script(
-            manifests=manifest_paths,
-            path_mapping=path_mapping_file_path,
+            manifests=manifest_paths_by_root.values(),  # type: ignore
```
jusiskin: Looks like this can be removed now:

https://github.com/aws-deadline/deadline-cloud/releases/tag/0.49.3

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -212,15 +225,15 @@ def start(
         # returns root path to PathMappingRule mapping
         dynamic_mapping_rules: dict[str, PathMappingRule] = (
             session._asset_sync.generate_dynamic_path_mapping(
-                session_dir=session._session.working_directory,
+                session_dir=session.openjd_session.working_directory,
```
jusiskin: What if we make a `Session.working_directory` property instead of exposing `Session.openjd_session` entirely?

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
jusiskin: From our discussions, it sounded like output manifests need to be uploaded to a conventionally structured S3 object key and this `output-` prefix is part of that convention.

That convention is assumed by the code to downloading output job attachments. IMO that detail should not be hard-coded in the worker agent, it should be abstracted as an API of the `deadline.job_attachments` library.

Perhaps there could be a function for uploading manifests. The function could accept the necessary inputs (e.g. bucket name, JA prefix, job/step/task IDs, boto client, manifest, etc.). What do you think?

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
jusiskin: >

We discussed offline and thought this was not possible without refactoring the code, but then I recalled that the existing `SyncInputJobAttachmentsAction` handles its own session action life-cycle. See:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/98b1567eef9e3ce57329d3fdc1846f8f1daed73c/src/deadline_worker_agent/sessions/actions/sync_input_job_attachments.py#L114-L122

As discussed this should be low overhead, but I would be interested to measure the overhead for launching the subprocess, waiting for Python imports etc. I suppose this overhead is only encountered when job attachments are enabled but there are no modified files.

What do you think? Should we try to optimize this or leave it as-is for now?

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
```diff
@@ -0,0 +1,191 @@
...
+from openjd.sessions import LOG as OPENJD_LOG, LogContent
+from openjd.model.v2023_09 import (
+    EmbeddedFileTypes as EmbeddedFileTypes_2023_09,
+    EmbeddedFileText as EmbeddedFileText_2023_09,
+    Action as Action_2023_09,
+    StepScript as StepScript_2023_09,
+    StepActions as StepActions_2023_09,
+)
+from openjd.model import ParameterValue
+from deadline.job_attachments.api.manifest import _manifest_snapshot
```
godobyte: This could cause an import error in windows, which need to be fixed before merging.

```
    Python could not import the service's module 
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\__init__.py", line 6, in <module>
        from .installer import install
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\installer\__init__.py", line 15, in <module>
        from deadline_worker_agent.installer.win_installer import (
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\installer\win_installer.py", line 36, in <module>
        from ..windows.win_service import WorkerAgentWindowsService
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\windows\win_service.py", line 14, in <module>
        from deadline_worker_agent.startup.entrypoint import entrypoint
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\startup\entrypoint.py", line 24, in <module>
        from ..worker import Worker
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\worker.py", line 25, in <module>
        from .scheduler import WorkerScheduler
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\scheduler\__init__.py", line 4, in <module>
        from .session_queue import SessionActionQueue
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\scheduler\session_queue.py", line 29, in <module>
        from ..sessions.actions import (
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\sessions\actions\__init__.py", line 10, in <module>
        from .run_attachment_upload import AttachmentUploadAction
      File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\sessions\actions\run_attachment_upload.py", line 24, in <module>
        from deadline.job_attachments.api.manifest import _manifest_snapshot
      File "C:\Program Files\Python311\Lib\site-packages\deadline\job_attachments\api\__init__.py", line 5, in <module>
        from .attachment import attachment_download, attachment_upload
      File "C:\Program Files\Python311\Lib\site-packages\deadline\job_attachments\api\attachment.py", line 18, in <module>
        from deadline.client.cli._groups.click_logger import ClickLogger
      File "C:\Program Files\Python311\Lib\site-packages\deadline\client\cli\__init__.py", line 6, in <module>
        from ._deadline_cli import main
      File "C:\Program Files\Python311\Lib\site-packages\deadline\client\cli\_deadline_cli.py", line 14, in <module>
        from ._common import _PROMPT_WHEN_COMPLETE
      File "C:\Program Files\Python311\Lib\site-packages\deadline\client\cli\_common.py", line 31, in <module>
        sigint_handler = SigIntHandler()
                         ^^^^^^^^^^^^^^^
      File "C:\Program Files\Python311\Lib\site-packages\deadline\client\cli\_groups\_sigint_handler.py", line 19, in __new__
        signal.signal(signal.SIGINT, cls._instance._handle_sigint)
      File "C:\Program Files\Python311\Lib\signal.py", line 58, in signal
        handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ValueError: signal only works in main thread of the main interpreter
     
```

Kudos to @jericht for helping me find the error log in windows instance!

src/deadline_worker_agent/sessions/session.py
```diff
@@ -232,6 +236,11 @@ def openjd_session_action_callback(session_id: str, action_status: ActionStatus)
         # 3. We're already capturing it to send to cloudwatch.
         self.logger = LoggerAdapter(OPENJD_LOG, extra={"session_id": self._id})
 
+    @property
+    def openjd_session(self) -> OPENJDSession:
+        """The openjd session for this session"""
+        return self._session
```
jusiskin, line 242: As discussed elsewhere, I wonder if there's a way we can expose a minimal subset of the OpenJD session:

1.  a `working_directory` property mapping to `self._session.working_directory`
2. a `add_path_mapping_rules()` method
3. others?

test/unit/scheduler/test_session_queue.py
```diff
@@ -223,6 +256,115 @@ def test(
...
+                ),
+                AttachmentDownloadAction(
+                    id="id",
+                    session_id="session-1234",
+                    job_attachment_details=JobAttachmentDetails(
+                        job_attachments_file_system=JobAttachmentsFileSystem.COPIED,
+                        manifests=[],
+                    ),
+                ),
+                id="attachmen download job input",
```
jusiskin: nit:

```suggestion
                id="attachment download job input",
```

test/unit/scheduler/test_session_queue.py
```diff
@@ -223,6 +256,115 @@ def test(
...
+                AttachmentDownloadAction(
+                    id="id",
+                    session_id="session-1234",
+                    step_details=StepDetails(
+                        step_template=_TEST_STEP_TEMPLATE,
+                        dependencies=["step-1"],
+                        step_id="step-1234",
+                    ),
+                ),
+                id="attachmen download step dependency",
```
jusiskin:
```suggestion
                id="attachment download step dependency",
```

test/unit/scheduler/test_session_queue.py
```diff
@@ -223,6 +256,115 @@ def test(
...
+                    id="id",
+                    session_id="session-1234",
+                    step_id="step-1",
+                    task_id="task-1",
+                ),
+                id="attachment upload action",
+            ),
+        ],
+    )
+    def test_attachments_download_actions(
```
jusiskin: nit: this test covers attachment upload actions too. How about `test_attachments_transfer_actions`?

test/unit/scheduler/test_session_queue.py
```diff
@@ -223,6 +256,115 @@ def test(
...
+
+        # THEN
+        assert len(session_queue._actions) == 1
+        assert "id" in session_queue._actions_by_id
+
+        # WHEN
+        next_action = session_queue.dequeue()
+
+        # THEN
+        assert type(next_action) is AttachmentUploadAction
```
jusiskin: For more meaningful test coverage, we should test when the queue has prior actions to ensure the inserted actions dequeues first.

test/unit/sessions/actions/test_run_attachment_upload.py
```diff
@@ -0,0 +1,202 @@
...
+
+        return session
+
+    @pytest.fixture(autouse=True)
+    def mock_manifest_snapshot(self) -> Generator[MagicMock, None, None]:
+        with patch.object(
+            actions_module.run_attachment_upload,
+            "_manifest_snapshot",
+            return_value=ManifestSnapshot(
+                manifest=f"{session_dir}/diff/output-hash_manifest-timestamp.manifest"
```
jusiskin: `session_dir` is not in the fixture's input arguments, so this is pointing to the function defined above.

It doesn't appear to impact the test, but maybe should correct this.

test/unit/sessions/actions/test_run_attachment_upload.py
```diff
@@ -0,0 +1,202 @@
...
+                ],
+            )
+
+        session.run_task.assert_called_once_with(
+            step_script=action._step_script,
+            task_parameter_values=dict[str, ParameterValue](),
+            os_env_vars={
+                "DEADLINE_SESSIONACTION_ID": "sessionaction-abc123",
+                "DEADLINE_STEP_ID": "step-1234",
+                "DEADLINE_TASK_ID": "task-1234",
```
jusiskin: nit: these values could reference the fixtures they originate from rather than replicate the values. That way if we ever parametrize these or change their values, we do not have to change this code.

test/unit/sessions/test_session.py
```diff
@@ -1506,6 +1515,82 @@ def sync_asset_outputs_side_effect(*, current_action: CurrentAction) -> None:
         assert session._current_action is None, "Current session action emptied"
         mock_sync_asset_outputs.assert_called_once_with(current_action=current_action)
 
+    @pytest.mark.skipif(
+        not ASSET_SYNC_JOB_USER_FEATURE,
+        reason="This test will be removed after releasing the asset sync job user feature",
```
jusiskin: nit: I think the reason should be "This test will be run unconditionally after releasing the asset sync job user featuer"

What do you think?

test/unit/sessions/test_session.py
```diff
@@ -1506,6 +1515,82 @@ def sync_asset_outputs_side_effect(*, current_action: CurrentAction) -> None:
...
+                actionType="SYNC_OUTPUT_JOB_ATTACHMENTS",
+                stepId=step_id,
+                taskId=task_id,
+            )
+        )
+
+    @pytest.mark.skipif(
+        ASSET_SYNC_JOB_USER_FEATURE,
+        reason="This test will be removed after releasing the asset sync job user feature",
+    )
```
jusiskin: Should we have equivalent coverage for when `ASSET_SYNC_JOB_USER_FEATURE` is set?

test/unit/sessions/test_session.py
```diff
@@ -1588,6 +1673,10 @@ def mock_now(*arg, **kwarg) -> datetime:
         assert session._current_action is None, "Current session action emptied"
         mock_sync_asset_outputs.assert_called_once_with(current_action=current_action)
 
+    @pytest.mark.skipif(
+        ASSET_SYNC_JOB_USER_FEATURE,
+        reason="This test will be removed after releasing the asset sync job user feature",
+    )
```
jusiskin: We're planning to remove this behaviour?

jusiskin (changes requested):

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Yes! I'm on it.

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
godobyte: Actually there is an issue with Windows for running snapshot here (https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/495/files#r1868679307). I'm moving snapshot to `attachment_upload.py`, this would be no longer applicable.

test/unit/sessions/actions/test_run_attachment_upload.py
godobyte: I'm removing this part in the next commit.

test/unit/sessions/test_session.py
godobyte: The behaviour would be different. This part needs a bit more work before adding a test. I'd take this as a follow-up.

test/unit/sessions/test_session.py
godobyte: I can take this as a follow-up.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Sure I can create `working_directory` property.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: We could, but I think that's out of scope for this PR. It's already large and we shouldn't pile on more changes.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: I'm not sure about the value it brings to only refactor this, `session._asset_sync._launch_vfs` should also be refactored in that case. Perhaps we should take a step back and look at the overall context surrounding exposing attributes from `session`.

src/deadline_worker_agent/sessions/session.py
godobyte: I can do 1 as part of this PR, 2 and 3 is likely out of scope considering additional refactoring may be needed for path mapping rules.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: When you have code from one class modifying/accessing the private members of another class (`session._asset_sync`), it breaks encapsulation. Doing this once or twice can be acceptable in cases (e.g. we do not control the code), but the more it happens the harder it becomes to reason about a unit of code (e.g. a class) when you cannot be certain if/when external code will modify or relies on its internal state. This is why it is always better to follow the practice of formalizing the public API and respecting encapsulation whenever possible.

>  `session._asset_sync._launch_vfs` should also be refactored in that case

I would not object to that either 😄

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: The intent was to to avoid exposing the entire `.openjd_session` property and expose only the subset of its API that we need to publicly.

Leaving it as a public attribute makes it more likely that external code does things that could conflict/interfere with the logic of the worker agent's outer `Session` class.

test/unit/sessions/test_session.py
jusiskin: Okay, but please change the reason to something like:

>   This test will be re-written before releasing the asset sync job user feature

test/unit/sessions/test_session.py
jusiskin: Same as above, let's adjust the reason to indicate it needs to be re-written and not removed.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -0,0 +1,77 @@
...
+            destination=str(output_path),
+            # `output` is used for job download to discover output manifests
+            # manifest file name need to contain the hash of root path for attachment CLI path mapping
+            name=f"output-{os.path.basename(path)}",
+            diff=path,
+        )
+        if manifest:
+            manifests.append(manifest.manifest)
+
+    return manifests
```
jusiskin: I'm having a bit of trouble understanding how this function works.

1.  What is the role of the `output_path`? It is fed into the `destination` argument of `_manifest_snapshot`. The basename of the path is `diff`
2. There is a `diff` argument which appears to be supplied by the values of `manifest_paths_by_root`. I suspect this a path pointing to a manifest snapshot of the previous contents of the asset root, right?
3.  At the end, we accumulate a list of other paths obtained from the return value of `_manifest_snapshot`'s `manifest` attribute. So we are not using `destination` in the end, but this maybe is being used as a file to preserve state between each loop iteration.

I wonder if code comments could help to explain what is going on here.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: Discussed offline and we will improve this in a future PR

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: Discussed offline and we will save this for a future PR.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
jusiskin: We will remove `openjd_session` as a public attribute in a future PR.

jusiskin (changes requested):

test/unit/sessions/test_session.py
godobyte: Updated!

test/unit/sessions/test_session.py
godobyte: Updated!

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: Added comments, please let you know if anything is unclear!

jusiskin: Looks much better, thanks!

jusiskin (dismissed):

src/deadline_worker_agent/scheduler/session_queue.py
```diff
@@ -287,6 +293,32 @@ def cancel_all(
                 )
             )
 
+    def insert_front(
+        self,
+        *,
+        action: AttachmentUploadActionApiModel,
+    ) -> None:
+        """Inserts an attachmen upload action at the front of the queue
```
leongdl: nit nit typo `attachmen`t

leongdl:
```suggestion
        """Inserts an attachment upload action at the front of the queue
```

jusiskin (approved):

leongdl (approved):

PR #499, opened 2024-12-02
test: make credentials tests more robust to wait for job to start

YutongLi291:
The credentials tests were not waiting properly for the job to start, causing them to be flaky sometimes.

This was because we are only checking for the lifecycle status, not the task run status.
Check that the taskRunStatus of the job is `RUNNING` or `SUCCEEDED` to makes sure that the job has started.
The tests will be less flaky
`
source .e2e_linux_infra.sh
hatch run e2e-test
`

No windows test changes
N0
No
----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -95,7 +95,19 @@ class TestJobSubmission:
             def is_job_started(current_job: Job) -> bool:
                 current_job.refresh_job_info(client=deadline_client)
                 LOG.info(f"Waiting for job {current_job.id} to be created")
-                return current_job.lifecycle_status != "CREATE_IN_PROGRESS"
+
+                assert current_job.task_run_status not in [
+                    TaskStatus.INTERRUPTING,
+                    TaskStatus.SUSPENDED,
+                    TaskStatus.CANCELED,
+                    TaskStatus.FAILED,
+                    TaskStatus.SUCCEEDED,
+                    TaskStatus.NOT_COMPATIBLE,
+                ], f"Job is not in a valid task run status for this test: {current_job.task_run_status}"
+                return (
+                    current_job.lifecycle_status != "CREATE_IN_PROGRESS"
+                    and current_job.task_run_status == TaskStatus.RUNNING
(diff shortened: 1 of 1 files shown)


test/e2e/test_job_submissions.py
```diff
@@ -95,7 +95,10 @@ def test_queue_credentials_file_is_secure_from_other_users(
             def is_job_started(current_job: Job) -> bool:
                 current_job.refresh_job_info(client=deadline_client)
                 LOG.info(f"Waiting for job {current_job.id} to be created")
-                return current_job.lifecycle_status != "CREATE_IN_PROGRESS"
+                return (
+                    current_job.lifecycle_status != "CREATE_IN_PROGRESS"
+                    and current_job.task_run_status in [TaskStatus.RUNNING, TaskStatus.SUCCEEDED]
```
jusiskin: I think in this test, we want to run SSM commands while the session is running, right? With that in mind, should the test raise an exception if the `task_run_status` is in any status that we don't want? Looking at the [API docs](https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_GetJob.html#deadlinecloud-GetJob-response-taskRunStatus), I'm thinking we want to fail if the job is in:

* `PENDING`
* `INTERRUPTING`
* `SUSPENDED`
* `CANCELED`
* `FAILED`
* `SUCCEEDED`
* `NOT_COMPATIBLE`

Same applies to the other `is_job_started` function below.

jusiskin (changes requested):

test/e2e/test_job_submissions.py
YutongLi291: If the session is in any of the other states, we would fail at this `is_job_Started` step because it would never be in the state that we want. But if we fail if the state is in `PENDING` for example, then that means that we aren't even waiting for the job to be started?

YutongLi291: Okay, makes sense, I'll publish a revision

jusiskin (approved):

jericht (approved):

PR #512, opened 2024-12-17
fix: more robust config file modifications in install-deadline-worker

### LINKED ISSUE #308: Running install-deadline-worker with a previous farm and fleet configured fails
## Problem

The `install-deadline-worker` command fails when specifying a `--fleet-id` and/or `--farm-id` that differ from the values in a pre-existing agent configuration file (e.g. `/etc/amazon/deadline/worker.toml`)

### Reproduction steps

1.  Run the agent installer for one fleet:

    ```
    install-deadline-worker --farm-id farm-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa --fleet-id fleet-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    ```
2.  Re-run the agent installer for another fleet or farm:

    ```
    install-deadline-worker --farm-id farm-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa --fleet-id fleet-bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
    ```

## Expected behavior

The agent modifies the farm and fleet IDs in the configuration file

## Actual behavior

The agent errors with:

```
farm_id = "farm-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
ERROR: Failed to configure fleet ID in /etc/amazon/deadline/worker.toml.
```

jusiskin:
#### Background

The worker agent provides a `install-deadline-worker` CLI command that is used for setting up a worker host. The command accepts configuration settings in its command-line arguments for configuration settings to be applied. Those settings are applied by modifying the worker agent configuration file ([docs](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/docs/configuration.md#3--config-file)), which uses the [TOML](https://toml.io/en/) file format.

#### Problem

The code for modifying the worker agent configuration file used simple pattern matching and replacement. For Linux, the `sed` command was used. On Windows, Python's [`re.sub(...)`](https://docs.python.org/3/library/re.html#re.sub) was used.

This implementation is fragile and comes with multiple issues:

1.  Any trailing inline TOML comments, for example:

    ```toml
    [worker]
    farm_id = "farm-123"  # our staging farm
    ```
    could cause the pattern matching to fail
2. If the pattern matching allows for trailing comments, the replacement may not preserve them after modification.
3. The patterns may be overly-strict, for example:

    https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/c26590d5e193dc01c33bba831b638e4c9fee81bf/src/deadline_worker_agent/installer/install.sh#L411-L412

   Which is the root cause of #308
4.  If the config file exists but some settings are not defined or commented out, the configuration cannot be applied. One example is when a new setting is introduced to the config file (#513). In such cases, the current implementation of `install-deadline-worker` will not be able to find an existing setting and fail.

Added `tomlkit` as a dependency. This is a Python package that allows for modifying TOML documents while preserving style (comments, blank space, ordering, indentation, etc...).

Functionality was added to `deadline_worker_agent.config.config_file.ConfigFile` to modify settings for a tomlkit document (or alternatively specifying a path to a config file and `tomlkit` is used to modify it).

The logic for modifying settings has the following algorithm:

```mermaid
flowchart LR

A((Start)) -->|Set| B{Exists?}
A -->|Unset| C{Exists?}
B -->|Exists| D[Modify]
D --> H((Done))
B -->|Commented out| E[Uncomment and set value]
E --> H
B -->|Does not exist| F[Create with doc comments]
F --> H
C -->|Exists| G[Comment out]
G --> H
C -->|Does not exist| H
```

This code was integrated into the `install-deadline-worker` logic. Of particular note, the Linux implementation of `install-deadline-worker` runs a bash script (`install.sh`) in a subprocess which handles the installation logic. To integrate with the new Python logic for TOML config file modification, a command-line interface was added to `src/deadline_worker_agent/config/__main__.py`.

The path to the Python interpreter is passed as an argument to `install.sh` which is used to invoke this Python command-line interface:

```sh
${python_interpreter_path} -m deadline_worker_agent.config ...
```

1.  The `install-deadline-worker` command uses more robust code for modifying the worker agent configuration file and preserving style and comments of the TOML document.
2.  Can use the `install-deadline-worker` to change the farm and fleet ID of an existing worker agent config file (fixes #308)

1.  Integration tests were added for the newly added Python functions and the `deadline_worker_agent.config` command-line interface. I've expanded the GitHub workflows to also run integration tests on Ubuntu (they were previously just run on Windows)
2. I ran the worker agent end-to-end tests which succeeded.

No, this is mostly fixing our implementation and has more robust behavior for edge cases.

No

----


jusiskin:
There are "expected checks" ([docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging)) that are still waiting here.

This is because this PR changes the name of the checks (see `.github/workflows/code_quality.yml`). Once I have review approvals and we are ready to merge this, we will modify the branch protection rules for `mainline` to reflect the new check names.

.github/workflows/code_quality.yml
```diff
@@ -2,21 +2,41 @@ name: Code Quality
...
     inputs:
       branch:
         required: false
         type: string
 
 jobs:
-  Test:
-    name: Python
+  UnitTest:
+    name: Unit Tests - ${{ matrix.os }}, python ${{ matrix.python-version }}
```
moorec-aws: I think we will run into some long repetitive naming here? The os and python version are already listed. Maybe change to `Python Unit Tests`. See: https://github.com/aws-deadline/deadline-cloud-worker-agent/actions/runs/12462070978

.github/workflows/code_quality.yml
```diff
@@ -2,21 +2,41 @@ name: Code Quality
 
 on:
   pull_request:
-    branches: [ mainline, feature_windows ]
```
moorec-aws, line 5: Lets change this to `[ mainline, feature_* ]` so that the code quality tests run on any feature branches that are created.

.github/workflows/release_bump.yml
```diff
@@ -75,7 +53,7 @@ jobs:
       group: windowse2e
 
   Bump:
-    needs: [LinuxE2ETests, WindowsE2ETests, WindowsIntegrationTests]
+    needs: [LinuxE2ETests, WindowsE2ETests, IntegrationTests]
```
moorec-aws: We don't need `IntegrationTests` because they are run in the code quality check now?

jusiskin: Good call. And maybe we change the step name from `UnitTests` &rarr; `Tests` while we're here

.github/workflows/code_quality.yml
jusiskin: I went with just `Unit Tests` since I figured "Python" was unnecessary. Updated now, let me know what you think.

.github/workflows/code_quality.yml
jusiskin: Done

.github/workflows/release_bump.yml
jusiskin: This is updated. Let me know what you think now.

src/deadline_worker_agent/config/__main__.py
```diff
@@ -0,0 +1,227 @@
...
+)
+
+
+logger = getLogger(__name__)
+
+
+class ParsedArguments(argparse.Namespace):
+    backup: bool
+    """Whether to backup the existing worker agent configuration file. The backup is created
+    with along-side the worker agent configuration file with a .bak extension added"""
```
AWS-Samuel:
```suggestion
    """Whether to backup the existing worker agent configuration file. The backup is created
    along-side the worker agent configuration file with a .bak extension added"""
```

AWS-Samuel: Should this default to True?

src/deadline_worker_agent/config/__main__.py
```diff
@@ -0,0 +1,227 @@
...
+        help=(
+            "The worker agent will be allowed to run with an EC2 instance profile associated "
+            "to the host instance."
+        ),
+        action="store_const",
+        const=True,
+        required=False,
+    )
+    allow_instance_profile_group.add_argument(
+        "--no-allow-ec2-instance-profile",
```
AWS-Samuel, line 80: Should this be `--disallow-instance-profile`? This was the previous install arg https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/ca76875b3d462d2334bab5ee1f16613ef9305c83/src/deadline_worker_agent/installer/install.sh#L44

The only usage I found for `no-allow` appears to be a logging bug: https://github.com/search?q=repo%3Aaws-deadline%2Fdeadline-cloud-worker-agent+--no-allow&type=code

src/deadline_worker_agent/config/__main__.py
```diff
@@ -0,0 +1,227 @@
...
+    parser.add_argument(
+        "--no-backup",
+        help=(
+            "Do not create a backup. The default behavior is to create a backup of the "
+            'configuration file to a path an added ".bak" extension.'
+        ),
+        action="store_false",
+        dest="backup",
+    )
+    return parser
```
AWS-Samuel, line 150: not all arguments in the current installer are supported here: https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/ca76875b3d462d2334bab5ee1f16613ef9305c83/src/deadline_worker_agent/installer/install.sh#L44

Should we document this?

src/deadline_worker_agent/installer/install.sh
```diff
@@ -47,13 +47,20 @@ start_service="no"
 telemetry_opt_out="no"
 warning_lines=()
 vfs_install_path="unset"
+python_interpreter_path=unset
```
AWS-Samuel: nit 
```suggestion
python_interpreter_path="unset"
```

src/deadline_worker_agent/installer/install.sh
```diff
@@ -47,13 +47,20 @@ start_service="no"
...
 usage()
 {
-    echo "Usage: install.sh --farm-id FARM_ID --fleet-id FLEET_ID"
-    echo "                  [--region REGION] [--user USER]"
-    echo "                  [--scripts-path SCRIPTS_PATH]"
+    echo "Usage: install.sh --farm-id FARM_ID"
+    echo "                  --fleet-id FLEET_ID"
+    echo "                  --scripts-path SCRIPTS_PATH"
+    echo "                  --python-interpreter-path PYTHON_INTERPRETER_PATH"
+    echo "                  --region REGION"
```
AWS-Samuel, line 58: Are these all required arguments now? 

I see logic below that seems to indicate as such. We should mark this change as breaking.

src/deadline_worker_agent/installer/install.sh
```diff
@@ -173,18 +184,8 @@ elif ! validate_deadline_id fleet "${fleet_id}"; then
...
-        echo "ERROR: Could not find deadline-worker-agent in search path"
-        exit 1
-    fi
-    client_library_program=$(which deadline)
-    if [[ "$?" != "0" ]]; then
-        echo "ERROR: Could not find deadline in search path"
-        exit 1
-    fi
-    set -e
+    echo "ERROR: --scripts-path is not specified"
```
AWS-Samuel, line 187: Why are we make this required and removing the previous resolving logic?

src/deadline_worker_agent/installer/install.sh
```diff
@@ -203,6 +204,13 @@ else
     fi
     set -e
 fi
+if [[ "${python_interpreter_path}" == "unset" ]]; then
+    echo "ERROR: --python-interpreter-path is not specified"
+    usage
+elif [[ ! -f "${python_interpreter_path}" ]]; then
+    echo "ERROR: The Python interpreter path is not found: \"${python_interpreter_path}\""
+    usage
+fi
```
AWS-Samuel, line 213: Can we default this to whichever python shows up first in the search path to make it non-required?

src/deadline_worker_agent/installer/win_installer.py
```diff
@@ -318,123 +323,63 @@ def update_config_file(
     replaces specific placeholders with the provided values.
 
     Parameters:
-    - deadline_config_sub_directory (str): Subdirectory for Deadline configuration files.
     - farm_id (str): The farm ID to set in the configuration.
     - fleet_id (str): The fleet ID to set in the configuration.
     - shutdown_on_stop (Optional[bool]): The shutdown_on_stop value to set. Does nothing if set to None.
+    - allow_ec2_instance_profile (Optional[bool]): Whether the agent should be configured to run with[out] an EC2 instance profile.
+        Does nothing if set to None.
```
AWS-Samuel: Nit, for boolean options I think it makes sense to default to `True/False` instead of `None` to make the result clearer when it isn't specified

AWS-Samuel (changes requested):

src/deadline_worker_agent/config/__main__.py
jusiskin: The attribute's value is always written to by `argparse.ArgumentParser.parse_args()` and has an effective default of `True`. The code adds a `--no-backup` argument to the parser with an `action` of `store_false`. See

https://github.com/jusiskin/deadline-cloud-worker-agent/blob/cda9398085fffb15641a757e60dde20070551be1/src/deadline_worker_agent/config/__main__.py#L141-L150

The behaviour or `store_false` ([docs](https://docs.python.org/3/library/argparse.html#action)) is:

>   `'store_true'` and `'store_false'` - These are special cases of `'store_const'` used for storing the values `True` and `False` respectively. In addition, they create default values of `False` and `True` respectively:
>   ```py
>   >>> parser = argparse.ArgumentParser()
>   >>> parser.add_argument('--foo', action='store_true')
>   >>> parser.add_argument('--bar', action='store_false')
>   >>> parser.add_argument('--baz', action='store_false')
>   >>> parser.parse_args('--foo --bar'.split())
>   Namespace(foo=True, bar=False, baz=True)
>   ```

src/deadline_worker_agent/config/__main__.py
jusiskin: I had used the `--no-*` convention in this new command-line interface to set the value to `False`. For example, it also includes `--no-shutdown-on-stop`, `--no-windows-job-user`, and `--no-backup`.

You're right though&hellip; this is now inconsistent with the command-line arguments for the `install-deadline-worker` entrypoint:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/7dca49ce16459c7f05688458b633468f875dbb4c/src/deadline_worker_agent/installer/__init__.py#L250

and the `deadline-worker` entrypoint:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/7dca49ce16459c7f05688458b633468f875dbb4c/src/deadline_worker_agent/config/cli_args.py#L110

On the one hand, I'm incline to go with `--no-*` since it would become easier to generalize this code for all configuration settings. On the other hand, customers will value the argument naming consistency across different command-line entrypoints.

---

Another consideration was that this new command-line interface is was intended to be a private interface and only temporary. I intentionally didn't document it as it was meant as a way for the Linux bash installation code to call it. We hope to one day migrate that bash code to Python. After that, we can remove this command-line interface.

On this basis, how do you feel about leaving this as-is?

src/deadline_worker_agent/config/__main__.py
jusiskin: It is documented in the command-line help below:

https://github.com/jusiskin/deadline-cloud-worker-agent/blob/cda9398085fffb15641a757e60dde20070551be1/src/deadline_worker_agent/installer/install.sh#L104-L108

src/deadline_worker_agent/installer/install.sh
jusiskin: >   Are these all required arguments now?

Yes. Both `--scripts-path` and `region` where required before but were not documented here. I just added them now. For `--python-interpreter-path`, yes this is technically a breaking change, but this script is not meant to be invoked directly by end-users. Instead, the `install-deadline-worker` entrypoint is the public interface. It then invokes this bash script and always provides these required arguments.

src/deadline_worker_agent/installer/install.sh
jusiskin: As mentioned above, this script is intended to be a private interface and the public entrypoint is `install-deadline-worker`. Before this PR, the `install-deadline-worker` entrypoint had always passed the `--scripts-path` argument and that behaviour has not changed.

This change prevents using this script in an unsupported way: directly invoking it w/o the `--scripts-path` argument, but I would not consider that to be a breaking change.

src/deadline_worker_agent/installer/install.sh
jusiskin: We want to use the same Python interpreter in which the worker agent program files are installed - especially when the worker agent is installed in a Python virtual environment. Otherwise the worker agent's dependencies may not be available or may have conflicting versions.

Again, because this is always provided by the `install-deadline-worker` entrypoint ([source](https://github.com/jusiskin/deadline-cloud-worker-agent/blob/cda9398085fffb15641a757e60dde20070551be1/src/deadline_worker_agent/installer/__init__.py#L125-L126)), this only provides a guardrail from calling this bash script from a different source which is not something the worker agent officially supports.

src/deadline_worker_agent/installer/win_installer.py
jusiskin: My reasoning for having this optional / allowing `None`, was to have a way for callers to not modify any prior existing config setting value.

But looking at this, the calling code always provides a `True` / `False` value, so this is unused. I can remove this from the API since we don't need it.

src/deadline_worker_agent/config/__main__.py
AWS-Samuel: Since this entry point is intended to be a private interface, I think you're right, that it's not as important to have consistency with the public entry point arguments. As such I am okay with leaving this as is

AWS-Samuel (dismissed):

moorec-aws (dismissed):

moorec-aws (approved):

AWS-Samuel (approved):

PR #513, opened 2024-12-17
feat: configurable session root directory

jusiskin:

Currently the Worker Agent has a hard-coded root directory under which session directories are created.

| Operating System | Session Root Directory |
| --- | --- |
| Linux | `/sessions` |
| Windows | `%PROGRAMDATA%\Amazon\OpenJD` |

Those using the worker agent may want to customize the directory where the sessions are created.

Added a new session root directory setting to the worker agent configuration. As with other worker agent settings, this can be set using multiple sources in the following order of precedence (first is higher priority):

1.  Command-line argument: `--session-root-dir`
2. Environment variable: `DEADLINE_WORKER_SESSION_ROOT_DIR`
3. Configuration file: `[worker]` &rarr; `session_root_dir`
4. Defaults:
    *   `/sessions` on Linux
    *   `%PROGRAMDATA%\Amazon\OpenJD` on Windows

The setting is supplied in the `install-deadline-worker` command using the `--session-root-dir` CLI argument which will create (if needed) the worker agent configuration file and modify the `session_root_dir` setting. The installer will also provision the directory as needed and ensure it has the correct file-system ACLs.

The setting is now loaded on launch and passed down through the call stack and supplied to `session_root_directory` kwarg of the `Session` class in `openjd-sessions` ([code ref](https://github.com/OpenJobDescription/openjd-sessions-for-python/blob/b9be7a477921e384337cff009444cdf80679eb46/src/openjd/sessions/_session.py#L320)).

Those using the worker agent can configure where session directories will be created on the worker host.

*  Integration test cases were added for the worker agent configuration file modification for the session root directory
*  End-to-end test was added that verifies the configured session root directory is applied. This test succeeded

The new `session_root_dir` was added to the `worker.toml.example` file with documentation comments

No

----

### DIFF
--- a/src/deadline_worker_agent/config/__main__.py
+++ b/src/deadline_worker_agent/config/__main__.py
@@ -41,6 +41,9 @@ class ParsedArguments(argparse.Namespace):
     """A Windows username to override the queue jobRunAs configuration. If False, then no
     modification is made. If None, then the setting is unset."""
 
+    session_root_dir: str | None = None
+    """Path to the parent directory where the worker agent creates per-session subdirectories under"""
+
 
 def create_argument_parser() -> argparse.ArgumentParser:
     """Creates the argparse ArgumentParser for the deadline_worker_agent.config module"""
@@ -61,6 +64,12 @@ def create_argument_parser() -> argparse.ArgumentParser:
         "--fleet-id",
         help="The unique identifier for the Deadline Cloud fleet",
         required=False,
+    )
+    worker_group.add_argument(
+        "--session-root-dir",
+        help="The parent directory that worker agent creates per-session subdirectories under",
+        required=False,
+        type=str,
     )
 
     aws_group = parser.add_argument_group("AWS", "Settings related to AWS and EC2 hosts")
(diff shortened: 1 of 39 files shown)


src/deadline_worker_agent/config/__main__.py
```diff
@@ -1,4 +1,5 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# type: ignore
```
AWS-Samuel:
```suggestion

```

src/deadline_worker_agent/windows/win_os_session.py
```diff
@@ -0,0 +1,30 @@
...
+@cache
+def is_windows_session_zero() -> bool:
+    """Returns whether the current Python process is running in Windows session 0.
+
+    Returns
+    -------
+    bool
+        True if the current process is running in Windows session 0
+    """
+    return _get_current_process_session() == 0
```
AWS-Samuel: I think we can delete this file as it already exists on mainline under `src/deadline_worker_agent/windows/win_session.py`

src/deadline_worker_agent/config/__main__.py
jusiskin: Fixed

src/deadline_worker_agent/windows/win_os_session.py
jusiskin: Ah good catch. This must have been a leftover from some rebasing. Fixed now

moorec-aws (approved):

AWS-Samuel (approved):

PR #528, opened 2025-01-20
chore(deps): update mypy from ~= 1.13 to ~= 1.14, types-psutil from ~= 5.9 to ~= 6.1 and change version expression of deadline-cloud-test-fixtures

evanspearman-a:
Dependabot is refusing to rebase changes, so we're going to manually update them:

1. Changed the way deadline-cloud-test-fixtures is pinned from `>= 0.17.1, == 0.17.*` to `>= 0.17.1, < 0.18` because `0.17.2` existing was causing dependabot to want to "upgrade" to `0.17.*`.
2. Updated mypy to 1.14. This caused an issue because we were previously converting an `Optional[int]` into `float` which mypy appears to be more strict about now. We now check to see if `psutil.cpu_count()` returns `None` and raise an exception if it is.
3. Updated types-psutil to 6.1

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -2,7 +2,7 @@ backoff == 2.2.*
 coverage[toml] ~= 7.6
 coverage-conditional-plugin == 0.9.*
 # The fixture code relies on a bugfix in 0.17.1
-deadline-cloud-test-fixtures >= 0.17.1, == 0.17.*
+deadline-cloud-test-fixtures >= 0.17.1, < 0.18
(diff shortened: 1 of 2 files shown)


evanspearman-a:
Note: The SonarCloud finding is incorrect. It's saying that `cpu_count is None` will always be `False`, but this isn't true since `psutil.cpu_count()` can return None.

jusiskin (approved):

edwards-aws (approved):

Issue #536, opened 2025-02-01
Bug: Worker agent aborts startup when detecting GPU memory on a disabled GPU


cilevitz:
### Describe the bug

On a system with a disabled NVIDIA GPU that has no memory allocated to the GPU, the worker agent exits in its initialization phase due to a critical error from `_get_gpu_memory()`.

### Expected Behaviour

The worker agent should complete initialization and start up even if it can't detect memory allocated to a GPU.

### Current Behaviour

The `_get_gpu_memory()` function expects `nvidia-smi` to return integer output followed by a string (e.g. `2048 MiB`), which can be parsed with int(). 

When a GPU has no memory allocated, `nvidia-smi` returns `[N/A]` for its memory size so `_get_gpu_memory()` fails and throws a `ValueError` which propagates upward through `detect_system_capabilities()` to  `entrypoint()` which finally aborts with a critical error.

### Reproduction Steps

Use a machine with one NVIDIA GPU in it and a non-NVIDIA GPU (i.e. Intel Graphics), with a monitor connected to the non-NVIDIA GPU.

Disable the NVIDIA GPU:
1. Open the Windows Settings app
2. Navigate to "Graphics settings"
3. Disable "Hardware-accelerated GPU scheduling"

With the NVIDIA GPU disabled, try to start the worker agent with `sc start DeadlineWorker` or `deadline-worker-agent`.

If running as a service, there will be no log output written.

If running as `deadline-worker-agent`, the following will appear at the console before the worker agent exits:
```
[2025-01-31 12:49:48,932][INFO    ] Number of GPUs: 1
[2025-01-31 12:49:48,962][CRITICAL] invalid literal for int() with base 10: '[N/A]'
[2025-01-31 12:49:48,983][INFO    ] 🚪 Worker Agent exiting
```

### Environment

OS: Windows 10 Pro version 22H2 build 19045.5371

NVIDIA GPU: GeForce RTX 3050
Non-NVIDIA GPU: Intel UHD Graphics

Python 3.12

Package version: deadline_cloud_worker_agent-0.27.5 from pypi

cilevitz:
A fix is ready for review in PR https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/537

crowecawcaw:
Thank you for taking the time to captures all the details in this bug report, especially the output of your `nvidia-smi` command. We were not able to recreate quite the same setup, but the report and the PR that you started had enough info to get this across the line.

PR #537, opened 2025-02-01
fix: Assume 0 MiB of memory is available on a disabled NVIDIA GPU

cilevitz:
…

As described in https://github.com/aws-deadline/deadline-cloud-worker-agent/issues/536:

On a system with a disabled NVIDIA GPU that has no memory allocated to the GPU, the worker agent exits in its initialization phase due to a critical error from `_get_gpu_memory()` trying to parse a string as an integer.

`nvidia-smi --query-gpu=memory.total` returns a memory value that can be parsed by `int()` if the GPU is active, or the string "`[N/A]`" if the GPU is disabled. There are no other mode flags (e.g. `display_active`, `display_mode`) that can be used to determine whether a GPU has no memory allocated to it.

Catching the `ValueError` raised when trying to parse the "`[N/A]`" string output from `nvidia-smi --query-gpu=memory.total` instead of letting it propagate upward allows the worker agent to finish initializing. A GPU that's present in a system but not active contributes zero memory towards that system's total GPU memory, and therefore `_get_gpu_memory()` is made to return 0 for a GPU if that GPU's memory size cannot be queried.

GPUs without allocated memory will be part of the total number of GPUs available on the machine, but they contribute 0 MiB to the total amount of GPU memory available on the machine. The worker agent emits a warning in its log that a non-numeric result was received from `nvidia-smi`.

The change was tested on an affected machine and the worker agent starts up successfully.

The docstring is updated and the code block has comments.

No.

----

### DIFF
--- a/src/deadline_worker_agent/capabilities.py
+++ b/src/deadline_worker_agent/capabilities.py
@@ -106,6 +106,7 @@ def _get_gpu_count(*, verbose: bool = True) -> int:
 def _get_gpu_memory(*, verbose: bool = True) -> int:
     """
     Get the total GPU memory available on the machine.
+    Disabled GPUs will appear to have 0 MiB memory if they have no memory currently allocated for use.
 
     Returns
     -------
@@ -138,7 +139,15 @@ def _get_gpu_memory(*, verbose: bool = True) -> int:
 
     mem_per_gpu: list[int] = []
     for line in output.splitlines():
-        mem_mib = int(line.replace("MiB", ""))
+        try:
+            # Active GPU: nvidia-smi returns an integer followed by "MiB"
+            # Disabled GPU with no memory currently allocated: nvidia-smi returns "[N/A]"
+            mem_mib = int(line.replace("MiB", ""))
+        except ValueError:
+            if verbose:
+                _logger.warning(f"Could not detect GPU memory, non-numeric result returned by nvidia-smi: {line}")
+            mem_mib = 0
+        
         mem_per_gpu.append(mem_mib)
 
     min_memory = min(mem_per_gpu)
--- a/src/deadline_worker_agent/capabilities.py
+++ b/src/deadline_worker_agent/capabilities.py
@@ -164,7 +164,7 @@ def _get_gpu_memory(*, verbose: bool = True) -> int:
         return 0
     output = output_bytes.decode().strip()
 
-    mem_per_gpu = _parse_gpu_memory(output)
+    mem_per_gpu = _parse_gpu_memory(output, verbose)
 
     min_memory = min(mem_per_gpu)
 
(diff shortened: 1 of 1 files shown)


src/deadline_worker_agent/capabilities.py
```diff
@@ -136,10 +164,7 @@ def _get_gpu_memory(*, verbose: bool = True) -> int:
         return 0
     output = output_bytes.decode().strip()
 
-    mem_per_gpu: list[int] = []
-    for line in output.splitlines():
-        mem_mib = int(line.replace("MiB", ""))
-        mem_per_gpu.append(mem_mib)
+    mem_per_gpu = _parse_gpu_memory(output)
```
AWS-Samuel:
```suggestion
    mem_per_gpu = _parse_gpu_memory(output, verbose)
```

I think we will want to pass on the `verbose` option when we call this function

src/deadline_worker_agent/capabilities.py
```diff
@@ -103,9 +103,37 @@ def _get_gpu_count(*, verbose: bool = True) -> int:
...
+    mem_per_gpu: list[int] = []
+    for line in lines.splitlines():
+        try:
+            # Active GPU: nvidia-smi returns an integer followed by "MiB"
+            # Disabled GPU with no memory currently allocated: nvidia-smi returns "[N/A]"
+            mem_mib = int(line.replace("MiB", ""))
+        except ValueError:
+            if verbose:
+                _logger.warning(f"Could not detect GPU memory, non-numeric result returned by nvidia-smi: {line}")
+            mem_mib = 0
```
AWS-Samuel, line 127:
```suggestion
            continue
```

How do you feel about simply skipping to the next line with `continue` if `nvidia-smi` returns a non integer response? This would allow us to keep the current pattern of not adding anything to the `mem_per_gpu` list when no GPU is detected.

AWS-Samuel (commented):
Thanks for discovering this edge case and putting up a fix! I have two bits of feedback. Let me know what you think!

src/deadline_worker_agent/capabilities.py
AWS-Samuel: Looking closer, returning an empty list will cause an error further on, so I'm okay with keeping this as is.

src/deadline_worker_agent/capabilities.py
cilevitz: Added `verbose` parameter in e78b9da.

AWS-Samuel (approved):

moorec-aws:
Hey @cilevitz, 

Looks like the code quality checks are failing. Can you please push up a fix? Be sure to run `hatch run lint && hatch run test` locally before pushing up changes.

Thanks!

crowecawcaw:
We had a second person encounter the same issue this week. Since the PR hasn't had recent activity and since we wanted to get a fix in promptly, I got the fix working and passing tests in https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/574 based on your code here. Thanks for contributing!

PR #541, opened 2025-02-07
style: switch to using ruff formatter

justinsaws:
We are removing `black` as our formatter since we are already using `ruff` for linting.

Re-format the code with `ruff`

Stylistic changes only.

`hatch run test`

No.

No.

----

### DIFF
--- a/test/unit/test_metrics.py
+++ b/test/unit/test_metrics.py
@@ -280,8 +280,8 @@ class TestHostMetricsLogger:
         ) -> None:
             # GIVEN
             DECIMAL_NUMBER_PATTERN = r"\d+(?:\.\d+)?"
+            # fmt: off
             EXPECTED_LOG_MESSAGE_PATTERN = " ".join(
-                # fmt: off
                 [
                     "cpu-usage-percent", DECIMAL_NUMBER_PATTERN,
                     "memory-total-bytes", DECIMAL_NUMBER_PATTERN,
@@ -297,8 +297,8 @@ class TestHostMetricsLogger:
                     "disk-read-bytes-per-second", rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE|NOT_SUPPORTED)",
                     "disk-write-bytes-per-second", rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE|NOT_SUPPORTED)",
                 ]
-                # fmt: on
             )
+            # fmt: on
             logger = logging.getLogger(__name__)
             caplog.set_level(0, logger.name)
             host_metrics_logger = HostMetricsLogger(logger=logger, interval_s=1)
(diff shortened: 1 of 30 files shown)


test/unit/test_metrics.py
```diff
@@ -283,19 +283,32 @@ def test_log_metrics_correct_encoding(
...
-                    "total-disk-bytes", DECIMAL_NUMBER_PATTERN,
-                    "total-disk-used-bytes", DECIMAL_NUMBER_PATTERN,
-                    "total-disk-used-percent", DECIMAL_NUMBER_PATTERN,
-                    "user-disk-available-bytes", DECIMAL_NUMBER_PATTERN,
-                    "network-sent-bytes-per-second", rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE)",
-                    "network-recv-bytes-per-second", rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE)",
-                    "disk-read-bytes-per-second", rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE|NOT_SUPPORTED)",
-                    "disk-write-bytes-per-second", rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE|NOT_SUPPORTED)",
+                    "cpu-usage-percent",
+                    DECIMAL_NUMBER_PATTERN,
```
epmog: We'll probably want to turn off formatting for this, just like we were doing for black

test/unit/test_metrics.py
```diff
@@ -283,19 +283,32 @@ def test_log_metrics_correct_encoding(
...
+                    "network-sent-bytes-per-second",
+                    rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE)",
+                    "network-recv-bytes-per-second",
+                    rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE)",
+                    "disk-read-bytes-per-second",
+                    rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE|NOT_SUPPORTED)",
+                    "disk-write-bytes-per-second",
+                    rf"(?:{DECIMAL_NUMBER_PATTERN}|NOT_AVAILABLE|NOT_SUPPORTED)",
                 ]
                 # fmt: on
```
jusiskin: It looks like the `# fmt: off` works a little differently with `ruff` ([docs](justinsaws/ruff_formatter_swap)). In particular:

>   `# fmt: on` and `# fmt: off` comments are enforced at the statement level

I think we need to move them before/after the entire statement here to preserve our pairing of arguments

jusiskin (dismissed):
just one nitpick to help preserve some overridden formatting, but not merge blocking.

test/unit/test_metrics.py
justinsaws: So, I took a look at this and ruff is supposed to recognize the black off/on commands so I'm not sure why this isn't working

test/unit/test_metrics.py
justinsaws: Fixed in latest revision.

epmog (approved):

jusiskin (approved):

PR #563, opened 2025-02-28
fix: install-deadline-worker on Windows creates session root directory without read and traversal permissions for Users

moorec-aws:
When trying to create folders with `mkdir` using bash on Windows in the session root directory, traversal permissions are required. We recently started creating the session root directory in the worker installer, the permissions are currently too restrictive to allow for this workflow.

Add the `Users` group to the session root dir with the following permissions to allow Traversal and Read to support `mkdir ` on bash.
```
FILE_LIST_DIRECTORY
FILE_TRAVERSE
FILE_GENERIC_READ
```

files can be created in the session rootdir using `mkdir` on bash

1. Made the changes, built the package and installed it on a Windows host.
2. Installed the Worker Agent and confirmed the permissions were as intended

UsersPermissions

4. Submitted a job with this template to confirm that `mkdir` now works.
```
specificationVersion: 'jobtemplate-2023-09'
name: OutPutTest
parameterDefinitions:
- name: OutputDir
  type: PATH
  objectType: DIRECTORY
  dataFlow: OUT
  userInterface:
    control: CHOOSE_DIRECTORY
    label: Output Directory
  default: "./output"
steps:
- name: WelcomeToAWSDeadlineCloud
  script:
    actions:
      onRun:
        command: bash
        args: ['{{Task.File.Run}}']
    embeddedFiles:
      - name: Run
        type: TEXT
        data: |
          set -xeuo pipefail

          mkdir -p '{{Param.OutputDir}}'

          echo "Hello World" > '{{Param.OutputDir}}/test.txt'

          cat '{{Param.OutputDir}}/test.txt'
```
6. Confirmed the permissions of the per-session directory to make sure that Users permissions were not being inherited.
ConfirmSessionDIrPermissions

No

No

----

### DIFF
--- a/src/deadline_worker_agent/file_system_operations.py
+++ b/src/deadline_worker_agent/file_system_operations.py
@@ -15,6 +15,7 @@ class FileSystemPermissionEnum(Enum):
     EXECUTE = "EXECUTE"
     READ_WRITE = "READ_WRITE"
     FULL_CONTROL = "FULL_CONTROL"
+    LIST_DIRECTORY_AND_READ = "LIST_DIRECTORY_AND_READ"
 
 
 def set_permissions(
@@ -82,6 +83,7 @@ def _set_windows_permissions(
     group: Optional[str] = None,
     group_permission: Optional[FileSystemPermissionEnum] = None,
     agent_user_permission: Optional[FileSystemPermissionEnum] = None,
+    users_group_permission: Optional[FileSystemPermissionEnum] = None,
 ):
     import win32security
     import ntsecuritycon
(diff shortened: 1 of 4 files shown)


jericht (dismissed):

moorec-aws:
I scoped down the permission since `FILE_GENERIC_READ` was already including some of the others I was listing:

https://learn.microsoft.com/en-us/windows/win32/fileio/file-security-and-access-rights

jusiskin (approved):
Great work. Looks good to me.

Can we slightly adjust the PR title and squashed commit message to focus on the problem that was fixed? How about:

```
fix: install-deadline-worker on Windows creates session root directory without read and traversal permissions for Users
```

jericht (approved):

PR #566, opened 2025-03-04
test: add e2e tests to verify the worker agent and its subprocesses have correct Linux capabilities

evanspearman-a:

We did not have any tests that explicitly checked to see if the worker agent itself and the subprocesses (eg. the job) have the correct Linux capabilities. The worker should only have `cap_kill`, whereas subprocesses should not have any capabilities.

Add said tests

We will have test coverage to ensure capabilities are not accidentally added or removed.

I ran the tests in the file the new tests are defined in on Linux. The tests are skipped on Windows.

N/A

No

----

### DIFF
--- a/test/e2e/test_cap_kill.py
+++ b/test/e2e/test_cap_kill.py
@@ -22,7 +22,9 @@ from e2e.conftest import DeadlineResources
 
 
 @pytest.fixture
-def sleep_job_in_bg_pid(session_worker: EC2InstanceWorker) -> Generator[int, None, None]:
+def sleep_job_in_bg_pid(
+    session_worker: EC2InstanceWorker,
+) -> Generator[int, None, None]:
     """Context manager that runs a sleep command in the background and yields the process ID of the
     sleep process. The context-manager will do a best-effort to kill the sleep job when exiting the
     context"""
@@ -56,7 +58,7 @@ def sleep_job_in_bg_pid(session_worker: EC2InstanceWorker) -> Generator[int, Non
 
 
 @pytest.mark.skipif(
-    os.environ["OPERATING_SYSTEM"] == "windows",
+    os.environ["OPERATING_SYSTEM"] != "linux",
     reason="Linux specific test",
 )
 @pytest.mark.usefixtures("session_worker")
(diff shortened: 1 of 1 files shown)


AWS-Samuel (dismissed):

test/e2e/test_cap_kill.py
```diff
@@ -140,3 +142,177 @@ def test_cap_kill_not_inherited_by_running_jobs(
...
+                    },
+                },
+            ],
+        },
+    )
+    job.wait_until_complete(client=deadline_client)
+
+    # THEN
+    #
+    # The output of `capsh --print` will look something like this:
```
AWS-Samuel, line 202: Nice job adding the sample output. This is super useful

test/e2e/test_cap_kill.py
```diff
@@ -140,3 +142,177 @@ def test_cap_kill_not_inherited_by_running_jobs(
         logs_client=logs_client,
         expected_pattern=possible_error_message_re_pattern,
     )
+
+
+@pytest.mark.skipif(
+    os.environ["OPERATING_SYSTEM"] == "windows",
```
jusiskin: nit: just a bit more future-proof if/when we support other operating systems

```suggestion
    os.environ["OPERATING_SYSTEM"] != "linux
```

test/e2e/test_cap_kill.py
```diff
@@ -140,3 +142,177 @@ def test_cap_kill_not_inherited_by_running_jobs(
...
+
+    job.assert_single_task_log_contains(
+        deadline_client=deadline_client,
+        logs_client=logs_client,
+        expected_pattern=re.escape("Ambient set =\n"),
+    )
+
+
+@pytest.mark.skipif(
+    os.environ["OPERATING_SYSTEM"] == "windows",
```
jusiskin: nit: same as above:

```suggestion
    os.environ["OPERATING_SYSTEM"] != "linux",
```

test/e2e/test_cap_kill.py
```diff
@@ -140,3 +142,177 @@ def test_cap_kill_not_inherited_by_running_jobs(
...
+    reason="Linux specific test",
+)
+@pytest.mark.usefixtures("session_worker")
+def test_worker_only_has_cap_kill(
+    deadline_client: DeadlineClient,
+    deadline_resources: DeadlineResources,
+) -> None:
+    """Tests to make sure that the worker agent has no capabilities besides cap_kill"""
```
jusiskin: Can we mention that this tests that the worker agent has `CAP_KILL` in its effective and permitted capability sets but NOT its inheritable capability set?

test/e2e/test_cap_kill.py
```diff
@@ -140,3 +142,177 @@ def test_cap_kill_not_inherited_by_running_jobs(
...
+                                # will look like:
+                                #
+                                # ```
+                                # <pid>: =ep
+                                # ```
+                                # whereas the output for a process that only has cap_kill (what the worker is supposed to have)
+                                # will look like this:
+                                # ```
+                                # <pid>: cap_kill=ep
+                                # ```
```
jusiskin, line 296: We should state that this is the expected output when `CAP_KILL` is in the effective and permitted capability sets of the worker agent process. We specifically don't want to see `cap_kill=eip` which would mean the agent has `CAP_KILL` in its inheritable capability set which would result in subprocesses having this capability.

jusiskin (changes requested):

jusiskin (approved):

AWS-Samuel (approved):

PR #574, opened 2025-03-11
fix: unexpected GPU memory configurations crash worker agent

### LINKED ISSUE #536: Bug: Worker agent aborts startup when detecting GPU memory on a disabled GPU

crowecawcaw:
Some GPU configurations cause `nvidia-smi` to return an output that we didn't expect. The unexpected output raises an exception which is uncaught and crashes the worker agent.

Fixes: https://github.com/aws-deadline/deadline-cloud-worker-agent/issues/536
Related: https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/537

Catch exceptions for GPU memory parsing errors.

Fixes a worker agent crash with some GPU configurations.

Unit tests.

n/a

No
----

### DIFF
--- a/src/deadline_worker_agent/capabilities.py
+++ b/src/deadline_worker_agent/capabilities.py
@@ -120,10 +120,15 @@ def _get_gpu_memory(*, verbose: bool = True) -> int:
 
         mem_per_gpu: list[int] = []
         for line in output.splitlines():
-            mem_mib = int(line.replace("MiB", ""))
-            mem_per_gpu.append(mem_mib)
+            try:
+                mem_mib = int(line.replace("MiB", ""))
+                mem_per_gpu.append(mem_mib)
+            except:
+                # If there's a parsing error on a line, skip it
+                pass
 
-        min_memory = min(mem_per_gpu)
+        # If no line had a valid memory amount, default to 0
+        min_memory = min(mem_per_gpu, default=0)
 
         if verbose:
             _logger.info("Minimum total memory of all GPUs: %s", min_memory)
--- a/src/deadline_worker_agent/capabilities.py
+++ b/src/deadline_worker_agent/capabilities.py
@@ -116,6 +116,18 @@ def _get_gpu_memory(*, verbose: bool = True) -> int:
         output_bytes = subprocess.check_output(
             ["nvidia-smi", "--query-gpu=memory.total", "--format=csv,noheader"]
         )
+        output = output_bytes.decode().strip()
+
+        mem_per_gpu: list[int] = []
+        for line in output.splitlines():
+            mem_mib = int(line.replace("MiB", ""))
+            mem_per_gpu.append(mem_mib)
+
+        min_memory = min(mem_per_gpu)
+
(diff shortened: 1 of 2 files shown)


src/deadline_worker_agent/capabilities.py
```diff
@@ -116,6 +116,18 @@ def _get_gpu_memory(*, verbose: bool = True) -> int:
...
+        mem_per_gpu: list[int] = []
+        for line in output.splitlines():
+            mem_mib = int(line.replace("MiB", ""))
+            mem_per_gpu.append(mem_mib)
+
+        min_memory = min(mem_per_gpu)
+
+        if verbose:
+            _logger.info("Minimum total memory of all GPUs: %s", min_memory)
+        return min_memory
```
jusiskin: Because we have the loop inside the `try`/`except` block, a single GPU with `[N/A]` will cause this function to report 0 when in actuality there may be a GPU available with memory.

I believe we'd want to instead report the minimum memory of all available GPU.

jusiskin (changes requested):

jusiskin (approved):

AWS-Samuel (approved):

PR #575, opened 2025-03-12
test: increase job length when testing session action cancellation

lucaseck:
It looks like there's a race condition in one of the canary tests that failed in the latest run because a session action that the test expected to be cancelled managed to finish running before it was actually cancelled.
```
if expected_canceled_action == "envEnter":
                        # If we canceled the envEnter, everything else should have been NEVER_ATTEMPTED
>                       assert session_action["status"] == "NEVER_ATTEMPTED"
E                       AssertionError: assert 'SUCCEEDED' == 'NEVER_ATTEMPTED'
E                         
E                         - NEVER_ATTEMPTED
E                         + SUCCEEDED
```
Increase the runtime from 40 seconds to 5 minutes(300 seconds) to give additional time for the tests to check that the job has been created, sessions exist and then cancel the job which should cancel the running session action all before the action is able to complete

Also cleaned up the naming and logging of some functions since it was unclear if we were waiting for the job to just be created or actually running
Less flaky tests hopefully
It wasn't
No
No
----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -95,7 +95,7 @@ class TestJobSubmission:
             )
             def is_job_started(current_job: Job) -> bool:
                 current_job.refresh_job_info(client=deadline_client)
-                LOG.info(f"Waiting for job {current_job.id} to be created")
+                LOG.info(f"Waiting for job {current_job.id} to be created and running")
 
                 assert current_job.task_run_status not in [
                     TaskStatus.INTERRUPTING,
@@ -183,7 +183,7 @@ class TestJobSubmission:
             )
             def is_job_started(current_job: Job) -> bool:
                 current_job.refresh_job_info(client=deadline_client)
-                LOG.info(f"Waiting for job {current_job.id} to be created")
+                LOG.info(f"Waiting for job {current_job.id} to be created and running")
 
                 assert current_job.task_run_status not in [
                     TaskStatus.INTERRUPTING,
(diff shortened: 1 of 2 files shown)


epmog (dismissed):

AWS-Samuel (dismissed):

epmog (approved):
LGTM ship it!

test/e2e/test_job_submissions.py
```diff
@@ -1241,7 +1241,7 @@ def test_worker_reports_never_attempted_tasks_if_task_is_canceled(
         )
         def is_job_started_with_sessions(current_job: Job) -> bool:
             current_job.refresh_job_info(client=deadline_client)
-            LOG.info(f"Waiting for job {current_job.id} to be created")
+            LOG.info(f"Waiting for job {current_job.id} to be created and running")
```
AWS-Samuel, line 1244: This one was technically fine, but because more logs are correct than before this change I'm not gonna block this PR on it

AWS-Samuel (approved):

test/e2e/test_job_submissions.py
lucaseck: By virtue of it having sessions are we not waiting for it to also be running?

epmog: Yea, the new one is more correct

AWS-Samuel: Ah yeah you're right, I didn't see the sessions stuff

PR #580, opened 2025-03-14
test: Adding GetLog permissions to the queue role for newly deployed test resources

hindleym:
- The queue role for the deployed test resources was missing the `GetLogEvents` permission.

- I added the `logs:GetLogEvents` permission to the deployed queue role in the `e2e_testing_infrastructure.yaml`

- Adding this permission to the queue role means that newly created test resources will have the required permissions to view the job logs in the Deadline Cloud Monitor to help with debugging tests.

- I deployed the new resources, confirmed the policy was built correctly, and confirmed that job logs could be viewed from the Monitor. Other E2E tests and integration tests are passing. Nothing changed here will impact them.

- N/A

- No

----

### DIFF
--- a/scripts/e2e_testing_infrastructure.yaml
+++ b/scripts/e2e_testing_infrastructure.yaml
@@ -560,6 +560,15 @@ Resources:
                     - ''
                     - - !GetAtt JobAttachmentsBucket.Arn
                       - /*
+              - Effect: Allow
+                Action:
+                  - logs:GetLogEvents
+                Resource:
+                  - !Join
+                    - ''
+                    - - !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/deadline/
+                      - !GetAtt DeadlineFarm.FarmId
+                      - /*
       Tags:
         - Key: Origin
           Value: DeadlineCloudAgentE2ETest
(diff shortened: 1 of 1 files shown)


scripts/e2e_testing_infrastructure.yaml
```diff
@@ -560,6 +560,15 @@ Resources:
...
                       - /*
+              - Effect: Allow
+                Action:
+                  - logs:GetLogEvents
+                Resource:
+                  - !Join
+                    - ''
+                    - - !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/deadline/
+                      - !GetAtt DeadlineFarm.FarmId
+                      - /*
```
jusiskin, line 575: nit: this is currently added to the `JobAttachmentsPermissions` policy. To help model this more semantically correct, we could have a separate policy (e.g. `LogPermissions`)

This is not critical, but I'll leave it up to you whether to adopt that suggestion.

jusiskin (dismissed):
Thanks for the contribution, @hindleym!

I appreciate the attention to detail in scoping down the permissions. Just one minor suggestion which is completely optional.

scripts/e2e_testing_infrastructure.yaml
hindleym: Yeh, I was considering that when I did this. I'll go ahead and make that change. I would rather it be more clear for debugging purpose.

hindleym:
Need to separate this change so I can make other pull requests

PR #590, opened 2025-03-20
test: Verify default Windows worker agent user and permissions

hindleym:
- We want more test coverage around the Windows Worker Agent installer

- I'm adding a test to confirm
  - Default `deadline-worker` is created by the installer
  - `deadline-worker` has admin privileges
  - `deadline-worker` has default Windows Security permissions

- Adding more test coverage to the worker agent for Windows installer

- Ran the tests locally and as part of the E2E tests

- N/A

- No

----

### DIFF
--- a/test/e2e/test_installer.py
+++ b/test/e2e/test_installer.py
@@ -8,7 +8,12 @@ import pytest
 import logging
 import os
 
-from deadline_test_fixtures import EC2InstanceWorker
+from e2e.utils import submit_custom_job
+from deadline_test_fixtures import (
+    DeadlineClient,
+    DeadlineResources,
+    EC2InstanceWorker,
+)
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


test/e2e/test_installer.py
```diff
@@ -30,3 +35,116 @@ def test_installer_shutdown_permission(
...
+                should_exist=False,
+            )
+        finally:
+            # Cleanup the temp directory
+            cmd_result = session_worker.send_command(
+                command='Remove-Item "$env:TEMP\security.cfg" -Force'
+            )
+            assert cmd_result.exit_code == 0, "Failed to cleanup security configuration file"
+
+        retain_instance_job.wait_until_complete(client=deadline_client)
```
YutongLi291, line 150: Should we have this in the "finally" clause? So that even if exceptions do happen we still wait for the job to complete.

hindleym: That makes sense. Although, I don't seem to need any jobs running for this particular test, so I'll get another revision.

hindleym:
Switching focus of these tests to verify a custom worker agent that is created by install. See [PR-594](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/594)

PR #594, opened 2025-04-03
test: Install Windows worker agent with custom worker agent user

hindleym:
Need more tests around the windows worker agent installer and custom worker agent users

Added a test class to the `test_installer.py` that can be iterated on to test variations of the Windows worker agent suer

Increased test time: The class has to spin up its own Windows EC2 instance for test

Tested locally by itself.
Windows E2E tests are running and passing

No

No

----


test/e2e/test_installer.py
```diff
@@ -25,8 +39,151 @@ def test_installer_shutdown_permission(
...
+    DEFAULT_AGENT_NAME = "deadline-worker"
+    DEFAULT_JOB_USER = "job-user"
+
+    WHOAMI_COMMAND = '((whoami).split("\\")[1])'
+
+    # Shared Class Variables
+    install_windows_worker_command: str
+    deadline_client: DeadlineClient
+    deadline_resources: DeadlineResources
+    class_worker: WindowsInstanceBuildWorker
```
leongdl: Question - all tests of this class share the same worker? Will it cause any pollution across test cases?

hindleym: In an attempt to reduce the need to spin up so many instances, I'm trying to knock out a series of related tests with a single instance. There shouldn't be any cross-test pollution. Worst case is that each of them could fail if the instance doesn't get properly deployed, but they'd all fail if that was an issue anyway.

test/e2e/test_installer.py
```diff
@@ -25,8 +39,151 @@ def test_installer_shutdown_permission(
...
+
+    @pytest.fixture(scope="class")
```
YutongLi291, line 56: Instead of this, maybe you could write it as `@pytest.fixture(autouse=True)` and setup all the infrastructure in the class as this fixture should always be used in this class.

YutongLi291 (dismissed):

test/e2e/test_installer.py
hindleym: Added the `autouse=True` flag to the fixture

leongdl (commented):
Looks ok to me, will approve once I give it another pass later today.

test/e2e/conftest.py
```diff
@@ -206,6 +206,25 @@ def _create_function_worker(
...
+    def _create_class_worker(
+        custom_worker_config: DeadlineWorkerConfiguration,
+    ):
+        with create_worker(custom_worker_config, ec2_worker_type, request) as worker:
+            created_workers.append(worker)
+            return worker
+
+    yield _create_class_worker
+    for worker in created_workers:
+        stop_worker(request, worker)
```
jusiskin: I'm not sure this fixture is necessary, since we already have a `class_worker` fixture here:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/0a63684d352f75c09934ba84f6e708490abe352f/test/e2e/conftest.py#L166-L175

All that is required to use this is that your `Test*` class to define an overridden `worker_config` fixture with `scope="class"`.

I could be wrong because I thought there was another place this was done in the test suite, but I'm not finding one.

test/e2e/test_installer.py
```diff
@@ -25,8 +38,147 @@ def test_installer_shutdown_permission(
...
+        request,
+        deadline_client: DeadlineClient,
+        deadline_resources: DeadlineResources,
+        worker_config: DeadlineWorkerConfiguration,
+        class_worker_factory: Callable[[DeadlineWorkerConfiguration], EC2InstanceWorker],
+    ):
+        cls = request.cls
+
+        cls.deadline_client = deadline_client
+        cls.deadline_resources = deadline_resources
```
jusiskin: nit: these do not need to be stored as class attributes if they are available as fixtures. Tests that need them can request them as arguments.

test/e2e/test_installer.py
```diff
@@ -25,8 +38,147 @@ def test_installer_shutdown_permission(
...
+            dataclasses.replace(
+                worker_config,
+                agent_user=self.CUSTOM_AGENT_NAME,
+            )
+        )
+
+    # Shared Class Methods
+    @staticmethod
+    def check_admin_permissions(
+        session: EC2InstanceWorker,
```
jusiskin: I wonder if we could better name of this argument since it refers to a worker. Could we change from `session` &rarr; `worker`?

test/e2e/test_installer.py
```diff
@@ -25,8 +38,147 @@ def test_installer_shutdown_permission(
...
+                username=self.CUSTOM_AGENT_NAME,
+                permissions=["SeServiceLogonRight", "SeAssignPrimary"],
+                should_exist=True,
+            )
+
+            # Check permissions that should not be assigned
+            self.check_security_permissions(
+                session=self.class_worker,
+                username=self.CUSTOM_AGENT_NAME,
+                permissions=["SeShutdown", "SeIncreaseQuota"],
```
jusiskin: As discussed offline....

The intention to have the `SeIncreaseQuota` user right assigned to the agent user based on this code:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/0a63684d352f75c09934ba84f6e708490abe352f/src/deadline_worker_agent/installer/win_installer.py#L912-L913

But when we test this, it does not appear to be present. This could be an issue with the installer code or the test code, but for now maybe we can separate this privilege into it's own test and use [`pytest.mark.xfail`](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) with a reason that we are investigating why the test is failing.

jusiskin (dismissed):
Thanks for contributing this test, @hindleym! It is much appreciated.

There are just a few suggested improvements before we merge it.

test/e2e/conftest.py
hindleym: Removed the factory changes and simplified the test setup to use the existing `class_worker`.

test/e2e/test_installer.py
hindleym: Updated the name

test/e2e/test_installer.py
hindleym: Separate that check into its own test marked with `xfail`

test/e2e/test_installer.py
hindleym: Simplified the test class

leongdl (approved):

YutongLi291 (approved):

test/e2e/conftest.py
leongdl: Reviewed - latest PR uses a class worker.

test/e2e/test_installer.py
leongdl: Addressed in latest revision. Change uses class test function fixtures.

test/e2e/test_installer.py
leongdl: Addressed in latest revision.

test/e2e/test_installer.py
hindleym: The worker agent user gets assigned Admin and the Administrators group has the Increase Quota and Shutdown permission by default. 
Removed the separate test and confirmed that the Admin Group SID existed within those permissions. [test_installer.py - Line 150](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/594/files#diff-2153f4ce914400a337d3645ad9b27e5512af5455a2bf6e8e7e5da34074257760R150)

jusiskin (approved):

PR #601, opened 2025-04-11
feat(experimental): run admin host configuration scripts once worker becomes STARTED

leongdl:
- Deadline Cloud is introducing features to configure a Fleet Worker prior to executing jobs. 
- On linux, scripts are Bash. On Windows, scripts are Powershell.
- On linux, scripts run as `root` to configure a system. On Windows, powershell is launched as an administrative level process.

- Implement logic to process host configuration responses from Deadline Cloud.
- Re-use existing `openjd-sessions` code to materialize files, run the scripts.
- On Windows, there is a de-tour due to Adminstrative UAC prompts. We need to launch processes with the verb `runAs`.
    - Note that for a headless system, Windows must have UAC turned off for admin accounts. Refer to [msdn](https://learn.microsoft.com/en-us/windows/security/application-security/application-control/user-account-control/settings-and-configuration?tabs=intune) documentation, specifically "Behavior of the elevation prompt for administrators in Admin Approval Mode". "0" must be set for ConsentPromptBehaviorAdmin key to remove UAC prompt. 
    - `REG ADD HKLM\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\System /v ConsentPromptBehaviorAdmin /t REG_DWORD /d 0 /f`

- Note that this feature is experimental and gated behind a feature flag. Users must export the environment variable `HOST_CONFIGURATION_FEATURE` to enable the feature. 
  - Usage of this feature is not fully productized. We do not recommend using this feature in production yet. 

- New Worker Agent functionality.

- hatch run build
- hatch run fmt
- hatch run test
- hatch run all:lint
- hatch run e2e on Windows and Linux EC2.
    - E2E test cases are not part of this PR and will be pushed at a later date.

Tested on MacOS and WIndows. 

- Yes.

- No, this feature is backwards compatible.

### Future works
Note that there are a few known caveats in the implementation we are aware of and maybe improved in the future.
- Credentials during the admin script configuration are not rotated. The caller role must have the same or longer expiry time. 
- Environment variables passed to the Worker Agent are done via powershell scripting. We are investigating alternatives with ShellExecuteEx.
- Script execution is not cancellable, as it is critical to worker startup prior to processing jobs. We may improve cancellation from Windows service and the Linux systemd service in the future.

- When a host configuration script execution fails, worker agent exits. This can cause the worker to "start" and re-attempt the configuration script. It is recommended to test the script on a standalone EC2 or Windows machine prior to testing on Deadline. Additional guidance will be provided in the near future.

----

### DIFF
--- a/src/deadline_worker_agent/startup/entrypoint.py
+++ b/src/deadline_worker_agent/startup/entrypoint.py
@@ -15,6 +15,8 @@ from threading import Event
 from typing import Optional
 from pathlib import Path
 
+from ..feature_flag import HOST_CONFIGURATION_FEATURE
+
 from ..api_models import WorkerStatus
 from ..aws.deadline import (
     update_worker,
@@ -30,13 +32,16 @@ from ..linux.capabilities import drop_kill_cap_from_inheritable
 from ..log_messages import (
     AgentInfoLogEvent,
     LogRecordStringTranslationFilter,
+    WorkerHostConfigurationStatus,
     WorkerLogEvent,
     WorkerLogEventOp,
+    WorkerHostConfigurationLogEvent,
 )
 from ..log_sync.cloudwatch import stream_cloudwatch_logs
 from ..log_sync.loggers import ROOT_LOGGER, logger as log_sync_logger
 from ..worker import Worker
 from .bootstrap import bootstrap_worker
+from .host_configuration_script import HostConfigurationScriptRunner
 
 __all__ = ["entrypoint"]
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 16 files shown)


jusiskin (approved):
LGTM.

Great job on this PR, @leongdl! This is great progress towards an important feature that will improve the functionality of Deadline Cloud.

Thanks for documenting the known follow-up items to look into before removing the feature flag and making this more broadly available.

src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -163,6 +168,78 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+                        farm_id=config.farm_id,
+                        fleet_id=config.fleet_id,
+                        worker_id=worker_id,
+                        message="Host Configuration Feature is not enabled.",
+                        status=WorkerHostConfigurationStatus.SKIPPED,
+                    )
+                )
+            elif (
+                worker_bootstrap.host_config
+                and worker_bootstrap.worker_info.host_configuration_succeeded
```
crowecawcaw, line 184: When would this happen?

src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -163,6 +168,78 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+                _logger.info(
+                    WorkerHostConfigurationLogEvent(
+                        farm_id=config.farm_id,
+                        fleet_id=config.fleet_id,
+                        worker_id=worker_id,
+                        message="Host Configuration has been setup before. Not running config scripts.",
+                        status=WorkerHostConfigurationStatus.SKIPPED,
+                    )
+                )
+            # Before the run looop starts, run the Host Configuration script.
```
crowecawcaw, line 195: Nit: looop

src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -163,6 +168,78 @@ def filter(self, record: logging.LogRecord) -> bool:
             # logs that we forward to CloudWatch.
             _log_agent_info()
 
+            # If there was a host config, and it was bootstrapped before, only log a message.
+            if worker_bootstrap.host_config and not HOST_CONFIGURATION_FEATURE:
```
leongdl, line 172: Future improvement - I will refactor this host config block into its own function..

src/deadline_worker_agent/startup/host_configuration_script.py
```diff
@@ -0,0 +1,204 @@
...
+
+        script_file_path = str(self._session_files_directory / script_file_name)
+        return script_file_path
+
+    def run(self) -> int:
+        """
+        Run the host configuration script
+        returns The exit code 0 for success, number otherwise.
+        """
+        if self._host_configuration_script is None:
```
crowecawcaw, line 123: Could we get to this branch if there's no host config? Seems like this would be an error case?

src/deadline_worker_agent/startup/host_configuration_script.py
```diff
@@ -0,0 +1,204 @@
...
+        # Async callback prints out a message based on run state.
+        self._action_event.wait()
+
+        if self._action_state is ActionState.SUCCESS and self.exit_code == 0:
+            return self.exit_code
+        else:
+            return self.exit_code if self.exit_code is not None else -1
+
+    def _run_win32(self, script_file_path: str) -> int:
+        if sys.platform == "win32":
```
crowecawcaw, line 165: Do we need the same check for POSIX? Or why do we need this check here?

src/deadline_worker_agent/startup/host_configuration_script.py
```diff
@@ -0,0 +1,204 @@
...
+                    message="No host configuration script provided.",
+                    status=WorkerHostConfigurationStatus.SKIPPED,
+                )
+            )
+            return 0
+
+        script_file_path = self._write_script_file()
+        with FileContext(script_file_path) as _:
+            if sys.platform == "win32":
+                exit_code = self._run_win32(script_file_path)
```
crowecawcaw, line 138: Should we enforce the timeout here?

crowecawcaw (approved):

src/deadline_worker_agent/startup/entrypoint.py
leongdl: Resolved in touch up follow up PR.

src/deadline_worker_agent/startup/host_configuration_script.py
leongdl: Sure, that's a nice to have improvement in case of refactoring mistakes.

src/deadline_worker_agent/startup/host_configuration_script.py
leongdl: This one is a guard in case the backend returned an empty structure. It technically should not, but just a safety net in case it does. It should do nothing in this case.

(EG: if the backend had a bad logical case all of a sudden)

src/deadline_worker_agent/startup/host_configuration_script.py
leongdl: No - This would lead to the worker agent restarting. Originally I did enforce the timeout at the worker agent level as well. 

What would happen is this:
1. The worker agent script times out, agent aborts and exits with -1.
2. Since the agent is setup as a daemon, it restarts and reports back to the backend it is "started" again.
3. The backend would think the agent crashed, and continue to monitor its health check time accordingly
4. Worker Agent runs again, and probably encounters the same error. 

We decided to always let the backend determine when a worker has timed out. This allows the backend better control of scaling and monitoring bad fleets in general. 

Mind you, customers can change this behavior if they so choose to, and as a service we will not support this change as per host config tenets.

PR #603, opened 2025-04-16
test: Verify deny shutdown on stop keeps instance

hindleym:
- There were no tests that made sure we weren't shutting down a customers workstation when it shouldn't be

- Added a test to ensure that the instance remains running
  - Modified the Windows Installer class worker to use the autoscaling resources
  - Reformatted the test class to have a job submitted immediately, so there is a job for the worker to do when the service is started.
  - Wait for the worker service to begin stopping and confirm that the instance remains running
- Updated the [deadline-cloudtest-fixtures/worker.py](https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/185) to better handle the `start-service` flag being set to false. 
  - (This is required for the tests here to work)

- Adds test coverage for circumstances when a customer doesn't want their workstation to shutdown when the job is complete.
- The [deadline-cloudtest-fixtures/worker.py](https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/185) had to be updated to handle the `start-service` flag for these tests
- Simplified the method for obtaining the shutdown_status from the `.toml` file and moved it to `utils.py`
  - Works with both Windows and Linux and is used in `test_installer.py` and `test_worker_config.py`

- Extensively tested the `TestWindowsInstaller` and `TestWorkerConfiguration` classes against the local `deadline-cloud-test-fixtures` changes I made.
- Windows and Linux E2E tests

No

- Required First: https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/185

----

### DIFF
--- a/test/e2e/test_installer.py
+++ b/test/e2e/test_installer.py
@@ -71,7 +71,7 @@ class TestWindowsInstaller:
             start_service=False,
             fleet=deadline_resources.scaling_fleet,
         )
-    
+
     @pytest.fixture(scope="class")
     def test_job(
         self,
@@ -204,9 +204,9 @@ Get-LocalUser | Select-Object Name, Enabled | Format-Table -AutoSize
         assert test_job.task_run_status == TaskStatus.SUCCEEDED
 
     def test_deny_shutdown_on_stop(
-          self,
-          class_worker: EC2InstanceWorker,
-          test_job: Job,
+        self,
+        class_worker: EC2InstanceWorker,
+        test_job: Job,
     ) -> None:
         # Check if the job has run for the set of tests
         if test_job.task_run_status != TaskStatus.SUCCEEDED:
@@ -215,11 +215,10 @@ Get-LocalUser | Select-Object Name, Enabled | Format-Table -AutoSize
 
         LOG.info("Wait for Worker Service to begin Stopping")
         # This can take over 5 minutes
-        class_worker.wait_until_worker_stopping(
-            seconds_between_checks=25
-        )
-  
+        class_worker.wait_until_worker_stopping(seconds_between_checks=25)
+
         ec2_client = boto3.client("ec2")
+
         @backoff.on_exception(
             backoff.constant,
             Exception,
@@ -230,10 +229,10 @@ Get-LocalUser | Select-Object Name, Enabled | Format-Table -AutoSize
             instance_status = ec2_client.describe_instance_status(
                 InstanceIds=[class_worker.instance_id], IncludeAllInstances=True
             )["InstanceStatuses"][0]["InstanceState"]
-            if instance_status['Name'] != "running":
+            if instance_status["Name"] != "running":
                 LOG.warning(f"Instance is not running, current state: {instance_status['Name']}")
                 return  # Exit the function early
-            
+
             cmd_result = class_worker.send_command(
                 command="""
 $content = Get-Content "C:\\ProgramData\\Amazon\\Deadline\\Logs\\worker-agent.log"
@@ -241,9 +240,7 @@ $pattern = "NOT shutting down the host"
 $content | Select-String -Pattern $pattern
 """
             )
-            assert cmd_result.exit_code == 0, (
-                "Failed to get shutdown status from worker-agent.log"
-            )
+            assert cmd_result.exit_code == 0, "Failed to get shutdown status from worker-agent.log"
             assert "NOT shutting down the host" in cmd_result.stdout, (
                 "Worker Agent should not be shutting down the host"
             )
--- a/test/e2e/test_installer.py
+++ b/test/e2e/test_installer.py
@@ -5,13 +5,17 @@ Deadline Cloud worker and checking that the result/output of the worker agent is
 """
 
 import pytest
+import backoff
 import boto3
 import botocore
 import dataclasses
 import logging
 import os
 
-from e2e.utils import submit_custom_job
+from e2e.utils import (
+    get_shutdown_on_stop_status_from_toml,
+    submit_custom_job,
+)
 from e2e.conftest import DeadlineResources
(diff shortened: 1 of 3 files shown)


hindleym:
Code Quality failures due to needing the test-fixture changes.

test/e2e/test_installer.py
```diff
@@ -87,8 +109,8 @@ def check_security_permissions(
         for permission in permissions:
             cmd_result = worker.send_command(
                 command=f"""
-secedit /export /cfg "$env:TEMP\security.cfg" | Out-Null
-Get-Content "$env:TEMP\security.cfg" | Select-String "{permission}"
+secedit /export /cfg "$env:TEMP\\security.cfg" | Out-Null
```
leongdl, line 112: I think these lines need UAC elevated access now. From a normal cmd prompt and power shell they fail.

Also wouldn't the `\\` fail from before this commit?

hindleym: The double-slash `\\` is because my previous PR caused warnings about invalid escape characters, like `\s`.

hindleym: This test is still occurring as part of the e2e-tests and is running as admin.

test/e2e/test_installer.py
```diff
@@ -206,4 +201,54 @@ def test_custom_agent_runs_job_as_user(
...
+        test_job: Job,
+    ) -> None:
+        # Check if the job has run for the set of tests
+        if test_job.task_run_status != TaskStatus.SUCCEEDED:
+            LOG.info("Job hasn't been completed, start the worker service")
+            class_worker.start_worker_service()
+
+        LOG.info("Wait for Worker Service to begin Stopping")
+        # This can take over 5 minutes
+        class_worker.wait_until_worker_stopping(seconds_between_checks=25)
```
hindleym: This will be updated to use `class_worker.wait_until_desired_worker_status()`

leongdl (commented):
Looks ok, please rebase after fixture PR approval. I'm waiting for the linting and QA stuff to pass then we can approve.

test/e2e/test_installer.py
```diff
@@ -206,4 +201,56 @@ def test_custom_agent_runs_job_as_user(
...
+        test_job: Job,
+    ) -> None:
+        # Check if the job has run for the set of tests
+        if test_job.task_run_status != TaskStatus.SUCCEEDED:
+            LOG.info("Job hasn't been completed, start the worker service")
+            class_worker.start_worker_service()
+
+        LOG.info("Wait for Worker Service to begin Stopping")
+        # This can take over 5 minutes
+        class_worker.wait_until_desired_worker_status(
```
hindleym, line 218: This will require a release of the test-fixtures before it is available

test/e2e/test_installer.py
```diff
@@ -206,4 +201,56 @@ def test_custom_agent_runs_job_as_user(
             expected_pattern=f"{self.DEFAULT_JOB_USER}",
```
leongdl: nit, is the echo in some identifiable pattern? EG: `The user name is: David`, just in case there are other lines in the log that has  just the word `David` now will pass

test/e2e/test_installer.py
```diff
@@ -206,4 +201,56 @@ def test_custom_agent_runs_job_as_user(
...
-        assert job_result.task_run_status == TaskStatus.SUCCEEDED
+        assert test_job.task_run_status == TaskStatus.SUCCEEDED
+
+    def test_deny_shutdown_on_stop(
+        self,
+        class_worker: EC2InstanceWorker,
+        test_job: Job,
+    ) -> None:
+        # Check if the job has run for the set of tests
+        if test_job.task_run_status != TaskStatus.SUCCEEDED:
```
leongdl, line 212: This is interesting, is there a condition where the worker isn't started? If there is, maybe the test_job fixture should wait for the job to be completed? 

It seems odd the tests may or may not be dependent on a successful job here.

hindleym: I did not add extra content to the expected output. I can do that pretty easily.

hindleym: I added this check because while I was integrating on these changes, I had marked the other tests as skipped, and the test job for this class hadn't been completed from the previous test. I figured I'd keep it, so that this test can still run independently.

test/e2e/test_installer.py
hindleym: I modified the script to include `Jobs Run As:` and confirm that exists with the output now too.

test/e2e/test_installer.py
leongdl: Yeah, in this case you can have a different fixture that waits for the job completion. Just a nit comment.

leongdl (approved):

test/e2e/test_installer.py
```diff
@@ -183,27 +183,74 @@ def test_no_default_worker_agent_user(
...
+            LOG.info("Job hasn't been completed, start the worker service")
+            class_worker.start_worker_service()
+
+        LOG.info("Wait for Worker Service to begin Stopping")
```
ttblanchard, line 216: Is there potential for a timeout here if there are multiple jobs in the queue? We are only waiting for the test job here, what about other jobs left around?

hindleym: There's only the one job as part of these tests, so there shouldn't be an instance where there are other jobs that haven't been completed. 

The only time I've had unwanted jobs show up in my queue was when I was force cancelling my test runs during my iterations. In which case, the fleet picked those jobs up first and still waited until this test job was complete.

test/e2e/test_installer.py
hindleym: Ahh, noted. I've got a follow-up PR that builds on this, I can see about making this change there.

ttblanchard (approved):

PR #604, opened 2025-04-19
refactor: Code cleanup and address comments from PR-601

leongdl:
- I took a raincheck on my prior PR which was getting too large.
- https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/601

- Addressing some comments, cleanups, and small refactoring to improve code maintainability.
- No logical changes in this PR.

### Changes
- Updating to "test:" commit message
- Added the better `assert` as suggested.
- Will follow up with a better ACE test once I do more reading.

- Improve on the code!

- hatch build
- hatch run test
- hatch run integ-test
- hatch run e2e-test (+ unmerged host config features e2e test)

- New methods have doc strings. No other functionality based chnages.

- No.

----

### DIFF
--- a/test/integ/startup/test_host_configuration.py
+++ b/test/integ/startup/test_host_configuration.py
@@ -285,8 +285,11 @@ Get-ChildItem env: | ForEach-Object { "$($_.Name)=$($_.Value)" }
         for log in expected_logs:
             assert any(log in m for m in messages)
 
-    def test_script_file_access(self, queue_handler: QueueHandler, tmp_path: Path):
-        """Tests the script file is only accessible by worker agent user."""
+    def test_script_and_log_file_access(self, queue_handler: QueueHandler, tmp_path: Path):
+        """
+        Tests the script file is only accessible by worker agent user.
+        On Windows, also check the log file is only accessible by worker agent user.
+        """
 
         # Given
         runner = self._create_host_configuration_script_runner(
@@ -309,10 +312,6 @@ Get-ChildItem env: | ForEach-Object { "$($_.Name)=$($_.Value)" }
             assert not mode & 0o002
             assert not mode & 0o001
         else:
-            import getpass
-            import win32con
-            import win32security
-            import ntsecuritycon
             from deadline_worker_agent.windows.win_admin_runner import _WindowsScriptRunner
 
             # Need to prepare the file permissions on Windows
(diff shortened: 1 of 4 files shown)


jusiskin:
From the PR description:

>   No logical changes in this PR.

We should not be using `fix:` in the PR/commit title as it will show up in the changelog as a bug fix. Instead, let's use `refactor:` or `chore:`.

jericht (commented):
LGTM, just need the commit message fix @jusiskin mentioned please

leongdl:
> From the PR description:
> 
> > No logical changes in this PR.
> 
> We should not be using `fix:` in the PR/commit title as it will show up in the changelog as a bug fix. Instead, let's use `refactor:` or `chore:`.

Sounds good, let me change it.

jericht (dismissed):

test/integ/startup/test_host_configuration.py
```diff
@@ -323,51 +322,54 @@ def test_script_file_access(self, queue_handler: QueueHandler, tmp_path: Path):
...
+    # Iterate through each ACE in the DACL
+    for i in range(dacl.GetAceCount()):
+        ace = dacl.GetAce(i)
+        (ace_type, ace_flags), ace_mask, sid = ace
+
+        if sid == users_group_sid:
+            # Check for read or write permission
+            if (ace_mask & ntsecuritycon.FILE_GENERIC_READ == ntsecuritycon.FILE_GENERIC_READ) or (
+                ace_mask & ntsecuritycon.FILE_GENERIC_WRITE == ntsecuritycon.FILE_GENERIC_WRITE
+            ):
```
jusiskin, line 362: This PR is just about reorganization of the code, so feel free to consider this out-of-scope, but....

I wonder if this test is sufficient.

`FILE_GENERIC_READ`/`FILE_GENERIC_WRITE` are actually bit-wise combinations of lower-level permissions. See:

https://github.com/mhammond/pywin32/blob/825834883a198e9b56301a4cda1694477169a488/win32/Lib/ntsecuritycon.py#L37-L51

We are only failing the test if all of the bits in that permission group are set, but not individual bits. Similarly, it ignores other permission bits that may be granted beyond least-privilege.

Can we modify the test to make sure that only the least-privilege bits are granted? AFAIK the "Users" group should not have any permissions bit, so couldn't we check that there isn't any ACE for the Users group?

test/unit/startup/test_entrypoint.py
```diff
@@ -905,3 +905,4 @@ def test_fleet_host_config(
         ]
         assert len(all_args) > 0
         assert expected in all_args
+        sys_exit_mock.assert_called_once()
```
jusiskin: Should we test the exit code being supplied here?

```suggestion
        sys_exit_mock.assert_called_once_with(1)
```

jusiskin (changes requested):
Good job cleaning up the code. I have two suggestions for your consideration.

test/unit/startup/test_entrypoint.py
leongdl: Good suggestion! Changed.

test/integ/startup/test_host_configuration.py
leongdl: I need to do some research on this and then re-implement it as I am not so familiar with ACE. I'll take it as a follow up of the 2nd PRs I have lined up.

leongdl:
> From the PR description:
> 
> > No logical changes in this PR.
> 
> We should not be using `fix:` in the PR/commit title as it will show up in the changelog as a bug fix. Instead, let's use `refactor:` or `chore:`.

Updated to `refactor:`

test/integ/startup/test_host_configuration.py
jusiskin: Sounds good. Let's just make sure we improve the tests before removing the feature flag.

jusiskin (approved):

jericht (approved):

PR #606, opened 2025-04-25
feat(experimental): add support for chunkInt parameter type

ryyakobe:
OpenJobDescription has introduced an [RFC for Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) to address application and scene loading overhead. This RFC proposes a mechanism for chunking in the Open Job Description job template specification to improve resource utilization. The worker agent needs to support the new `CHUNK[INT]` parameter type to enable this functionality. This PR is the first step toward supporting task chunking in the Deadline Cloud worker agent.

Added support for the "chunkInt" parameter type.

This change enables the worker agent to begin supporting the task chunking feature defined in the OpenJobDescription RFC. The feature is still experimental but the changes won't impact existing stable functionality.

[x] `hatch build && hatch run fmt && hatch run lint && hatch run test`
[x] Tested with sample job using the new parameter type to verify proper parsing and handling. Verified that task run session actions using this new parameter type run correctly.

I couldn't find documentation for existing parameter types. Since the new type is still experimental and could change in the future, we might not need to document it for now. However, we should consider adding a section for "Supported Parameter Types" to `docs/worker_api_protocol.md`.

No, this is an additive change.

----

### DIFF
--- a/test/unit/sessions/job_entities/test_parameters_from_api_response.py
+++ b/test/unit/sessions/job_entities/test_parameters_from_api_response.py
@@ -1,0 +1,33 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import pytest
+
+from openjd.model import ParameterValueType
+
+from deadline_worker_agent.sessions.job_entities.job_details import parameters_from_api_response
+
+
+class TestParametersFromApiResponse:
+    @pytest.mark.parametrize(
+        "param_name, param_dict, expected_type, expected_value",
+        [
+            ("stringParam", {"string": "value"}, ParameterValueType.STRING, "value"),
+            ("pathParam", {"path": "/path/to/file"}, ParameterValueType.PATH, "/path/to/file"),
+            ("intParam", {"int": "42"}, ParameterValueType.INT, "42"),
+            ("floatParam", {"float": "3.14"}, ParameterValueType.FLOAT, "3.14"),
+            ("chunkIntParam", {"chunkInt": "1-5"}, ParameterValueType.CHUNK_INT, "1-5"),
+        ],
+    )
+    def test_parameters_from_api_response(
+        self, param_name, param_dict, expected_type, expected_value
+    ):
+        # GIVEN
+        params = {param_name: param_dict}
+
+        # WHEN
+        result = parameters_from_api_response(params)
+
+        # THEN
+        assert len(result) == 1
+        assert result[param_name].type == expected_type
+        assert result[param_name].value == expected_value
--- a/src/deadline_worker_agent/api_models.py
+++ b/src/deadline_worker_agent/api_models.py
@@ -73,13 +73,19 @@ class FloatParameter(TypedDict):
     float: str
 
 
+class ChunkIntParameter(TypedDict):
+    chunkInt: str
+
+
 class TaskRunAction(TypedDict):
     sessionActionId: str
     actionType: StepActionType
     taskId: str
     stepId: str
     parameters: NotRequired[
-        dict[str, StringParameter | PathParameter | IntParameter | FloatParameter]
+        dict[
+            str, StringParameter | PathParameter | IntParameter | FloatParameter | ChunkIntParameter
+        ]
     ]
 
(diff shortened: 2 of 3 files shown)


mwiebe (dismissed):

test/unit/scheduler/test_session_queue_chunk_int.py
```diff
@@ -0,0 +1,146 @@
...
+        assert result.task_id == "taskId"
+
+        # Verify the parameters were correctly converted
+        expected_params = {
+            "strP": {"type": ParameterValueType.STRING, "value": "stringValue"},
+            "pathP": {"type": ParameterValueType.PATH, "value": "./test_tmp_dir"},
+            "intP": {"type": ParameterValueType.INT, "value": "12"},
+            "floatP": {"type": ParameterValueType.FLOAT, "value": "1.2"},
+            "chunkP": {"type": ParameterValueType.CHUNK_INT, "value": "1-5"},
+        }
```
AWS-Samuel: Nit, this test would be better structured with each of these parametrized through `pytest.mark.parametrize`.  We could rename the test to simply something like `test_task_parameter_conversion` since it's not specific to chunking

Not blocking

test/unit/sessions/job_entities/test_parameters_from_api_response.py
```diff
@@ -0,0 +1,130 @@
...
+    StringParameter,
+    PathParameter,
+    IntParameter,
+    FloatParameter,
+    ChunkIntParameter,
+)
+
+
+class TestParametersFromApiResponse:
+    def test_parameters_from_api_response_with_chunk_int(self):
```
AWS-Samuel: Same feedback as above

test/unit/sessions/job_entities/test_parameters_from_api_response.py
```diff
@@ -0,0 +1,130 @@
...
+        # Test case 2: Single frame case (should be expressed as a range with same start and end)
+        params2: dict[
+            str,
+            Union[
+                StringParameter, PathParameter, IntParameter, FloatParameter, ChunkIntParameter, str
+            ],
+        ] = {"chunkIntParam": {"chunkInt": "100-100"}}
+        result2 = parameters_from_api_response(params2)
+        assert result2["chunkIntParam"].type == ParameterValueType.CHUNK_INT
+        assert result2["chunkIntParam"].value == "100-100"
```
AWS-Samuel: Another nit, since this is a separate test case it should be it's own test, or parametrized using `pytest.mark.parametrize`. Use can use the `id` kwarg in order to label specific cases like this.  Here's an example in this repo of what that looks in the [temporary credentials tests](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/4c126648a657df83e065c77274c965b8d0ad283e/test/unit/aws_credentials/test_temporary_credentials.py#L310-L318).

test/unit/sessions/job_entities/test_parameters_from_api_response.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_parameters_from_api_response_with_noncontiguous_chunks(self):
```
AWS-Samuel: Same feedback as above. The benefit to parametrizing here is that if something breaks, it doesn't prevent all the other test cases from running. This speeds up iteration if multiple changes would be needed to get all the test cases to pass

AWS-Samuel (commented):
Overall changes look good, but I think we can make the tests easier to read, and more condensed by utilizing some parametrization

test/unit/scheduler/test_session_queue_chunk_int.py
ryyakobe: I removed this whole file, as I was trying to get a bit more than a unit test for my changes. I decided to keep the changes as small as possible, so the test are also minimal now. I parameterized the test as you suggested.

AWS-Samuel (approved):

mwiebe (approved):

PR #608, opened 2025-04-25
feat(HOST_CONFIGURATION_FEATURE): delimit host configuration script with banners in worker logs and shutdown host on failure

leongdl:
- During testing, we found the debugging experience of host configuration to be difficult. The Host Configuration logs were JSON, and it was hard to find the logging section associated with Host Configuration. 

- After team discussion, we decided on 2 changes.
  1. Add logging banners similar to OpenJD Sessions for Python start and end. 
  2. Shutdown the worker host if shutdown is allowed. On Service Managed Fleets shutdown is always allowed.
 
- Short blurb on the script failure experience.
  - Existing experience. When Host Config scripts fail, Worker Agent exits with error code 1. On Service Managed fleets where the agent is a daemon, the worker restarts. This leads to an infinite loop where the Worker Agent starts, runs the script, dies and re-starts.
  - With the updated experience. When Host Config scripts fail, the worker Agent Shuts down the host. This works on Deadline Service Managed Fleets, and Customer Managed Fleets with the shutdown option enabled. If shutdown is not allowed, the previous loop behavior still continues. 
  - Given the premise of this feature, we believe this is the right balance of debugging vs customer annoyance.

- Improve the debuggability of Host Configuration on Deadline Cloud SMF fleets.

- hatch build
- hatch run fmt
- hatch run all:lint
- hatch run test
- hatch run integ-test (windows & OSX)
- hatch run e2e-test (to test shutdown, E2E tests are not publicly published yet.)

- Code comments only.

No

----

### DIFF
--- a/src/deadline_worker_agent/startup/entrypoint.py
+++ b/src/deadline_worker_agent/startup/entrypoint.py
@@ -173,6 +173,7 @@ def entrypoint(cli_args: Optional[list[str]] = None, *, stop: Optional[Event] = 
             # Run Worker Host Configuration.
             _host_configuration(
                 config=config,
+                deadline_client=deadline_client,
                 worker_bootstrap=worker_bootstrap,
                 worker_id=worker_id,
                 session=session,
@@ -432,10 +433,11 @@ def _configure_base_logging(
 
 
 def _host_configuration(
-    worker_bootstrap: WorkerBootstrap,
     config: Configuration,
-    worker_id: str,
+    deadline_client: DeadlineClient,
     session: WorkerBoto3Session,
+    worker_bootstrap: WorkerBootstrap,
+    worker_id: str,
 ) -> None:
     """
     Runs all business logic related to Host Configuration.
@@ -513,7 +515,45 @@ def _host_configuration(
                     exit_code=exit_code,
                 )
             )
-            sys.exit(1)
+            # Set this worker to stopped.
+            try:
+                update_worker(
+                    deadline_client=deadline_client,
+                    farm_id=config.farm_id,
+                    fleet_id=config.fleet_id,
+                    worker_id=worker_id,
+                    status=WorkerStatus.STOPPED,
+                )
+            except Exception as e:
+                _logger.error(
+                    WorkerLogEvent(
+                        op=WorkerLogEventOp.STATUS,
+                        farm_id=config.farm_id,
+                        fleet_id=config.fleet_id,
+                        worker_id=worker_id,
+                        message="Failed to set status to STOPPED: %s" % str(e),
+                    )
+                )
+
+            # Attempt to shutdown if possible.
+            if config.no_shutdown:
+                _logger.info("NOT shutting down the host. Local configuration settings say not to.")
+                sys.exit(1)
+
+            _logger.critical(
+                WorkerHostConfigurationLogEvent(
+                    farm_id=config.farm_id,
+                    fleet_id=config.fleet_id,
+                    worker_id=worker_id,
+                    message="Worker Agent host configuration failed. Attempting host shut down.",
+                    status=WorkerHostConfigurationStatus.FAILED,
+                    exit_code=exit_code,
+                )
+            )
+            while _repeatedly_attempt_host_shutdown():
+                _host_shutdown(config=config)
+                # Sleep for 30s and then try again; hopefully we never wake up
+                sleep(30)
 
 
 def _remove_logging_handler(handler: logging.Handler) -> None:
--- a/test/unit/startup/test_entrypoint.py
+++ b/test/unit/startup/test_entrypoint.py
@@ -835,24 +835,33 @@ def test_host_config_already_run_before(
 
 
 @pytest.mark.parametrize(
-    ("has_host_config", "exit_code"),
+    ("has_host_config", "exit_code", "can_shutdown"),
     (
-        pytest.param(True, 0),
-        pytest.param(True, 1),
-        pytest.param(False, 1),
+        pytest.param(True, 0, True, id="Has HostConfig, run success, can shutdown"),
+        pytest.param(True, 1, True, id="Has HostConfig, run failed, can shutdown"),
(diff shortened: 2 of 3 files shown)


jusiskin:
Can we use the `feat(HOST_CONFIGURATION_FEATURE):` PR and commit title prefix since this is for an experimental feature behind the `HOST_CONFIGURATION_FEATURE` feature flag?

src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -513,7 +515,44 @@ def _host_configuration(
...
+                    WorkerHostConfigurationLogEvent(
+                        farm_id=config.farm_id,
+                        fleet_id=config.fleet_id,
+                        worker_id=worker_id,
+                        message="Worker Agent host configuration failed. Attempting host shut down.",
+                        status=WorkerHostConfigurationStatus.FAILED,
+                        exit_code=exit_code,
+                    )
+                )
+                _host_shutdown(config=config)
```
jusiskin: I think we want this log message to happen before entering the while loop, right? Otherwise it will be repeated for each attempt.

I see the `_host_shutdown` logs for each attempt so we only need the log message here once.

src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -513,7 +515,44 @@ def _host_configuration(
...
-            sys.exit(1)
+            # Set this worker to stopped.
+            try:
+                update_worker(
+                    deadline_client=deadline_client,
+                    farm_id=config.farm_id,
+                    fleet_id=config.fleet_id,
+                    worker_id=worker_id,
+                    status=WorkerStatus.STOPPED,
+                )
```
jusiskin, line 526: Not specific to this line-of-code, but for the feature in general....

We will want to document the new worker protocol changes related to this feature in https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/docs/worker_api_protocol.md

This can certainly be a follow-up since it would increase the scope of this PR.

jusiskin:
> Can we use the `feat(HOST_CONFIGURATION_FEATURE):` PR and commit title prefix since this is for an experimental feature behind the `HOST_CONFIGURATION_FEATURE` feature flag?

I'm thinking of this:

```
feat(HOST_CONFIGURATION_FEATURE): delimit host configuration script with banners in worker logs and shutdown host on failure
```

What do you think?

jusiskin (changes requested):
Looks good mostly. I have two suggestions.

src/deadline_worker_agent/startup/entrypoint.py
leongdl: thanks! I will 100% do this.

src/deadline_worker_agent/startup/entrypoint.py
leongdl: Good point let me move it.

test/unit/startup/test_entrypoint.py
```diff
@@ -905,4 +925,21 @@ def test_fleet_host_config(
...
+
+        # Make sure we called to the backend to STOPPED.
+        update_worker_mock.assert_any_call(
+            deadline_client=ANY,
+            farm_id=configuration_load.return_value.farm_id,
+            fleet_id=configuration_load.return_value.fleet_id,
+            worker_id=worker_id,
+            status=WorkerStatus.STOPPED,
+        )
+        # I it can shutdown, assert that we attempted to call the shutdown method.
```
jericht:
```suggestion
        # If it can shutdown, assert that we attempted to call the shutdown method.
```

jericht (dismissed):
LGTM

test/unit/startup/test_entrypoint.py
jericht: This is the tiniest of tiny nits. Does not need to be addressed unless you need another revision

leongdl:
> Can we use the `feat(HOST_CONFIGURATION_FEATURE):` PR and commit title prefix since this is for an experimental feature behind the `HOST_CONFIGURATION_FEATURE` feature flag?

Done!

leongdl:
> > Can we use the `feat(HOST_CONFIGURATION_FEATURE):` PR and commit title prefix since this is for an experimental feature behind the `HOST_CONFIGURATION_FEATURE` feature flag?
> 
> I'm thinking of this:
> 
> ```
> feat(HOST_CONFIGURATION_FEATURE): delimit host configuration script with banners in worker logs and shutdown host on failure
> ```
> 
> What do you think?

Makes sense. I will update the PR and title.

test/unit/startup/test_entrypoint.py
leongdl: Fixed :)

src/deadline_worker_agent/startup/entrypoint.py
leongdl: My plan is to ship this PR, and the windows file permission test. Get the WA release going then i can focus on writing a better description of the protocol change.

leongdl:
> Can we use the `feat(HOST_CONFIGURATION_FEATURE):` PR and commit title prefix since this is for an experimental feature behind the `HOST_CONFIGURATION_FEATURE` feature flag?

Agreed!

leongdl:
> > Can we use the `feat(HOST_CONFIGURATION_FEATURE):` PR and commit title prefix since this is for an experimental feature behind the `HOST_CONFIGURATION_FEATURE` feature flag?
> 
> I'm thinking of this:
> 
> ```
> feat(HOST_CONFIGURATION_FEATURE): delimit host configuration script with banners in worker logs and shutdown host on failure
> ```
> 
> What do you think?

Agreed, thanks for the suggestion!

jusiskin (approved):

jericht (approved):

src/deadline_worker_agent/startup/entrypoint.py
leongdl: Added docs here: https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/647/files

PR #609, opened 2025-04-27
feat(HOST_CONFIGURATION_FEATURE): Prevent Windows ACL inheritance for host config script and log and grant full control to Administrators

leongdl:
- From my Prior PR, we had an open comment to improve the security stance of the logging and host config script.
- After deep diving, I found the files had inherited access. 
- Also Current User + Admin had a looser file access permission bits.
- The previous implementation also checked for User access as a subset of permission bits.

- Disable File inheritance - See https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-control, SE_DACL_PROTECTED
- Explicit permission set for current user (worker agent), and Administrators
- Explicit check users group is not allowed
- Explicit check no other DACL SIDs exists (eg other users or groups)

### Addendum
- Also updated openjd-sessions to `0.10.2` to incorporate latest version.

- The files involved with Host Configuration are fully locked down.

- hatch build
- hatch run fmt
- hatch run lint
- hatch run test
- hatch run integ-test
- hatch run e2e-test (windows, Host Config test not published yet.)

- No.

- No.

----

### DIFF
--- a/src/deadline_worker_agent/file_system_operations.py
+++ b/src/deadline_worker_agent/file_system_operations.py
@@ -40,6 +40,9 @@ def touch_file(
     user_permission: Optional[FileSystemPermissionEnum] = None,
     permitted_user: Optional[SessionUser] = None,
     agent_user_permission: Optional[FileSystemPermissionEnum] = None,
+    group: Optional[str] = None,
+    group_permission: Optional[FileSystemPermissionEnum] = None,
+    disable_permission_inheritance: bool = False,
 ):
     if os.name == "nt":
         permitted_user = cast(WindowsSessionUser, permitted_user)
@@ -52,6 +55,9 @@ def touch_file(
             user=permitted_user.user if permitted_user else None,
             user_permission=user_permission,
             agent_user_permission=agent_user_permission,
+            group=group,
+            group_permission=group_permission,
+            disable_permission_inheritance=disable_permission_inheritance,
         )
 
 
@@ -84,6 +90,7 @@ def _set_windows_permissions(
     group_permission: Optional[FileSystemPermissionEnum] = None,
     agent_user_permission: Optional[FileSystemPermissionEnum] = None,
     users_group_permission: Optional[FileSystemPermissionEnum] = None,
+    disable_permission_inheritance: bool = False,
 ):
     import win32security
     import ntsecuritycon
@@ -143,6 +150,12 @@ def _set_windows_permissions(
 
     # Get the security descriptor of the object
     sd = win32security.GetFileSecurity(str(path.resolve()), win32security.DACL_SECURITY_INFORMATION)
+
+    # Disable inheritance and convert inherited ACEs to explicit ACEs
+    if disable_permission_inheritance:
+        sd.SetSecurityDescriptorControl(
+            win32security.SE_DACL_PROTECTED, win32security.SE_DACL_PROTECTED
+        )
 
     # Set the security descriptor's DACL to the newly-created DACL
     # Arguments:
--- a/src/deadline_worker_agent/windows/win_admin_runner.py
+++ b/src/deadline_worker_agent/windows/win_admin_runner.py
@@ -124,8 +124,14 @@ class _WindowsScriptRunner:
         touch_file(
             file_path=Path(self._logfile),
             agent_user_permission=FileSystemPermissionEnum.FULL_CONTROL,
+            group="Administrators",
+            group_permission=FileSystemPermissionEnum.FULL_CONTROL,
+            disable_permission_inheritance=True,
         )
         touch_file(
             file_path=Path(self._script_path),
             agent_user_permission=FileSystemPermissionEnum.FULL_CONTROL,
+            group="Administrators",
+            group_permission=FileSystemPermissionEnum.FULL_CONTROL,
+            disable_permission_inheritance=True,
         )
--- a/test/integ/startup/test_host_configuration.py
+++ b/test/integ/startup/test_host_configuration.py
@@ -335,41 +335,82 @@ def _windows_file_permissions_test(file_path: str) -> None:
     import win32security
     import ntsecuritycon
 
-    users_group_sid, _, _ = win32security.LookupAccountName(None, "Users")
+    # Get relevant SIDs
     current_user_sid, _, _ = win32security.LookupAccountName(None, getpass.getuser())
+    administrators_sid, _, _ = win32security.LookupAccountName(None, "Administrators")
+    users_group_sid, _, _ = win32security.LookupAccountName(None, "Users")
+
+    # Get security descriptor
     sd = win32security.GetFileSecurity(
         str(file_path),
         win32con.DACL_SECURITY_INFORMATION | win32con.OWNER_SECURITY_INFORMATION,
     )
     dacl = sd.GetSecurityDescriptorDacl()
 
-    users_group_permission_found: bool = False
-    current_user_permission_read_found: bool = False
-    current_user_permission_write_found: bool = False
-
     if dacl is None:
-        assert False, "All users have access to the file."
+        assert False, "No DACL found - all users have access to the file."
 
-    # Iterate through each ACE in the DACL
+    # Track permissions for allowed entities
+    current_user_permissions = 0
+    admin_permissions = 0
+    other_sids_found = []
+
+    # Explicit check that Users group has no permissions at all
     for i in range(dacl.GetAceCount()):
         ace = dacl.GetAce(i)
         (ace_type, ace_flags), ace_mask, sid = ace
 
+        assert ace_type == ntsecuritycon.ACCESS_ALLOWED_ACE_TYPE, (
+            f"Unexpected ace type found for sid {sid}"
+        )
+
         if sid == users_group_sid:
-            # Check for read or write permission
-            if (ace_mask & ntsecuritycon.FILE_GENERIC_READ == ntsecuritycon.FILE_GENERIC_READ) or (
-                ace_mask & ntsecuritycon.FILE_GENERIC_WRITE == ntsecuritycon.FILE_GENERIC_WRITE
-            ):
-                users_group_permission_found = True
-        elif sid == current_user_sid:
-            # Check for read permission
-            if ace_mask & ntsecuritycon.FILE_GENERIC_READ == ntsecuritycon.FILE_GENERIC_READ:
-                current_user_permission_read_found = True
+            assert False, (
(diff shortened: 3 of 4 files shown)


test/integ/startup/test_host_configuration.py
```diff
@@ -335,41 +335,74 @@ def _windows_file_permissions_test(file_path: str) -> None:
...
+
+    # Check current user permissions
+    assert (current_user_permissions & required_permissions) == required_permissions, (
+        f"Current user missing required permissions. Has: {current_user_permissions}, "
+        f"Needs: {required_permissions}"
+    )
+
+    # Check basic read/write for current user
+    has_read = (current_user_permissions & ntsecuritycon.FILE_GENERIC_READ) != 0
+    has_write = (current_user_permissions & ntsecuritycon.FILE_GENERIC_WRITE) != 0
```
jusiskin: Is this redundant based on lines 380-392?

test/integ/startup/test_host_configuration.py
```diff
@@ -335,41 +335,74 @@ def _windows_file_permissions_test(file_path: str) -> None:
...
+
+    assert has_read, "Current user should have read permissions"
+    assert has_write, "Current user should have write permissions"
+
+    # Check Administrator group has required permissions
+    assert (admin_permissions & required_permissions) != 0, (
+        "Administrators group should have required permissions"
+    )
```
jusiskin: This is just checking for overlap with the required permissions? Is that what we want?

test/integ/startup/test_host_configuration.py
```diff
@@ -335,41 +335,74 @@ def _windows_file_permissions_test(file_path: str) -> None:
...
+    assert not other_sids_found, f"Found unexpected SIDs with access: {other_sids_found}"
+
+    # Define required permissions for current user
+    required_permissions = (
+        ntsecuritycon.FILE_READ_DATA
+        | ntsecuritycon.FILE_READ_ATTRIBUTES
+        | ntsecuritycon.FILE_WRITE_DATA
+        | ntsecuritycon.FILE_APPEND_DATA
+        | ntsecuritycon.FILE_WRITE_ATTRIBUTES
+    )
```
jusiskin: Looking at the functional changes above, we are granting full control. This is only looking at file access permissions. Should we expand the set of expected permissions?

src/deadline_worker_agent/file_system_operations.py
```diff
@@ -144,6 +151,12 @@ def _set_windows_permissions(
     # Get the security descriptor of the object
     sd = win32security.GetFileSecurity(str(path.resolve()), win32security.DACL_SECURITY_INFORMATION)
 
+    # Disable inheritance and convert inherited ACEs to explicit ACEs
+    if disable_permission_inheritance:
+        sd.SetSecurityDescriptorControl(
+            win32security.SE_DACL_PROTECTED, win32security.SE_DACL_PROTECTED
```
jusiskin, line 157: The [documentation](https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-control) I found for this constant says:

>   Prevents the DACL of the security descriptor from being modified by inheritable ACEs.

Is this controlling whether the DACL will inherit ACEs from parent objects in the file-system? Or whether the security descriptor itself can be modified based on ACE inheritance (e.g. parent dir has full control and allows changing this DACL).

jusiskin (changes requested):

src/deadline_worker_agent/file_system_operations.py
leongdl: I believe we are "setting" the file security descriptor from 152 to NOT inherit from the parent. So the former.

To be clear `This ensures that only the permissions explicitly set in your code will apply to the object, and no permissions will be inherited from parent directories.`

But I think this integ test proves itself correct though. 
1. We explicitly check users is not added. 
2. Current user SID is present with the right bits.
3. Administrator group SID is present with the right bits.
4. No other SIDs are detected. 

Based on this, there permissions on the file are "exactly" what we want, otherwise the test will fail.

test/integ/startup/test_host_configuration.py
leongdl: Yeah true, left over from prior implementation when we had more generic checking. I was only thinking if `FILE_GENERIC_READ` had any superset that I may have missed in the `required_permissions` bitset.

Deleted 395 to 399 since they are redundent.

test/integ/startup/test_host_configuration.py
leongdl: For host configs yes - Administrators have sufficient access to read, write, use the file. 

However, since we gave it the input `FileSystemPermissionEnum.FULL_CONTROL.value: ntsecuritycon.GENERIC_ALL`, we can directly match to `ntsecuritycon.GENERIC_ALL`.

See comment above.

test/integ/startup/test_host_configuration.py
leongdl: I prefer not to - this is the minimal set to "use the file" as a worker agent. 

`ntsecuritycon.GENERIC_ALL` is also checked in the `Administrators` Group, so technically the worker agent has the full set.

test/integ/startup/test_host_configuration.py
```diff
@@ -335,41 +335,76 @@ def _windows_file_permissions_test(file_path: str) -> None:
...
+        else:  # We already checked Users group
+            # Keep track of any other SIDs that have access
+            other_sids_found.append((win32security.LookupAccountSid(None, sid)[0], ace_mask))
+
+    # Check that no other SIDs have access
+    assert not other_sids_found, f"Found unexpected SIDs with access: {other_sids_found}"
+
+    # Define required permissions for current user and admin.
+    # This is the scoped down set from checking the underlying application.
+    required_permissions = (
```
leongdl, line 385: Updated - based on understanding `FULL_CONTROL`, this actually translates as follows:

```
Let me compare 2032127 with ntsecuritycon.GENERIC_ALL (0x10000000 or 268435456):

# GENERIC_ALL includes all possible permissions
GENERIC_ALL = 0x10000000  # 268435456

# Current permissions: 2032127
# In binary: 0001 1111 0000 0000 1111 1111

# Missing permissions compared to GENERIC_ALL:
# - MAXIMUM_ALLOWED (0x02000000)
# - ACCESS_SYSTEM_SECURITY (0x01000000)
# - GENERIC_READ (0x80000000)
# - GENERIC_WRITE (0x40000000)
# - GENERIC_EXECUTE (0x20000000)

# The GENERIC_* flags are actually mapped to specific permissions:
# GENERIC_READ maps to FILE_GENERIC_READ
# GENERIC_WRITE maps to FILE_GENERIC_WRITE
# GENERIC_EXECUTE maps to FILE_GENERIC_EXECUTE

The main difference is that GENERIC_ALL includes the generic access rights (GENERIC_READ, GENERIC_WRITE, GENERIC_EXECUTE) which are high-level permission masks that get mapped to specific permissions, while 2032127 has the specific low-level permissions directly.

```

leongdl: I asked Q  to help properly understand this :)

src/deadline_worker_agent/windows/win_admin_runner.py
```diff
@@ -124,8 +124,14 @@ def _prepare_file_permissions(self):
         touch_file(
             file_path=Path(self._logfile),
             agent_user_permission=FileSystemPermissionEnum.FULL_CONTROL,
+            group="Administrators",
```
AWS-Samuel, line 127: Does the group `Administrators` always exist, or is it OS language dependant? If it is language dependant we may want to consider using the [well known SID](https://learn.microsoft.com/en-us/windows/win32/secauthz/well-known-sids) instead

AWS-Samuel (changes requested):

src/deadline_worker_agent/windows/win_admin_runner.py
leongdl: https://learn.microsoft.com/en-us/windows/security/identity-protection/access-control/local-accounts

Yes as per windows `Administrators` always exists.

AWS-Samuel: Thanks for confirming

AWS-Samuel (dismissed):

test/integ/startup/test_host_configuration.py
```diff
@@ -335,41 +335,82 @@ def _windows_file_permissions_test(file_path: str) -> None:
...
+    current_user_permissions = 0
+    admin_permissions = 0
+    other_sids_found = []
+
+    # Explicit check that Users group has no permissions at all
     for i in range(dacl.GetAceCount()):
         ace = dacl.GetAce(i)
         (ace_type, ace_flags), ace_mask, sid = ace
 
+        assert ace_type == ntsecuritycon.ACCESS_ALLOWED_ACE_TYPE, (
```
leongdl, line 363: Allow only `aces`

test/integ/startup/test_host_configuration.py
```diff
@@ -335,41 +335,82 @@ def _windows_file_permissions_test(file_path: str) -> None:
...
-                current_user_permission_write_found = True
-
-    assert not users_group_permission_found
-    assert current_user_permission_read_found
-    assert current_user_permission_write_found
+            assert False, (
+                f"Users group should not have any permissions, but found ACE with mask: {ace_mask}"
+            )
+
+        if sid == current_user_sid:
```
leongdl, line 372: current user and admin only. Any other user asserts. I made a specific `User` group assert since its explicit and very bad.

test/integ/startup/test_host_configuration.py
```diff
@@ -335,41 +335,82 @@ def _windows_file_permissions_test(file_path: str) -> None:
...
+        | ntsecuritycon.FILE_WRITE_ATTRIBUTES  # 0x100
+        | ntsecuritycon.DELETE  # 0x10000
+        | ntsecuritycon.READ_CONTROL  # 0x20000
+        | ntsecuritycon.WRITE_DAC  # 0x40000
+        | ntsecuritycon.WRITE_OWNER  # 0x80000
+        | ntsecuritycon.SYNCHRONIZE  # 0x100000
+    )
+
+    # Check current user permissions
+    assert current_user_permissions == required_permissions, (
```
leongdl, line 403: Only Allowed ACES are scoped exactly to the required permission set.

AWS-Samuel (approved):

jusiskin (approved):
Thanks for your patience and cooperation during this code review. This improves the security of this feature and adds important automated security regression test coverage.

PR #611, opened 2025-05-01
test: User override with jobs run as worker agent tests

hindleym:
We were missing some tests around the functionality of the worker agent when a queue is configured to run jobs as the worker agent user.

- Prior, deployed relevant resources to the ClientSoftware for pipeline testing. Adding JobsRunAsAgentUserQueue
- Updated the E2E testing infrastructure scripts
- Added a few tests the `test_override_job_user.py`
  - Verify Windows jobs do not run as worker agent user
  - Verify if a windows job user is provided, that user takes priority over the worker agent user
  - Verify that jobs can run as the linux worker agent

- Adds more test coverage of the variations of worker agent configurations we provide
- Uses up Queue quotas for the related accounts
- Developers will need to redeploy test resources

- Verified locally that both Windows and Linux tests run as expected

N/A

Technically, it's breaking for people running the tests locally. Users will need to redeploy the test resources so the CFN properly updates.

----


test/e2e/config-template.sh
```diff
@@ -89,6 +89,7 @@ export FARM_ID
 # Required - The ID of the queues to use
 export QUEUE_A_ID
 export QUEUE_B_ID
+export JOBS_RUN_AS_AGENT_USER_QUEUE_ID
```
leongdl, line 92: You will also need to update the underlying CDK infrastructure hosting the tests, this only updates the local at desk test. 

AKA, it'll fail after merging and running on the Client Software pipeline infrastructure.

hindleym: I have made those changes to the Client Software pipeline.

leongdl: ok

scripts/e2e_testing_infrastructure.yaml
```diff
@@ -8,20 +8,20 @@ Description: Infrastructure for running the Deadline Cloud Worker Agent end-to-e
 
 Outputs:
   Account:
-    Value: !Sub '${AWS::AccountId}'
+    Value: !Sub "${AWS::AccountId}"
```
leongdl, line 11: Just a nit question for the worker agent team on formatting. Not sure if this is intended?

leongdl (approved):

ttblanchard (approved):

PR #612, opened 2025-05-05
fix(ASSET_SYNC_JOB_USER_FEATURE): NonValidInputError failures and uploading files from input job attachment directories

godobyte:
This PR fix 2 bugs in the attachment sync

1. when merging multiple manifests in the attachment upload, the root is not source root, rather the mapped local root in WA. As a result it doesn't generate a hash that matches the source root, causing attachment upload basing on hash mapping cannot find the destination and fail to upload.
4. outputRelativeDirectories defined in a job were not handled, causing all files under root and updated detected as diff being uploaded as output.

1. map and pass in source root so `manifest_merge` would handle the hash generation correctly.
2. process outputRelativeDirectories and use `include` in `manifest snapshot` to capture only the files in outputRelativeDirectories.

Handle ASSET_SYNC_JOB_USER_FEATURE edge cases.

Ran a previously failed job successfully.
```
    "taskRunStatus": "SUCCEEDED", 
    "endedAt": "2025-05-05T20:39:25.542Z", 
    "description": "This is a test job for step-step dependencies with job attachments data\nflowing along the dependency edges.\n", 
    "jobId": "job-da9c998ee1bb4555b49812dd51b72222", 
    "taskRunStatusCounts": {
        "SCHEDULED": 0, 
        "NOT_COMPATIBLE": 0, 
        "SUCCEEDED": 8, 
        "CANCELED": 0, 
        "INTERRUPTING": 0, 
        "ASSIGNED": 0, 
        "FAILED": 0, 
        "RUNNING": 0, 
        "SUSPENDED": 0, 
        "READY": 0, 
        "STARTING": 0, 
        "PENDING": 0
    }
```

N/A

No
----


src/deadline_worker_agent/sessions/session.py
```diff
@@ -263,6 +264,16 @@ def add_manifest_path(self, root: str, path: str):
         else:
             self._manifest_paths_by_root[root] = [path]
 
+    @property
+    def manifest_out_rel_dirs_by_source(self) -> dict[str, list[str]]:
+        return self._manifest_out_rel_dirs_by_source
```
jericht, line 270: nit: At a glance it's kinda hard to tell what this property is for. Can we add a comment linking it back to the `outputRelativeDirectories` property of a manifest?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -48,26 +48,32 @@ def merge(manifest_paths_by_root: dict[str, list[str]]) -> dict[str, str]:
         if len(paths) == 1:
             merged_manifests[root] = paths[0]
         else:
+            file_name_list = [os.path.basename(path) for path in paths]
+            # generating the name while keepign the original hash
```
jericht:
```suggestion
            # generating the name while keeping the original hash
```

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -48,26 +48,32 @@ def merge(manifest_paths_by_root: dict[str, list[str]]) -> dict[str, str]:
...
+            name = "_".join(set("_".join(file_name_list).split("_")))
```
jericht:
```suggestion
            name = "_".join(set((file_name_list)))
```

Any reason we're joining by `_` and spitting by it again?

jericht (dismissed):

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -197,18 +197,25 @@ def start(
...
+                manifest_properties: ManifestProperties = ManifestProperties(
+                    rootPath=ja_manifest_properties.root_path,
+                    fileSystemLocationName=ja_manifest_properties.file_system_location_name,
+                    rootPathFormat=PathFormat(ja_manifest_properties.root_path_format),
+                    inputManifestPath=ja_manifest_properties.input_manifest_path,
+                    inputManifestHash=ja_manifest_properties.input_manifest_hash,
+                    outputRelativeDirectories=ja_manifest_properties.output_relative_directories,
+                )
+                manifest_properties_list.append(manifest_properties)
+                session.add_manifest_out_rel_dirs(
```
leongdl, line 211: Would be great to have a comment how this works, It was not immediately clear what this does.

src/deadline_worker_agent/sessions/session.py
leongdl: 100% It took me a while to correlate this

leongdl:
Can this the test case where we found this regression please be added as a Worker Agent E2E test? Ideally it should not get past a merge, and definitely should not pass a release.

godobyte:
> Can this the test case where we found this regression please be added as a Worker Agent E2E test? Ideally it should not get past a merge, and definitely should not pass a release.

I'm in the progress of adding that already.

jusiskin:
Two suggestions around the PR/commit title:

1.  Is there any way we can rephrase the PR/commit title to describe the impact or UX of the bug? I'm not sure someone reading the changelog would
2.  This should be a `fix(...):` not a `feat(...):`, right?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
jusiskin: I'm not sure I understand this comment. Is there a way to generate the name without keeping the original hash?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -48,26 +48,32 @@ def merge(manifest_paths_by_root: dict[str, list[str]]) -> dict[str, str]:
...
             output: Optional[ManifestMerge] = _manifest_merge(
                 root=root,
                 # paths to manifest files to be merged
                 manifest_files=paths,
                 # directory to put the generated merged manifests
                 destination=str(manifest_path),
-                name="merge",
+                name=f"merge-{name}",
```
jusiskin: Is this `name` argument the filename of the merged manifest written in the `destination` directory? Is the point of the change here to avoid collisions when there are multiple roots? What happens when there are lots of manifests for a single root? Could this filename become too big?

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
```diff
@@ -164,9 +169,15 @@ def start(
         )
 
         manifest_paths_by_root = session.manifest_paths_by_root
+        out_rel_dirs_by_source = session.manifest_out_rel_dirs_by_source
+        mapped_include_by_root: dict[str, list[str]] = self._get_include_local(
+            include_by_source=out_rel_dirs_by_source,
```
jusiskin: It could just be me, but I feel like there's a lot of terminology that is being used interchangeably here and it is somewhat hard to follow:

*   The script's input arguments are `--idm` which is short for `--include-dirs-map`
*   Here we refer to the session's `.manifest_out_rel_dirs_by_source`
*   The `_get_include_local` function accepts a `include_by_source` argument which is passed the `out_rel_dirs_by_source`.

Is there any way we can increase the readability/maintainability here - maybe by consolidating our terminology?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: Take `file_name_list = ['40ae15f70df956ca6fb2f5f1cdc6cdfb_job', '40ae15f70df956ca6fb2f5f1cdc6cdfb_step']` as an example, implementation in the PR produces `job_step_40ae15f70df956ca6fb2f5f1cdc6cdfb`, while the suggested one produces `40ae15f70df956ca6fb2f5f1cdc6cdfb_step_40ae15f70df956ca6fb2f5f1cdc6cdfb_job`. The current one seems cleaner, but I don't mind too much as it's short lived under session directory.

jusiskin (changes requested):

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: This doesn't handle multiple roots. Manifest merge handles one root every time. The problem here is that the root here is not source root, rather the mapped local root in WA. As a result it doesn't generate a hash that matches the source root, causing attachment upload basing on hash mapping cannot find the destination and fail to upload. 

Here we maintain the hash and pass in the hash through the prefix name. Normally it's only 2 files, one for job and the other (merged) for step-step dependencies. An example would be `['40ae15f70df956ca6fb2f5f1cdc6cdfb_job', '40ae15f70df956ca6fb2f5f1cdc6cdfb_step']` for the merge, which is what we've written to disk using `_check_and_write_local_manifests` from run attachment download.

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
godobyte: Sure I can do an audit and try to be consistent.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: The original hash here means the hash of the root submission path. Since we're merging under the local root under WA here, and use submission root to compare the hash, we need a way to keep the original submission root hash in the file name. Does that make sense? I can add this to the code comment above in a rev.

godobyte: Actually I noticed the root is only used to generate the hash, an alternative is that instead of passing in WA hash, we should pass in the submission source root path, that probably makes more sense.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
jusiskin: >   This doesn't handle multiple roots.

I was referring to the for loop that this code resides in. At least the code reads as-if it were processing multiple roots - each with potentially multiple manifest files:

https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/8ae11ca611e8fe49066e784861bc89f34cb689e6/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py#L47-L61

Based on my surface level understanding, the loop appears to be generating merged output manifest files in a single manifest directory. I'm wondering why the `name` argument was modified here to include all of the base names of the manifest files.

godobyte: Attachment upload and download relies on file name contains a hash of the source_path when path mapping presents, otherwise they cannot correspond manifest files to the correct destination path.

For merge, the problem here is the root passed in is the WA local mapped mapping (destination_path) rather than the source_path. Here are couple of ways to fix it, the original way was to keep the hash in the name prefix. Thinking about it more, passing in the source_path as root makes more sense, I've updated the PR to use that approach.

jusiskin (approved):

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
leongdl: This comment just says pass it on to attachment upload. But what is it for?

leongdl: To do mapping right?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
jericht: Ah okay, I was just confused at the syntax. Looked weird to join by `_` then split by `_`, seemed like a no-op. It wasn't clear that we wanted to split the filenames with `_` in them as well. We could do something like this to be more clear, but not necessary:

```py
name = "_".join(set([part for file_name in file_name_list for part in file_name.split("_")]))
```

src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
```diff
@@ -84,8 +87,10 @@ def set_step_script(
             "{{ Session.PathMappingRulesFile }}",
             "-s3",
             s3_settings.to_s3_root_uri(),
-            "-mm",
+            "-mp",
             json.dumps(manifest_paths_by_root),
+            "-od",
```
leongdl, line 92: This isn't in the doc string :(

godobyte: Good catch, will add it to the current PR I'm working on.

leongdl (approved):
Can add the docs as a follow up since you have another fix coming in.

PR #616, opened 2025-05-08
test: Verify Windows worker agent user credentials

hindleym:
We allow the customer to provide a password for the Windows worker agent user but did not have test automation that confirmed the credentials are being correctly set.

Using Secrets Manager to set the password when installing the worker agent. Then using the powershell command `ValidateCredentials` to confirm the username and password are correct.

More test coverage for the Windows worker agent installer.

Locally ran the individual test, the installer tests, and the e2e-tests

No

No

----

### DIFF
--- a/test/e2e/test_installer.py
+++ b/test/e2e/test_installer.py
@@ -21,6 +21,7 @@ from deadline_test_fixtures import (
     DeadlineClient,
     DeadlineWorkerConfiguration,
     EC2InstanceWorker,
+    WindowsInstanceWorkerBase,
     Job,
     TaskStatus,
 )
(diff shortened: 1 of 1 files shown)


test/e2e/test_installer.py
```diff
@@ -86,6 +89,23 @@ def test_job(
             run_script=self.WHOAMI_COMMAND,
         )
 
+    @pytest.fixture(scope="class")
+    def completed_job(
```
hindleym, line 93: Added this fixture based on the comment [here](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/603/files#r2072484350)

jusiskin (approved):

AlexTranAmz (approved):

PR #626, opened 2025-05-12
test: add test to ensure job users cannot access cached worker credentials on Windows

jusiskin:

There was missing test coverage for the worker agent on Windows to ensure that the worker agent secures file-system permission to the cached worker AWS credentials and they are not accessible by job users.

Add a test on Windows that submits a job and attempts to access the cached worker AWS credentials. The test asserts that the job fails.

The test will guard against regressions on the file-system permissions of the cached AWS credentials.

1.  Made a change to the worker agent to allow the job user to access the worker credentials
2.  Ran the newly-added test and ensured it failed
3.  Undo the changes from 1 and re-run the test
4.  Confirmed the test passed

No

No

----

### DIFF
--- a/test/e2e/test_credential_handling.py
+++ b/test/e2e/test_credential_handling.py
@@ -7,17 +7,28 @@ attacker position in a supposed different security boundary.
 """
 
 import logging
+import boto3
+import botocore
 import pytest
 import os
 
-from deadline_test_fixtures import CommandResult, DeadlineWorkerConfiguration, EC2InstanceWorker
+from deadline_test_fixtures import (
+    CommandResult,
+    DeadlineClient,
+    DeadlineWorkerConfiguration,
+    EC2InstanceWorker,
+    Job,
+    TaskStatus,
+)
+
+from .conftest import DeadlineResources
 
 
 @pytest.mark.skipif(
     os.environ["OPERATING_SYSTEM"] == "windows",
     reason="Linux specific test",
 )
-def test_access_worker_credential_file_from_job(
+def test_access_worker_credential_file_from_job_linux(
     session_worker: EC2InstanceWorker,
     worker_config: DeadlineWorkerConfiguration,
 ) -> None:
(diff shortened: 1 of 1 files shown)


jericht (dismissed):

test/e2e/test_credential_handling.py
```diff
@@ -55,6 +64,88 @@ def test_access_worker_credential_file_from_job(
...
+                    },
+                },
+            ],
+        },
+    )
+    # Wait until the job is completed
+    job.wait_until_complete(client=deadline_client)
+
+    # THEN
+    assert job.task_run_status == TaskStatus.FAILED
```
AWS-Samuel, line 152: Should we look for the error message: `Error accessing credential files:` too? Otherwise we are susceptible to this test passing due to anything which could cause the job to fail, such as incorrect powershell syntax

jusiskin: Good feedback! I've added the log check and pushed up the changes. I've re-ran the tests and confirm they pass with the code in `mainline` and fail when granting access to the job user.

AWS-Samuel (approved):

test/e2e/test_credential_handling.py
```diff
@@ -55,6 +66,97 @@ def test_access_worker_credential_file_from_job(
...
+    job.assert_single_task_log_contains(
+        deadline_client=deadline_client,
+        expected_pattern="Error accessing credential files: ",
```
moorec-aws, line 155: Since this is a multi-layer permissions check. Should this be `Error accessing credential files: ` OR `No credential files found in the cache directory` in the event that it makes it passed the first layer of permission checks?

jusiskin: If the code is able to list the contents of the directory but there are no credential files, then the test will fail because `Error accessing credential files: ` is not in the session log.

The `No credential files found in the cache directory` should not happen because the worker is supposed to cache its credentials once it obtains them, but it is more of a clue for those investigating the test failure to understand what went wrong.

moorec-aws (approved):

PR #631, opened 2025-05-15
fix(experimental): Job Attachment repeatedly upload output from previous tasks in the same session

godobyte:
Attachment upload action doesn't record output from previous session actions, they ends up being detected as diff and in subsequent session actions. This led to incorrect files being tracked in output manifest, and being downloaded as job output.

```
# Task Frame 8 Output
2025/05/12 16:39:18-07:00 Found difference at: data_dir/Step1-2.8.out, Status: FileStatus.NEW

# Task Frame 9 Output
2025/05/12 16:39:20-07:00 Found difference at: data_dir/Step1-2.9.out, Status: FileStatus.NEW
2025/05/12 16:39:20-07:00 Found difference at: data_dir/Step1-2.8.out, Status: FileStatus.NEW

# Task Frame 10 Output
2025/05/12 16:39:22-07:00 Found difference at: data_dir/Step1-2.9.out, Status: FileStatus.NEW
2025/05/12 16:39:22-07:00 Found difference at: data_dir/Step1-2.10.out, Status: FileStatus.NEW
2025/05/12 16:39:22-07:00 Found difference at: data_dir/Step1-2.8.out, Status: FileStatus.NEW

# Task Frame 11 Output
2025/05/12 16:39:24-07:00 Found difference at: data_dir/Step1-2.10.out, Status: FileStatus.NEW
2025/05/12 16:39:24-07:00 Found difference at: data_dir/Step1-2.11.out, Status: FileStatus.NEW
2025/05/12 16:39:24-07:00 Found difference at: data_dir/Step1-2.8.out, Status: FileStatus.NEW
2025/05/12 16:39:24-07:00 Found difference at: data_dir/Step1-2.9.out, Status: FileStatus.NEW
```

The solution is to capture the output from previous task runs in the same session before checking diff. This includes
1. Report the written output manifest via ActionOutputCaptureFilter
2. Add that to session class variable via `session.add_manifest_path`
3. As a result of above, attachment_upload script would merge the output manifests from previous task run with input before checking the diff

Upload the correct output files.

```
# Task Frame 8 Output
2025-05-14T23:08:23.082Z Found difference at: data_dir/Step1-2.8.out, Status: FileStatus.NEW

# Task Frame 9 Output
2025-05-14T23:08:24.700Z Found difference at: data_dir/Step1-2.9.out, Status: FileStatus.NEW

# Task Frame 10 Output
2025-05-14T23:08:26.323Z Found difference at: data_dir/Step1-2.10.out, Status: FileStatus.NEW

# Task Frame 11 Output
2025-05-14T23:08:27.444Z Found difference at: data_dir/Step1-2.11.out, Status: FileStatus.NEW
```

N/A

N/A - it's still under feature flag

----


src/deadline_worker_agent/sessions/log_config.py
```diff
@@ -113,6 +116,100 @@ def filter(self, record: logging.LogRecord) -> bool:
...
 
 
+class ActionOutputMessageKind(Enum):
+    JA_SNAPSHOT = "ja_snapshot"
+    JA_UPLOAD = "ja_upload"
+
+
+class ActionOutputCaptureFilter(logging.Filter):
+    _FILTER_MATCHER = re.compile(
+        (
```
leongdl, line 133: Yikes, this regex needs an explaination!

src/deadline_worker_agent/sessions/log_config.py
```diff
@@ -113,6 +116,100 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+        if not hasattr(record, "session_id") or getattr(record, "session_id") != self._session_id:
+            # Not a record for us to process
+            return True
+        if not isinstance(record.msg, str):
+            # If something sends a non-string to the logger (e.g. via logger.exception) then
+            # don't try to string match it.
+            return True
+
+        match = ActionOutputCaptureFilter._FILTER_MATCHER.match(record.msg)
+        if match and match.lastindex is not None:
```
leongdl, line 176: Needs some explanation of this logic matching, it is hard to review.

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1155,6 +1164,22 @@ def _action_updated_impl(
                     taskId=current_action.definition.task_id,
                 )
 
+                def action_output_log_filter_callback(
```
leongdl: Does this method need to be a nested inline method? It was much harder to process .

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1155,6 +1164,22 @@ def _action_updated_impl(
...
+                    """
+                    if message_type == ActionOutputMessageKind.JA_SNAPSHOT:
+                        self.add_manifest_path(root=value["root"], path=value["manifest"])
+
+                if not self._action_output_log_filter:
+                    self._action_output_log_filter = ActionOutputCaptureFilter(
+                        session_id=self.id, callback=action_output_log_filter_callback
+                    )
+
+                OPENJD_LOG.addFilter(self._action_output_log_filter)
```
leongdl, line 1181: Why is 1181 not part of the if statement from 1176?

leongdl: Do we intend to keep an instance of the log filter around and keep adding and removing it?

Or better question is, is this action output log filter a session life time object?

test/unit/sessions/test_log_config.py
```diff
@@ -152,3 +157,219 @@ def test_from_boto_log_provision_error(
...
+    def test_filter_with_non_string_message(self):
+        """Test that the filter handles non-string messages correctly"""
+        # Create a mock callback
+        mock_callback = MagicMock()
+
+        # Create the filter with the callback
+        action_output_capture_filter = ActionOutputCaptureFilter(
+            session_id=self.session_id, callback=mock_callback
+        )
+        self.log_handler.addFilter(action_output_capture_filter)
```
leongdl, line 218: Would we need to reset the log_handler in between tests? I don't see a remove in the tests

src/deadline_worker_agent/sessions/session.py
godobyte: The action output log filter is designed to have session lifetime, which has session id per construction. This filter spans multiple session actions, I didn't find the need to create one per session action, hence reusing it within the same session.

src/deadline_worker_agent/sessions/session.py
godobyte: Are you talking about inline method like below? I didn't go with this since this function will handle more ActionOutputMessageKind (i.e. `JA_UPLOAD`) in the very near future.

```
self._action_output_log_filter = ActionOutputCaptureFilter(
    session_id=self.id,
    callback=lambda message_type, value: (
        self.add_manifest_path(root=value["root"], path=value["manifest"])
        if message_type == ActionOutputMessageKind.JA_SNAPSHOT else None
    )
)
```

test/unit/sessions/test_log_config.py
godobyte: Since the setup_method is run before each test method, each test is already starting with a clean logging environment. The handler along with the filter from the previous test would have been removed at the start of the next test.

src/deadline_worker_agent/sessions/session.py
erico-aws: I think he means having `action_output_log_filter_callback()` defined in the `_action_updated_impl()` function.

test/unit/sessions/test_log_config.py
```diff
@@ -152,3 +157,219 @@ def test_from_boto_log_provision_error(
...
+
+        # Create a log record with non-matching session_id
+        record = logging.LogRecord(
+            name="test_logger",
+            level=logging.INFO,
+            pathname="",
+            lineno=0,
+            msg="Test message with non-matching session_id",
```
erico-aws: I think you want the message here to have a matching regex. We want to make sure that the callback isn't being called because of the session-id being different and not because it doesn't match the pattern.

godobyte: Ah good catch, will update in a rev.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -95,6 +95,7 @@ def snapshot(
         )
         if manifest:
             manifests.append(manifest.manifest)
+            print(f"ja_snapshot: {json.dumps({'root': root, 'manifest': manifest.manifest})}")
```
leongdl, line 98: Are we making that json.dumps into a data structure? Would be nice for the data exchange to be officially sanctioned as a data class. I think I saw it in your other PR?

src/deadline_worker_agent/sessions/session.py
leongdl: Yes, what Eric suggested.

src/deadline_worker_agent/sessions/session.py
leongdl: I see, this pattern maybe easier encapsulated as a Python context. Adding / removing the filter over time may lead it to either get added incorrectly or never removed.

If there's a clearer pattern to create it and add it, that will work too.

test/unit/sessions/test_log_config.py
leongdl: got it - it wasn't implicitly called out in the code :)

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: Yes here is the PR

https://github.com/aws-deadline/deadline-cloud/pull/685

src/deadline_worker_agent/sessions/session.py
godobyte: `full_output_prefix` is a simple join of all the input parameters. I noticed this is how previously done, which implicitly added timestamp from the caller. It's not the best, but seems better than the current one. I can incorporate it.

```
s3_settings.full_output_prefix(
                    farm_id=self.farm_id,
                    queue_id=queue_id,
                    job_id=job_id,
                    step_id=step_id,
                    task_id=task_id,
                    session_action_id=session_action_id_with_time_stamp,
                )
```

godobyte: A context manager would only work if the filter's lifecycle was contained within a
single block of code, but here it needs to span across different method calls and execution paths.

In the OpenJD library, the log filter is attached for the whole session, which means it's added during session initialization and removed during cleanup. I didn't see the need to have this filter to stay that long, hence add and remove specifically for the attachment upload action. We could expand this filter to follow that OpenJD log filter pattern when we see the need.

src/deadline_worker_agent/sessions/session.py
godobyte: I defined it here to be close to where it's actually needed, thinking it could help make the code more readable and easy to process. The nesting might look better once we remove the `if ASSET_SYNC_JOB_USER_FEATURE` feature flag?

erico-aws: Personally, I don't like nesting functions unless it's necessary. I find them too hard to read.

godobyte: Sure I can move it out of `_action_updated_impl()`

erico-aws (approved):

leongdl (approved):

PR #632, opened 2025-05-15
test: add e2e test to verify complex job attachment data flow on Linux

godobyte:
There was missing test coverage for the worker agent working with complex Job Attachment data flows that verifies each task's input and output.

Add the test case to worker e2e tests to improve job attachment w.r.t job run coverage, including 

1. dependencies verification between steps
2. input and data dir check
3. and output file check

The test will guard against regressions on the job attachment lifecycle.

Run the test with ASSET_SYNC_JOB_USER_FEATURE flag turned off

```
260.09s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux
133.48s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux
6.05s teardown test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux
=================== 1 passed, 1 warning in 401.34s (0:06:41) ===================
```

Run the test with ASSET_SYNC_JOB_USER_FEATURE flag turned on
```
220.43s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux
153.29s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux
6.11s teardown test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux
=================== 1 passed, 1 warning in 381.53s (0:06:21) ===================
```

N/A

No

----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -32,13 +32,21 @@ import uuid
 import os
 import configparser
 import tempfile
-from e2e.utils import wait_for_job_output, submit_sleep_job, submit_custom_job
+from e2e.utils import (
+    wait_for_job_output,
+    submit_sleep_job,
+    submit_custom_job,
+    submit_job_from_bundle,
+    verify_output_dir_matches,
+)
 
 
 LOG = logging.getLogger(__name__)
 
 
 class TestJobSubmission:
+    JOB_OUTPUT_PATH = os.path.join(os.getcwd(), "job_output")
+
     def test_success(
         self,
         deadline_resources,
@@ -1900,6 +1908,49 @@ class TestJobSubmission:
         job.wait_until_complete(client=deadline_client)
 
         assert job.task_run_status == TaskStatus.SUCCEEDED
+
+    @pytest.mark.skipif(
+        os.environ["OPERATING_SYSTEM"] == "windows",
+        reason="Linux specific job bundle to test job attachments dependency data flow",
+    )
+    def test_worker_job_attachments_dep_data_flow_linux(
+        self,
+        deadline_resources: DeadlineResources,
+        deadline_client: DeadlineClient,
+        session_worker: EC2InstanceWorker,
+    ) -> None:
+        job_bundle_path: str = os.path.join(
+            os.path.dirname(__file__),
+            "job_attachment_bundle",
+            "dep_data_flow",
+        )
+
+        job = submit_job_from_bundle(
+            deadline_client=deadline_client,
+            farm=deadline_resources.farm,
+            queue=deadline_resources.queue_a,
+            bundle_path=job_bundle_path,
+        )
+
+        job.wait_until_complete(client=deadline_client)
+        assert job.task_run_status == TaskStatus.SUCCEEDED
+
+        # Get job output path
+        os.makedirs(name=self.JOB_OUTPUT_PATH, exist_ok=True)
+        output_root_path = tempfile.mkdtemp(dir=self.JOB_OUTPUT_PATH, prefix="dep_data_flow_linux")
+        output_path: dict[str, list[str]] = wait_for_job_output(
+            job=job,
+            deadline_client=deadline_client,
+            deadline_resources=deadline_resources,
+            output_root_path=output_root_path,
+        )
+        LOG.info(f"output_path dict is: {output_path}")
+
+        # Verify the final output file exists and contains expected content
+        verify_output_dir_matches(
+            reference_dir_path=f"{os.path.dirname(__file__)}/job_attachment_bundle/dep_data_flow/correct_output",
+            output_dir_path=output_root_path + "/data_dir",
+        )
 
     @pytest.mark.skip(reason="Queue role permissions are failing the test during E2E test runs")
     def test_worker_fails_job_attachment_sync_when_non_valid_queue_role(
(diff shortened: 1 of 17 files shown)


test/e2e/test_job_submissions.py
```diff
@@ -1901,6 +1909,49 @@ def test_worker_job_attachments_no_outputs_does_not_fail_job(
 
         assert job.task_run_status == TaskStatus.SUCCEEDED
 
+    @pytest.mark.skipif(
+        os.environ["OPERATING_SYSTEM"] == "windows",
+        reason="Linux specific job bundle to test job attachments dependency data flow",
```
leongdl, line 1914: Will we log a ticket to enable to this for windows too?

test/e2e/utils.py
```diff
@@ -88,6 +104,128 @@ def submit_sleep_job(
...
+    farm: Farm,
+    queue: Queue,
+    bundle_path: str,
+    job_attachments_file_system: str = "COPIED",
+    queue_parameter_definitions: List[dict] = [],
+    max_retries_per_task: Optional[int] = None,
+) -> Job:
+    bundle_path = os.path.normpath(bundle_path)
+    LOG.info(f"Submitting bundle {bundle_path} to farm {farm.id} and queue {queue.id}")
+    yaml_path = bundle_path + "/template.yaml"
```
leongdl: Its not always template.yaml, maybe make it an argument?

test/e2e/utils.py
```diff
@@ -88,6 +104,128 @@ def submit_sleep_job(
...
+    LOG.info(f"Matches: {match}, Mismatches: {mismatch}, Errors: {errors}")
+
+    assert len(match) == len(reference_files), (
+        f"Reference files ({len(reference_files)}) did not equal match ({len(match)})"
+    )
+    assert len(mismatch) == 0, "Number of mismatched files is non-zero"
+    assert len(errors) == 0, "Number of errors is non-zero"
+
+
+def get_all_files_in_dir_rel_path(dir_path: str) -> List[str]:
```
leongdl: Would glob functions be better here?

test/e2e/test_job_submissions.py
godobyte: We've already created one =)

test/e2e/utils.py
godobyte: Updated to use glob now.

test/e2e/test_job_submissions.py
```diff
@@ -32,13 +32,21 @@
...
+    submit_job_from_bundle,
+    verify_output_dir_matches,
+)
 
 
 LOG = logging.getLogger(__name__)
 
 
 class TestJobSubmission:
+    JOB_OUTPUT_PATH = os.path.join(os.getcwd(), "job_output")
```
jericht, line 48: nit: any reason we can't use the [tmp_path fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html) here and make the files in there?

test/e2e/utils.py
```diff
@@ -36,7 +48,12 @@ def wait_for_job_output(
         task_id=None,
     )
     output_paths_by_root = job_output_downloader.get_output_paths_by_root()
+    LOG.info(f"Output paths by root: {job_output_downloader.outputs_by_root}")
     # Download file and place it into the output_paths_by_root
+    if output_root_path is not None:
+        job_output_downloader.set_root_path(
+            list(output_paths_by_root.keys())[0], os.path.abspath(output_root_path)
+        )
```
jericht, line 62: What happens if there is more than one root?

test/e2e/utils.py
```diff
@@ -88,6 +105,142 @@ def submit_sleep_job(
...
+    json_path = f"{bundle_path}/{template_file_name}.json"
+    if os.path.isfile(yaml_path):
+        with open(yaml_path) as f:
+            job_template = yaml.safe_load(f.read())
+    elif os.path.isfile(json_path):
+        with open(json_path) as f:
+            job_template = json.loads(f.read())
+    else:
+        LOG.error(
+            f"Was expecting to find either template.yaml or template.json in directory {bundle_path} but found none."
```
jericht:
```suggestion
            f"Was expecting to find either {template_file_name}.yaml or {template_file_name}.json in directory {bundle_path} but found none."
```

You could also just directly use `yaml_path` and `json_path` variables in here

test/e2e/utils.py
jericht: nit: I think this function is ok, but it could also be done as a "one liner" if we want to trim down on simple functions like this 😄

```py
rel_files = [
    os.path.relpath(f, dir_path)
    for f in glob(os.path.join(dir_path, "**"), recursive=True)
    if os.path.isfile(f)
]
```

I think the code would read easier if this was directly in the calling fn rather than having to go here, but just a personal preference. Maybe a different function name could achieve the same readability?
```py
reference_files = recursively_list_files_as_relative_paths(dir_path="/my/root")
```

test/e2e/test_job_submissions.py
godobyte: I was using this fix path so we could git ignore it after running locally. The sub-directory for specific test cases are using temp to run multiple times.

test/e2e/utils.py
godobyte: Good question, it currently doesn't support more than one root. I've added a check to fail fast.

leongdl (approved):

jericht (approved):

PR #634, opened 2025-05-16
fix: Cleanup host config log file if the file already exist

leongdl:
- Customer Managed Fleet use of worker agent can exit on start if the Host Config log file was pre-created in the incorrect format. 
- The root cause is because the log file is simply amended, and not checked for format or content.

- Keep it simple, delete the log file before running Host Configuration runs.
- This ensures any pre-existing file does not affect the worker agent. 

- Release ready!

- hatch build
- hatch run test
- hatch run integ-test (with and without deleting file)

- NA.

- No.

----

### DIFF
--- a/src/deadline_worker_agent/windows/win_admin_runner.py
+++ b/src/deadline_worker_agent/windows/win_admin_runner.py
@@ -43,9 +43,9 @@ class _WindowsScriptRunner:
         """Run the executable with command. Tails the prescribed log file until process exit.
         Returns the process exit code.
         """
-        self._prepare_file_permissions()
+        with FileContext(file_path=self._logfile, delete_existing=True) as _:
+            self._prepare_file_permissions()
 
-        with FileContext(self._logfile) as _:
             # Run the command, allow the window to show.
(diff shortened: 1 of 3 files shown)


jericht (approved):

src/deadline_worker_agent/windows/win_admin_runner.py
```diff
@@ -43,9 +43,9 @@ def _run(self, executable: str, command: str) -> int:
         """Run the executable with command. Tails the prescribed log file until process exit.
         Returns the process exit code.
         """
-        self._prepare_file_permissions()
+        with FileContext(file_path=self._logfile, delete_existing=True) as _:
```
epmog, line 46: This is only a Windows issue? No changes needed for Linux?

leongdl: Correct - On Linux, logs are re-directed automatically since it builds on OpenJD Sessions. On Windows is special because of "runas" to launch the elevated shell. Windows does not allow std, stderr redirection so I have to use a file.

epmog (approved):

PR #641, opened 2025-05-21
fix(experimental): Job output download takes old session action ouput when there is a rerun

godobyte:
Current Asset Sync s3 path structure
```
"JobAttachments/Manifests/farm-4ccee4852a054838833fbc885e82cdc1/queue-7cb6b66eafdc444786abba40212d8767/job-9b0b892ec7014dbc9d2c66c373b02acb/step-a14b9e263fea4ede80690856f793bb0d/task-a14b9e263fea4ede80690856f793bb0d-0/2025-05-06T02:58:03.824934Z_sessionaction-0536902081374befa072f24fec2055a5-1/0a0c7e23f8a8a1f02645c671fbe29465_output"
```

For attachment upload as job user, we use
```
s3_path = f"{os.environ.get('DEADLINE_FARM_ID')}/{os.environ.get('DEADLINE_QUEUE_ID')}/{os.environ.get('DEADLINE_JOB_ID')}/{os.environ.get('DEADLINE_STEP_ID')}/{os.environ.get('DEADLINE_TASK_ID')}/{os.environ.get('DEADLINE_SESSIONACTION_ID')}"

```

Upon checking further, timestamp is used when multiple sessions are for the same step, the download takes latest one.
https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/download.py#L250

```
# 2. Select all files in the last subfolder (alphabetically) under each "task-{any}" folder.
    for task_folder, files in task_prefixes.items():
        last_subfolder = sorted(
            set(f.split("/")[len(task_folder.split("/"))] for f in files), reverse=True
        )[0]
        manifests_keys
```

That means when there are multiple session actions for the same task, there is not guarantee the latest session action upload result would be used while downloading output.

Add timestamp back.

Downloading job outputs would always find the latest session action to take the output manifest from.

1. Added unit tests
2. Run a job with the code and verified the s3 path `farm-XXX/queue-XXX/job-XXX/step-XXX/task-378b978bd8f5466bb40ad722fcfad1ce-0/2025-05-21T18:40:35.873053Z_sessionaction-9dc2e61e4df349699e9e565423df9a7e-2/`

N/A

No, it's still under a feature flag.

----


test/unit/sessions/actions/scripts/test_attachment_upload.py
```diff
@@ -288,3 +289,49 @@ def test__manifest_snapshot_diff_include(self, mock_manifest_snapshot: Mock):
...
+        # Setup environment variables
+        with patch.dict(
+            os.environ,
+            {
+                "DEADLINE_FARM_ID": "farm-123",
+                "DEADLINE_QUEUE_ID": "queue-456",
+                "DEADLINE_JOB_ID": "job-789",
+                "DEADLINE_STEP_ID": "step-012",
+                "DEADLINE_TASK_ID": "task-345",
+                "DEADLINE_SESSIONACTION_ID": "session-678",
```
ryyakobe: nit: `sessionaction`, not `session`.

test/unit/sessions/actions/scripts/test_attachment_upload.py
```diff
@@ -288,3 +289,49 @@ def test__manifest_snapshot_diff_include(self, mock_manifest_snapshot: Mock):
             diff="/path/to/base/manifest2",
             include=["/path/to/include/dir3/**"],
         )
+
+    @patch("deadline_worker_agent.sessions.actions.scripts.attachment_upload.api.attachment_upload")
+    @patch("deadline_worker_agent.sessions.actions.scripts.attachment_upload.boto3.session.Session")
+    @patch("deadline_worker_agent.sessions.actions.scripts.attachment_upload.datetime")
+    def test_upload_with_environment_variables(
```
ryyakobe, line 296: Is there a way to write some kind of test where we have 2 session actions with different timestamps to verify that indeed the newest one is downloaded?

godobyte: That would be a rerun case for upload and need to be verified via download, which is better fit for WA e2e tests. 

An alternative would be seed the s3 bucket and run download to verify, which should belong to deadline-cloud repo for job attachment testing.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -31,7 +32,21 @@
 
 
 def upload(s3_root_uri: str, path_mapping_rules: str, manifests: list[str]) -> None:
-    s3_path = f"{os.environ.get('DEADLINE_FARM_ID')}/{os.environ.get('DEADLINE_QUEUE_ID')}/{os.environ.get('DEADLINE_JOB_ID')}/{os.environ.get('DEADLINE_STEP_ID')}/{os.environ.get('DEADLINE_TASK_ID')}/{os.environ.get('DEADLINE_SESSIONACTION_ID')}"
+    current_dt = datetime.now(timezone.utc)
+    iso_string = current_dt.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```
leongdl: Question, is this the same format as before? Next time, add a link so its easier for reviewers.

leongdl: https://github.com/search?q=org%3Aaws-deadline+_float_to_iso_datetime_string&type=code

Yes it is, not sure if we want to re-use the utility method? So the format stays consistent in all code bases.

leongdl: Actually, can we use this utility please? https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/models.py#L278

Lets not re-invent the path join :)

godobyte: `full_output_prefix` is a simple join of all the input parameters which doesn't have timestamp in the contract either. I noticed this is how previously done, which implicitly added timestamp from the caller. It's not the best, but seems better than the current one. I can incorporate it.

```
s3_settings.full_output_prefix(
                    farm_id=self.farm_id,
                    queue_id=queue_id,
                    job_id=job_id,
                    step_id=step_id,
                    task_id=task_id,
                    session_action_id=session_action_id_with_time_stamp,
                )
```

godobyte: Actually it didn't work. This full_output_prefix include additional partitions `path/Manifests/farm-123/queue-456/job-789/step-012/task-345/2025-01-01T12:00:00.000000Z_sessionaction-678`, while the attachment upload requires `farm-123/queue-456/job-789/step-012/task-345/2025-01-01T12:00:00.000000Z_sessionaction-678`

If we really want to use logic defined in the deadline-cloud lib, we need to create a new function. We can still merge this PR as-is to fix the issue, and use the new function once it's released from deadline-cloud.

godobyte: I didn't use `_float_to_iso_datetime_string` as it seems internal to deadline-cloud, and only used for asset sync, which means we can probably clean it up after turning on the feature flag.

leongdl: Then define a new utility function we intend to keep, because the file name will be re-used across job attachments code?

godobyte: This PR is for adding timestamp to the session action id to the s3 path, which may not be useful as we go forward.

leongdl: But that doesn't change why we have 1 hard code here and possibly other places though. I care more about not having this duplicated regardless of future changes

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -31,7 +31,23 @@
...
 def upload(s3_root_uri: str, path_mapping_rules: str, manifests: list[str]) -> None:
-    s3_path = f"{os.environ.get('DEADLINE_FARM_ID')}/{os.environ.get('DEADLINE_QUEUE_ID')}/{os.environ.get('DEADLINE_JOB_ID')}/{os.environ.get('DEADLINE_STEP_ID')}/{os.environ.get('DEADLINE_TASK_ID')}/{os.environ.get('DEADLINE_SESSIONACTION_ID')}"
+    # Helper function to get environment variable or raise error if missing
+    def get_env_or_raise(name):
+        value = os.environ.get(name)
+        if value is None:
+            raise ValueError(f"Required environment variable '{name}' is not set")
+        return value
+
+    s3_path = JobAttachmentS3Settings.partial_session_action_manifest_prefix(
```
leongdl, line 41: This is OK for today - but obviously will change for Chunking manifest upload. @erico-aws FYI

leongdl (approved):
OK -but make sure Eric is aware.

ryyakobe (approved):

jericht (approved):

PR #643, opened 2025-05-22
ci: update release workflows to be tag based

moorec-aws:
We want to move to tag based releases. 

Depends on: https://github.com/aws-deadline/.github/pull/39

1. The `Release: Publish` workflow is dispatched automatically on changelog merge via CI and can also now be dispatched manually with a tag input. The release process now uses the tagged commit from the changelog instead of using the `release` branch.
3. Updated the `code_quality` workflow with an options to run against tagged checkouts.
4. `Release: Bump` no longer runs tests, these are run in the `Release: Publish` workflow using the release tag. We are now passing the tag to the e2e test workflows so that tests are run against the tagged release code.
5. We are still pushing to the release branch for now so that we can keep our canaries running on the latest version of our released code. `PushToReleaseBranch` was added which now keeps the release branch up to date with the latest tagged release.

Tag based releases, and the release workflow can be re-run.

Tested here: https://github.com/moorec-aws/deadline-cloud-worker-agent/actions/runs/15217267933

No

No

----


AWS-Samuel (dismissed):

.github/workflows/code_quality.yml
```diff
@@ -20,6 +23,7 @@ jobs:
     with:
       os: ${{ matrix.os }}
       python-version: ${{ matrix.python-version }}
+      ref: ${{inputs.tag}}
```
epmog, line 26: Does this not need to be all different types of inputs (branch/ref)?

moorec-aws: They are optional, if nothing is provided the reusable workflow will checkout HEAD from the default branch (mainline).

AWS-Samuel (approved):

epmog (approved):

PR #613, opened 2025-05-07
docs: dev guide landing page

jusiskin:
### What was the problem/requirement? (What/Why)

Contributors looking to onboard to worker agent development are not met with sufficient documentation to learn and become productive quickly.

### What was the solution? (How)

This is the first in a series of pull requests (intentionally to keep code reviews smaller and more focused) that will introduce developer on-boarding documentation.

This PR creates a new developer documentation landing page and placeholder pages to sections that will be populated in future PRs. The developer landing page provides an overview of a developer on-boarding journey into broken out topics organized into three phases:

1.  Learning
2. Development environment setup
3. Development practices

The phases are further broken down into topics.

As part of this change, the worker API protocol documentation has been relocated in the repository under the new `docs/dev` directory since the protocol documentation targets developer audiences.

### What is the impact of this change?

This provides an outline for the upcoming developer documentation.

### How was this change tested?

Previewed the rendered markdown documentation in GitHub.

### Was this change documented?

This change adds documentation

### Is this a breaking change?

No

----

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

PR #645, opened 2025-05-23
docs: developer environment setup documentation

jusiskin:

Contributors looking to onboard to worker agent development are not met with sufficient documentation to learn and become productive quickly.

In #613, we created a high-level structure for the developer-focused documentation with placeholders for the content. The developer environment setup topic remains undeocumented

Added developer environment setup documentation. This includes:

*   An summary/overview at the beginning including a diagram
*   Prerequisites
*   How to setup an EC2 keypair
*   Setting up an Amazon Linux 2023 EC2 dev instance, including:
    *   the instance launch options
    *   how to connect to the instance
    *   the software to install on the instance
    *   setting up the repo on the instance
    *   setting up a git remote from the local workstation to the instance
*   Setting up a Windows Server 2022 EC2 dev instance
    *   (same as above)
*   Development Workflow Tips
*   Security Best Practices
*   Troubleshooting
*   Next Steps

Developers who are looking to onboard to worker agent development have guided documentation explaining how to setup a developer environment.

Previewed the rendered markdown in GitHub.

This change IS documentation.

No

----


leongdl (approved):
Nice! Was looking for the Windows SSH section for some time. Thanks for adding it.

docs/dev/development_environment_setup.md
```diff
@@ -1,3 +1,387 @@
...
+*   A fork of the `deadline-cloud-worker-agent` repository.
+*   Basic familiarity with Git, Python, bash, and powershell
+*   AWS CLI v2 installed and configured on your local machine
+*   Session Manager plugin for AWS CLI installed (see [installation instructions](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html))
+
+## Creating an EC2 Key Pair
+
+The development environment scripts use a conventional key pair name: `deadline-worker-agent-dev-keypair`. You must create this key pair in your AWS account before using the scripts:
+
+1.  Navigate to the EC2 console in the AWS Management Console
```
leongdl, line 28: nit, any keypair will work. Personally I already have a keypair I use for github and I am re-using it. Generating a new one can be optional. `ssh keygen` works just as well on MAC and AL2023.

jusiskin: Definitely. This is a guided development environment setup that will work if starting from square one, but developers can deviate from the setup if they have a need.

docs/dev/development_environment_setup.md
```diff
@@ -1,3 +1,387 @@
...
+2.  Click "Launch Instance"
+3.  Configure your instance with the following settings:
+    -   Name: `deadline-worker-dev-linux`
+    -   AMI: Amazon Linux 2023 (AL2023)
+    -   Instance type: Recommend `t3.medium` or larger
+    -   Key pair: Select `deadline-worker-agent-dev-keypair` (required)
+    -   Network settings:
+        -   VPC: Select your VPC. You can use the default VPC that comes with your AWS account
+        -   Auto-assign public IP: Enabled
```
justinsaws, line 61: Is a public IP required? If we are using SSM to communicate, we should probably be recommending to not use a public IP and instead use the SSM VPC Endpoint for communication so the instance isn't open to the internet at large.

docs/dev/development_environment_setup.md
```diff
@@ -1,3 +1,387 @@
...
+1.  Navigate to the EC2 console in the AWS Management Console
+2.  Click "Launch Instance"
+3.  Configure your instance with the following settings:
+    -   Name: `deadline-worker-dev-windows`
+    -   AMI: Windows Server 2022 Base
+    -   Instance type: Recommend `t3.large` or larger
+    -   Key pair: Select `deadline-worker-agent-dev-keypair` (required)
+    -   Network settings:
+        -   VPC: Select your VPC
+        -   Auto-assign public IP: Enabled
```
justinsaws, line 172: Same comment as above here, we technically don't need a public IP address.

docs/dev/development_environment_setup.md
jusiskin: Initially I had this turned off, but the instance was unable to reach SSM. It requires setting up a VPC endpoint. Currently the documentation allows using the default VPC that is provided with new AWS accounts, and I didn't want to guide readers to modify the VPC in a way that may affect existing infrastructure using it.

The documentation tells readers to use a security group without any ingress, to protect the instance from inbound external connection attempts. The documentation also gives a hint towards an even more secure and isolated setup in the "Security Best Practices" section but leaves it as an exercise to motivated readers.

Alternatively, we could instruct readers to create a separate VPC, but that would bloat the documentation and increase the scope.

Eventually, I hope to automate a lot of this environment setup with CloudFormation templates and scripting. Would you be okay with leaving this and we can iterate towards a better setup when we automate?

justinsaws: Yeah, that sounds reasonable to me. CloudFormation templates or even a small CDK application that does all this setup for the developer would be ideal but this guide is a step in the right direction.

justinsaws (approved):

PR #648, opened 2025-05-26
fix: Add warning messages when persistence directories exist

leongdl:
- From testing, CMF customers are able to create and customize the Worker Agent persistence directory. This can potentially open security loop holes such as allowing any user on the system to view logs, or cached credentials.
- Users may accidentally create the cache folders with general read / write access.
- Note that Deadline Cloud provisions these persistence directories by default with the correct least privilege access. Releases are secure by default. Tests exist and are part of the win_installer tests in integration.

- Instead of overwriting the folder with "exists ok", we will now print a warning.
- There is part of the shared responsibility of running Deadline Worker Agent to ensure the execution environment is secure.
    - For example, a customer debugging a CMF instance may change the cache folder for debugging but forgets to change it back.

- Better messaging for folder access permissions.

- Run the install deadline worker agent script on top of an existing installation. See that the new message is printed to the installation log. For example:
Screenshot 2025-05-26 at 3 19 21 PM

- Not Applicable.

- Not Applicable.

----

### DIFF
--- a/src/deadline_worker_agent/installer/win_installer.py
+++ b/src/deadline_worker_agent/installer/win_installer.py
@@ -435,24 +435,43 @@ def provision_directories(
     logging.info(f"Done provisioning root directory ({deadline_dir})")
 
     deadline_log_subdir = os.path.join(deadline_dir, "Logs")
-    logging.info(f"Provisioning log directory ({deadline_log_subdir})")
-    os.makedirs(deadline_log_subdir, exist_ok=True)
-    logging.info(f"Done provisioning log directory ({deadline_log_subdir})")
+    if os.path.exists(deadline_log_subdir):
+        logging.warning(
+            f"Log directory already exists, permissions maybe inherited for all users to view files in this directory. ({deadline_log_subdir})"
+        )
+    else:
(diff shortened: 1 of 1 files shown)


src/deadline_worker_agent/installer/win_installer.py
```diff
@@ -435,24 +435,43 @@ def provision_directories(
     logging.info(f"Done provisioning root directory ({deadline_dir})")
 
     deadline_log_subdir = os.path.join(deadline_dir, "Logs")
-    logging.info(f"Provisioning log directory ({deadline_log_subdir})")
-    os.makedirs(deadline_log_subdir, exist_ok=True)
-    logging.info(f"Done provisioning log directory ({deadline_log_subdir})")
+    if os.path.exists(deadline_log_subdir):
+        logging.warning(
+            f"Log directory already exists, permissions maybe inherited for all users to view files in this directory. ({deadline_log_subdir})"
```
jericht, line 440: nit
```suggestion
            f"Log directory already exists, permissions may be inherited for all users to view files in this directory. ({deadline_log_subdir})"
```

same for the other logs

jericht (approved):

seant-aws (approved):

PR #660, opened 2025-06-07
feat: support total manifest and output directories length over 1024 during windows asset sync

YutongLi291:
On windows, there is a 1024 character limit when using a process to call commands. This may cause issues during asset sync step in the job as manifest paths and total output directories may exceed this character limit.

Embed these arguments for manifest paths and output directories in an embedded file in the job asset sync step instead.

Windows workers can now run jobs with very many steps and/or output directories without failing during attachment upload.

```
hatch build 
hatch run fmt 

hatch run test
```
In both Linux and windows workers:

``hatch run e2e-test``

Added a e2e test to verify that windows workers are able to run jobs with many steps and output directories.

No
No
----

### DIFF
--- a/src/deadline_worker_agent/api_models.py
+++ b/src/deadline_worker_agent/api_models.py
@@ -416,6 +416,13 @@ class BatchGetJobEntityResponse(TypedDict):
     errors: list[EntityError]
 
 
+class ManifestInfo(TypedDict):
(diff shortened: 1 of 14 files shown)


YutongLi291:
SonarCloud is failing for duplication in code for a test file.

PR #661, opened 2025-06-07
fix(experimental): Error sync inputs due to no path mapping rule found for the source path

godobyte:
When storage profile path mapping doesn't exist and file system location name exists, the current implementation would trigger a error syncing input due to no path mapping found for the source path.

Path map to local working directory for this edge case.

No longer throw error for this case, the job can run successfully.

1. Added unit tests
2. Run WA E2E tests

N/A

No

----


leongdl (approved):

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -236,6 +237,27 @@ def start(
             )
         )
 
+        # TODO - update and formalize path mapping logic
+        # Temporary patch to generate path mapping for manifests with fileSystemLocationName
+        generated_path_mapping: dict[str, PathMappingRule] = dict()
```
leongdl, line 242: Approved - we did an in person code review.

leongdl (approved):

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -259,6 +281,7 @@ def start(
             # successfully launched VFS
             return
 
+        dynamic_mapping_rules.update(generated_path_mapping)
```
jericht, line 284: nit: this could be up in the new `if` block to keep it closer. Also a no-op if the branch is skipped

test/unit/sessions/actions/test_run_attachment_download.py
```diff
@@ -192,3 +201,61 @@ def test_attachment_download_action_start(
...
+        # WHEN
+        action.start(session=session, executor=executor)
+        s3_settings = JobAttachmentS3Settings(
+            s3BucketName=job_details.job_attachment_settings.s3_bucket_name,
+            rootPrefix=job_details.job_attachment_settings.root_prefix,
+        )
+
+        # THEN
+        # Verify _get_unique_dest_dir_name was called with the root path
+        mock_get_unique_dest_dir_name.assert_called_once_with("/foo/bar")
```
jericht, line 236: Is `/foo/bar` from `job_details.job_attachment_settings.root_prefix`?

test/unit/sessions/actions/test_run_attachment_download.py
```diff
@@ -192,3 +201,61 @@ def test_attachment_download_action_start(
...
+
+        # Check specific arguments individually
+        assert call_args[1]["session_dir"] == Path(session_dir)
+        assert call_args[1]["s3_settings"] == s3_settings
+        assert call_args[1]["queue_id"] == TestStart.QUEUE_ID
+        assert call_args[1]["job_id"] == TestStart.JOB_ID
+        assert call_args[1]["step_dependencies"] == []
+        assert call_args[1]["storage_profiles_path_mapping_rules"] == {
+            "/foo/bar": str(session.working_directory.joinpath(TestStart.DIR_NAME))
+        }
```
jericht, line 251: nit: Any reason we didn't use `assert_called_once_with` here like the assertions below?

jericht (approved):

test/unit/sessions/actions/test_run_attachment_download.py
godobyte: It's from job_attachment_manifest_properties root path, which is used for mapping destination.

test/unit/sessions/actions/test_run_attachment_download.py
godobyte: For some reason using ANY for call arg attachments keeps giving error. Decided to check individually as we care mostly about `storage_profiles_path_mapping_rules`.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Good call, it makes more sense to be in the if block, will update in a follow-up when I work on this part of code.

PR #663, opened 2025-06-09
feat(experimental): add output manifest reporting

erico-aws:

We want to report the output manifest s3 location to the service.

The attachment upload script gets the returned manifest infos from `api.attachment_upload()`. It prints out all manifest infos to the logs in the feature flag is enabled. In the session callback, we read the Manifest information and assign it to `_upload_manifest_list`. The updated manifests are cross-referenced with the job attachment manifests. Manifests without an update send an empty dictionary value, otherwise they send the updated output information. 

This allows the service to know where the output manifests are in s3.

Added unit tests.

Also, I've done testing with internal models. I've tested with jobs that produce no output, some output, run as different users and ones that have multiple asset roots. 

No

No
----

### DIFF
--- a/src/deadline_worker_agent/feature_flag.py
+++ b/src/deadline_worker_agent/feature_flag.py
@@ -7,3 +7,6 @@ import os
 ASSET_SYNC_JOB_USER_FEATURE = (
     os.environ.get("ASSET_SYNC_JOB_USER_FEATURE", "false").lower() == "true"
 )
+
+# Feature flag for output manifest reporting
+MANIFEST_REPORTING_FEATURE = os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
--- a/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
+++ b/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
@@ -8,6 +8,7 @@ import os
 import boto3
 import json
 from typing import Optional
+from dataclasses import asdict
 
 from deadline.job_attachments import api
 from deadline.job_attachments.api.manifest import _manifest_snapshot, _manifest_merge
@@ -48,13 +49,22 @@ def upload(s3_root_uri: str, path_mapping_rules: str, manifests: list[str]) -> N
         time=time.time(),
     )
 
-    api.attachment_upload(
+    # Call attachment_upload and get manifest information
+    manifest_infos = api.attachment_upload(
         manifests=manifests,
         s3_root_uri=s3_root_uri,
         boto3_session=boto3.session.Session(),
         path_mapping_rules=path_mapping_rules,
         upload_manifest_path=s3_path,
     )
+
+    # Check if manifest reporting feature is enabled via environment variable
+    manifest_reporting_enabled = (
+        os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
+    )
+
+    if manifest_reporting_enabled:
+        print(f"ja_upload: {json.dumps([asdict(info) for info in manifest_infos])}")
 
 
 def merge(
--- a/src/deadline_worker_agent/sessions/session.py
+++ b/src/deadline_worker_agent/sessions/session.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import json
 from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
@@ -30,8 +31,12 @@ from deadline_worker_agent.api_models import (
     SyncInputJobAttachmentsAction,
     AttachmentDownloadAction,
     AttachmentUploadAction,
+    ManifestInfo,
 )
-from deadline_worker_agent.feature_flag import ASSET_SYNC_JOB_USER_FEATURE
+from deadline_worker_agent.feature_flag import (
+    ASSET_SYNC_JOB_USER_FEATURE,
+    MANIFEST_REPORTING_FEATURE,
+)
 
 if TYPE_CHECKING:
     from ..api_models import CompletedActionStatus, EnvironmentAction, TaskRunAction
@@ -60,6 +65,7 @@ from deadline.job_attachments.models import (
     JobAttachmentS3Settings,
     ManifestProperties,
     PathFormat,
+    UploadManifestInfo,
 )
 from deadline.job_attachments.os_file_permission import (
     FileSystemPermissionSettings,
@@ -169,6 +175,7 @@ class Session:
     _retain_session_dir: bool = False
     _job_details: JobDetails
     _job_attachment_details: JobAttachmentDetails | None = None
+    _upload_manifest_list: list[UploadManifestInfo] = []
 
     # Event that is set only when this Session is not running at all
     # i.e. it has exited, or never started, its main run loop/logic.
@@ -1069,7 +1076,7 @@ class Session:
         ):
             self._action_updated_impl(action_status=action_status, now=now)
 
-    def _action_output_log_filter_callback(
+    def action_output_log_filter_callback(
         self, message_type: ActionOutputMessageKind, value: Any
     ) -> None:
         """Callback for the action output log filter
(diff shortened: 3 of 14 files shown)


src/deadline_worker_agent/feature_flag.py
```diff
@@ -7,3 +7,8 @@
 ASSET_SYNC_JOB_USER_FEATURE = (
     os.environ.get("ASSET_SYNC_JOB_USER_FEATURE", "false").lower() == "true"
 )
+
+HOST_CONFIGURATION_FEATURE = os.environ.get("HOST_CONFIGURATION_FEATURE", "false").lower() == "true"
```
ryyakobe: Is this change intentional?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -48,14 +49,28 @@ def get_env_or_raise(name):
...
     )
 
+    # Check if manifest reporting feature is enabled via environment variable
+    manifest_reporting_enabled = (
+        os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
+    )
+
+    if manifest_reporting_enabled:
+        try:
+            print (f"ja_upload: {json.dumps([asdict(info) for info in manifest_infos])}")
```
ryyakobe: nit: do we need space after `print`? We don't add space on line 71

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -48,14 +49,28 @@ def get_env_or_raise(name):
...
+
+        except Exception as e:
+            print(f"Failed to write manifest information: {e}")
+            # Continue with the upload process even if writing the manifest information fails
```
ryyakobe: If we would always report manifests as suggested in my other comment, then this could become an issue. Outputs might be uploaded but session action would have empty manifests and download code might not even look at s3. 

After typing the above, I noticed that this try/excepts block is just for a print command :) The above paragraph is still interesting to explore, e.g. what to do if manifest upload succeeds but reporting fails. However, for the code at hand, I think we should just remove the try/except block.

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1069,14 +1076,30 @@ def update_action(self, action_status: ActionStatus) -> None:
...
+        ):
+            try:
+                manifest_list_data = json.loads(value)
+                self._upload_manifest_list = [UploadManifestInfo(**item) for item in manifest_list_data]
+            except (TypeError, ValueError) as e:
+                logger.error(
+                    f"Failed to parse manifest snapshot: {e}",
+                    exc_info=True,
+                )
+                self._upload_manifest_list = []
```
ryyakobe, line 1098: Does this mean we would fail silently?

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1240,6 +1299,7 @@ def _on_done_with_sync_asset_outputs(
                 self._action_update_lock,
                 self._current_action_lock,
             ):
+                # Pass manifests to _handle_action_update
```
ryyakobe: Is this comment still relevant?

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1370,7 +1438,8 @@ def _sync_asset_outputs(
         from .actions import RunStepTaskAction
 
         assert isinstance(current_action.definition, RunStepTaskAction)
-        upload_summary_statistics: SummaryStatistics = self._asset_sync.sync_outputs(
```
ryyakobe, line 1374: just curious: why did we remove `SummaryStatistics` type hint?

ryyakobe (changes requested):
Thanks Eric! Changes look great, just a few minor comments.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
erico-aws: I guess I'm not sure what we should do if we fail to report the manifest information (Either print it properly here, or encode it in the session). The work is complete and uploaded so it is technically accessible to the user but you're right that downloading the output may not work.

src/deadline_worker_agent/feature_flag.py
erico-aws: No, I think this was some rebase shenanigans.

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1239,7 +1298,7 @@ def _on_done_with_sync_asset_outputs(
                 #     2.  current action lock
                 self._action_update_lock,
                 self._current_action_lock,
-            ):
+            ):
```
ryyakobe: nit: accidental space?

ryyakobe (dismissed):
Not sure about all the code scanner failures in log_config.py but other changes look good to me! Thanks, Eric!

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1069,14 +1076,26 @@ def update_action(self, action_status: ActionStatus) -> None:
         ):
             self._action_updated_impl(action_status=action_status, now=now)
 
-    def _action_output_log_filter_callback(
+    def action_output_log_filter_callback(
```
godobyte: curious - why are we changing this?

ryyakobe (dismissed):

src/deadline_worker_agent/sessions/session.py
erico-aws: I've switched it back to private.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
erico-aws: Changed this to no longer catch the exception. Any failure to write the manifest information will fail the action.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -48,14 +49,23 @@ def get_env_or_raise(name):
...
         upload_manifest_path=s3_path,
     )
 
+    # Check if manifest reporting feature is enabled via environment variable
+    manifest_reporting_enabled = (
+        os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
+    )
+
+    if manifest_reporting_enabled:
+        print(f"ja_upload: {json.dumps([asdict(info) for info in manifest_infos])}")
```
AWS-Samuel: Can we add a comment explaining why this message is being printed? 

It looks like `ja_upload:` is a key word that is detected in the worker agent log filter, and then we dump the manifest info as `JSON` so that we can re-load it as a manifest info in the worker agent process? 

Without this context it looks like just a debug print message and could accidentally be broken

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1140,6 +1159,42 @@ def _action_updated_impl(
...
+                    # For each input manifest, find the corresponding output manifest
+                    for input_manifest in job_attachment_details.manifests:
+                        asset_root = input_manifest.root_path
+                        manifest_info = None
+
+                        for output_manifest in self._upload_manifest_list:
+                            if (
+                                output_manifest.source_path is not None
+                                and output_manifest.source_path == asset_root
+                            ):
```
AWS-Samuel, line 1180: Is there guaranteed to only be one that matches this?

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1140,6 +1159,42 @@ def _action_updated_impl(
...
+                        manifest_info_obj: ManifestInfo = {}
+
+                        if manifest_info and manifest_info.output_manifest_path is not None:
+                            manifest_info_obj["outputManifestPath"] = (
+                                manifest_info.output_manifest_path
+                            )
+                        if manifest_info and manifest_info.output_manifest_hash is not None:
+                            manifest_info_obj["outputManifestHash"] = (
+                                manifest_info.output_manifest_hash
+                            )
```
AWS-Samuel: What happens if only one of `outputManifestPath` or `outputManifestHash` are in the `manifest_info` object?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
erico-aws: I've added some comments here to make it clearer.

src/deadline_worker_agent/sessions/session.py
erico-aws: Yeah, asset roots are guaranteed to be unique.

src/deadline_worker_agent/sessions/session.py
erico-aws: I think this was an artifact of a previous revision. I've removed some of these checks here to simplify things.

AWS-Samuel (approved):

AWS-Samuel (approved):

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1249,7 +1308,14 @@ def _handle_action_update(
         action_status: ActionStatus,
         current_action: CurrentAction,
         now: datetime,
+        manifests: list[ManifestInfo] | None = None,
     ):
+        # avoid circular import
+        from .actions import RunStepTaskAction
+
+        if manifests is None and isinstance(current_action.definition, RunStepTaskAction):
```
ryyakobe, line 1316: Nice! Thanks for adding this!

ryyakobe (approved):

PR #669, opened 2025-06-13
fix(experimental): ASSET_SYNC_JOB_USER_FEATURE does not attempt to launch VFS

godobyte:
ASSET_SYNC_JOB_USER_FEATURE does not attempt to launch VFS even though the condition should be met. There is a bug in passing the variable for the condition check and launching VFS.

1. Fix the wrong variable being passed in, i.e. `os.environ` vs `session._env`
2. Add echo step script to make sure the session proceeds to next action
3. Remove the temporary code to fall back to previous path for VIRTUAL

ASSET_SYNC_JOB_USER_FEATURE should work with VFS

Built the whl, deploy with AMI and ran the test
```
test_step_dep_bundle[VIRTUAL-linux] passed 1 out of the required 1 times. Success!
```

N/A

No

----

### DIFF
--- a/src/deadline_worker_agent/sessions/actions/run_attachment_download.py
+++ b/src/deadline_worker_agent/sessions/actions/run_attachment_download.py
@@ -310,13 +310,15 @@ class AttachmentDownloadAction(OpenjdAction):
             merged_manifests_by_root=merged_manifests_by_root,
             s3_settings=s3_settings,
         ):
-            # successfully launched VFS, LINUX only, for the session to proceed
+            # Successfully launched VFS, running a echo step with openjd
+            # for the session to proceed to the next action
+            # LINUX and VIRTUAL only
             session.run_task(
                 step_script=StepScript_2023_09(
                     actions=StepActions_2023_09(
                         onRun=Action_2023_09(
                             command="echo",
-                            args=["VFS launched successfully for VIRTUAL"],
+                            args=["Job Attachments mode VIRTUAL, VFS launched"],
                         )
                     ),
                 ),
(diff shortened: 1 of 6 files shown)


src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -312,16 +303,37 @@ def start(
...
-            log_task_banner=False,
-        )
+        ):
+            # successfully launched VFS, LINUX only, for the session to proceed
+            session.run_task(
+                step_script=StepScript_2023_09(
+                    actions=StepActions_2023_09(
+                        onRun=Action_2023_09(
+                            command="echo",
+                            args=["VFS launched successfully for VIRTUAL"],
```
leongdl: nit, Job Attachments mode VIRTUAL, VFS launched. Or something like that. VFS is an acronym.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -312,16 +303,37 @@ def start(
...
             s3_settings=s3_settings,
-        )
-        assert self._step_script is not None
-        session.run_task(
-            step_script=self._step_script,
-            task_parameter_values=dict[str, ParameterValue](),
-            log_task_banner=False,
-        )
+        ):
+            # successfully launched VFS, LINUX only, for the session to proceed
```
leongdl: Nit: maybe needs some explaination.

leongdl (dismissed):
Good fix! We can fix the comments and nits after we get this feature properly completed.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Updated, needed a small fix on Windows linting error anyways.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: Updated, needed a small fix on Windows linting error anyways.

leongdl (dismissed):

leongdl (approved):

justinsaws (approved):

PR #690, opened 2025-07-07
fix: avoid embedded file name conflicts with download/upload script

YutongLi291:
The current asset sync as job user implementation creates two scripts as embedded files. `upload.py` and `download.py`, which are ran during job attachments sync to sync job attachments from/to the worker.

However, the names of these files are explicitly named, and if someone submits a job with embedded files `upload.py` and/or `download.py`, unexpected behaviour would happen, in which the file would get overridden or we could run simply into a permissions issue.

Remove the explicit naming of the `upload.py` and `download.py` files. 

This would make it so that the files for the upload/download scripts are named a randomly generated unique ID. Thus, it would not conflict with any user uploaded files.

Users of deadline cloud will be able to use custom embedded files named `upload.py` and `download.py` without rrunning into any issues

Added a new test that uses embedded files named `upload.py` and `download.py`, as well as job attachments.  Verified that they work with the `ASSET_SYNC_JOB_USER_FEATURE` feature flag on.

`hatch run test`
`hatch run integ-test`

```
source .e2e_linux_infra.sh
hatch run e2e-test


source .e2e_windows_infra.sh
hatch run e2e-test

```

Screenshot 2025-07-07 at 3 07 07 PM
image

No
No
----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -3026,3 +3026,164 @@ echo -n $(cat {{Param.DataDir}}/files/test_input_file)Hello > {{Param.DataDir}}/
             )
 
             job.wait_until_complete(client=deadline_client)
+
+    def test_job_submission_asset_sync_behaviour_expected_without_errors(
+        self,
+        deadline_resources,
+        session_worker: EC2InstanceWorker,
+        deadline_client: DeadlineClient,
+        tmp_path: pathlib.Path,
+    ) -> None:
+        """
+        Verify that asset sync as job user works as expected,
+        with manifest cleanup, job attachments, and embedded files working properly.
+        """
+        upload_py_content = """#!/usr/bin/env python3
+print("Upload script executed successfully")"""
+
+        download_py_content = """#!/usr/bin/env python3
+import os
+print("Download script executed successfully")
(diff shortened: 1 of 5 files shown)


godobyte (approved):

test/e2e/test_job_submissions.py
```diff
@@ -3026,3 +3026,164 @@ def worker_stop(worker: EC2InstanceWorker) -> bool:
...
+
+        # Verify job attachments output
+        output_path = wait_for_job_output(
+            job=job, deadline_client=deadline_client, deadline_resources=deadline_resources
+        )
+        output_file = os.path.join(list(output_path.keys())[0], "output.txt")
+        with open(output_file, "r") as f:
+            assert f.read() == "Job attachments working"
+
+        ## TODO: add verification that manifest cleanup completes successfully
```
jericht, line 3189: Are we doing this in a follow up PR?

jericht (approved):

test/e2e/test_job_submissions.py
YutongLi291: Yes, we are doing a fix for this issue and will have a PR out later

PR #695, opened 2025-07-14
fix: allow permissions for session manifest cleanup on linux

YutongLi291:
Session Cleanup Manifest File Deletion Permission denied for the asset sync code path

```rm: cannot remove '/sessions/session-dc5ec925f5794b6c8ffa87b76aa90cbe35u_g4tp/manifests/1889c7447896306816f345f86edb3210_job': Permission denied```

This is because the manifests files are created by worker-agent user, and the session cleanup is run as a job user. As a result it got permission denied for cleanup.

Make is so that manifests are sent to the worker as embedded files instead. Embedded files have job user write/delete permissions, so job user is able to cleanup session manifests properly.

Only the "copied" codepath is fixed, and not the VFS code path, as there is a known bug with VFS fallbacks that would cause the VFS codepath to fail that would need to be fixed first.

Also added a E2E test case 

During copied workflows, session manifest cleanup will run properly and cleanup the manifests without giving permissions errors.

`hatch run test`
`hatch run integ-test`
`hatch run e2e-test` for both Linux and Windows environments
No
No
----

### DIFF
--- a/src/deadline_worker_agent/sessions/actions/run_attachment_download.py
+++ b/src/deadline_worker_agent/sessions/actions/run_attachment_download.py
@@ -9,7 +9,7 @@ import sys
 from pathlib import Path, PurePosixPath, PureWindowsPath
 from logging import LoggerAdapter
 from typing import Any, TYPE_CHECKING, Optional
-from dataclasses import asdict
+from dataclasses import asdict, dataclass
 
 from deadline.job_attachments.asset_manifests import BaseAssetManifest
 from deadline.job_attachments.models import (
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/deadline_worker_agent/sessions/actions/run_attachment_download.py [hunk 1]
```diff
@@ -296,49 +323,72 @@ def start(
...
-                    actions=StepActions_2023_09(
-                        onRun=Action_2023_09(
-                            command=CommandString("echo"),
-                            args=[ArgString("Job Attachments mode VIRTUAL, VFS launched")],
-                        )
-                    ),
-                ),
-                task_parameter_values=dict[str, ParameterValue](),
-                log_task_banner=False,
+        ## If file system is Virtual, try to launch VFS after writing manifests to local paths
```
YutongLi291, line 326: Virtual workflow is the same as before for now.

godobyte:
This a PR is a work-in-progress, I'll take over and work on this later.

leongdl:
Stale PR - closing this PR.

PR #705, opened 2025-07-24
test: add e2e test to verify output sync works as expected when job has no input

godobyte:
We discovered a regression from asset sync as job user feature, where a job doesn't run output sync if it doesn't have input. This was discovered through manual job submission, we should have automated testing to catch issue earlier.

Add the case to automated testing.

Better test coverage.

Ran the e2e test on both windows and linux.

```
185.68s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only
30.79s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only
6.11s teardown test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only
=================== 1 passed, 1 warning in 224.46s (0:03:44) ===================
```

N/A

No

----


AWS-Samuel (approved):
Is there a unit test we could add which would also catch this regression?

godobyte:
> Is there a unit test we could add which would also catch this regression?

Probably. I'll add unit test coverage for this scenario in the upcoming fix PR.

jairaws (approved):

test/e2e/job_attachment_bundle/output_only/template.yaml
```diff
@@ -0,0 +1,57 @@
...
+- name: FilesPerTask
+  type: INT
+  default: 600
+- name: Tasks
+  type: STRING
+  default: 1-600
+- name: FileSize
+  type: INT
+  default: 100
+- name: OutputDir
```
leongdl, line 25: We should have another case where we have an INOUT dataflow, but no input. 

Just to get the correct coverage.

godobyte: I'll sync offline and work on this as a follow-up. Wanted to audit the current coverage before adding more tests.

leongdl: Chatted offline, we need IN, OUT, INOUT (with input data), INOUT (with no input data)

We need the code coverage.

PR #727, opened 2025-09-03
feat(experimental): add worker session data model for job attachment sync operations

godobyte:
The worker agent needed a more structured way to manage job attachment data during session execution, specifically to track local file system paths and manifest information without relying solely on hash-based correlation. 

The existing system lacked a dedicated data model for worker-specific attachment operations, making it difficult to maintain state and coordinate between attachment upload/download operations and session management.

- Added `WorkerManifestProperties` class: A new data structure that extends `ManifestProperties` with worker-specific local path information
- Enhanced Session class: Added methods to manage a dictionary of worker manifest properties keyed by local root paths

- Provides a structured approach to managing worker-specific attachment data during session execution
- Enables better tracking of local file system paths and manifest locations
- Lays groundwork for improved job attachment sync data flow

Added tests for the data model and session integration

Yes added docstrings and inline comments.

No

----

### DIFF
--- a/src/deadline_worker_agent/sessions/attachment_models.py
+++ b/src/deadline_worker_agent/sessions/attachment_models.py
@@ -128,25 +128,9 @@ class WorkerManifestProperties:
         Returns:
             Dict[str, Any]: A dictionary representation of the worker manifest properties
         """
-        # Manually serialize ManifestProperties to avoid dependency on unreleased to_dict method
-        manifest_dict: Dict[str, Any] = {"rootPath": self.manifest_properties.rootPath}
-        manifest_dict["rootPathFormat"] = self.manifest_properties.rootPathFormat.value
-
-        if self.manifest_properties.fileSystemLocationName:
-            manifest_dict["fileSystemLocationName"] = (
-                self.manifest_properties.fileSystemLocationName
-            )
-        if self.manifest_properties.inputManifestPath:
-            manifest_dict["inputManifestPath"] = self.manifest_properties.inputManifestPath
-        if self.manifest_properties.inputManifestHash:
-            manifest_dict["inputManifestHash"] = self.manifest_properties.inputManifestHash
-        if self.manifest_properties.outputRelativeDirectories:
-            manifest_dict["outputRelativeDirectories"] = (
-                self.manifest_properties.outputRelativeDirectories
-            )
 
         return {
-            "manifestProperties": manifest_dict,
+            "manifestProperties": self.manifest_properties.to_dict(),
             "localManifestPaths": self.local_manifest_paths,
             "localRootPath": self.local_root_path,
         }
--- a/src/deadline_worker_agent/sessions/attachment_models.py
+++ b/src/deadline_worker_agent/sessions/attachment_models.py
@@ -141,7 +141,9 @@ class WorkerManifestProperties:
         if self.manifest_properties.inputManifestHash:
             manifest_dict["inputManifestHash"] = self.manifest_properties.inputManifestHash
         if self.manifest_properties.outputRelativeDirectories:
-            manifest_dict["outputRelativeDirectories"] = self.manifest_properties.outputRelativeDirectories
+            manifest_dict["outputRelativeDirectories"] = (
+                self.manifest_properties.outputRelativeDirectories
+            )
 
         return {
             "manifestProperties": manifest_dict,
--- a/src/deadline_worker_agent/sessions/attachment_models.py
+++ b/src/deadline_worker_agent/sessions/attachment_models.py
@@ -34,10 +34,22 @@ class WorkerManifestProperties:
     local_manifest_paths: list[str] = field(default_factory=list)
     """Local file system paths where the manifest files are stored (supports step dependencies)"""
 
-    def __post_init__(self):
-        """Validate that required fields are set after initialization."""
-        if not self.local_root_path:
-            raise ValueError("local_root_path must be set for WorkerManifestProperties")
+    def __init__(
+        self,
+        manifest_properties: ManifestProperties,
+        local_root_path: str,
+        local_manifest_paths: Optional[List[str]] = None,
+    ):
+        """
+        Initialize WorkerManifestProperties.
+        Args:
+            manifest_properties: The original manifest properties
+            local_root_path: Local root path for attachment files
+            local_manifest_paths: Optional list of local paths for manifest files (supports step dependencies)
+        """
+        self.manifest_properties = manifest_properties
+        self.local_root_path = local_root_path
+        self.local_manifest_paths = list(local_manifest_paths) if local_manifest_paths else []
 
     @property
     def root_path(self) -> str:
@@ -74,12 +86,7 @@ class WorkerManifestProperties:
         Convert to an OpenJD-compatible path mapping rule.
         Returns:
             PathMappingRule: A path mapping rule using local root path as destination
-        Raises:
-            ValueError: If local_root_path is empty
         """
-        if not self.local_root_path:
-            raise ValueError("local_root_path must be set to create path mapping rule")
-
         return PathMappingRule(
             source_path_format=self.manifest_properties.rootPathFormat.value,
             source_path=self.manifest_properties.rootPath,
@@ -95,31 +102,10 @@ class WorkerManifestProperties:
         Note:
             This follows the same pattern used in asset_sync.py for manifest name generation.
         """
+        # TODO - add doc and/or define this formally as a data contract in job attachment
         hash_alg = AssetManifest.get_default_hash_alg()
         return hash_data(
             f"{self.file_system_location_name or ''}{self.root_path}".encode(), hash_alg
-        )
-
-    @classmethod
-    def from_manifest_properties(
-        cls,
-        manifest_properties: ManifestProperties,
-        local_root_path: str,
-        local_manifest_paths: Optional[List[str]] = None,
-    ) -> "WorkerManifestProperties":
-        """
-        Create WorkerManifestProperties from ManifestProperties and local paths.
-        Args:
-            manifest_properties: The original manifest properties
-            local_root_path: Local root path for attachment files
-            local_manifest_paths: Optional list of local paths for manifest files (supports step dependencies)
-        Returns:
-            WorkerManifestProperties: A new worker manifest properties instance
-        """
-        return cls(
-            manifest_properties=manifest_properties,
-            local_manifest_paths=local_manifest_paths or [],
-            local_root_path=local_root_path,
         )
 
     def to_dict(self) -> Dict[str, Any]:
--- a/src/deadline_worker_agent/sessions/attachment_models.py
+++ b/src/deadline_worker_agent/sessions/attachment_models.py
@@ -9,7 +9,6 @@ while maintaining backward compatibility for CLI operations.
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field
 from typing import Dict, List, Optional, Any
 
 from deadline.job_attachments.models import ManifestProperties, PathMappingRule
@@ -17,22 +16,12 @@ from deadline.job_attachments.asset_manifests import hash_data
 from deadline.job_attachments.asset_manifests.v2023_03_03.asset_manifest import AssetManifest
 
 
-@dataclass
 class WorkerManifestProperties:
     """
     Worker-specific manifest properties that extend ManifestProperties with local paths.
     This class contains the original manifest properties along with worker-specific
     information about local file system paths for manifest files and root directories.
     """
-
-    manifest_properties: ManifestProperties
-    """The original manifest properties from job attachment details"""
-
-    local_root_path: str
-    """Local file system root path where attachment files are stored"""
-
-    local_manifest_paths: list[str] = field(default_factory=list)
-    """Local file system paths where the manifest files are stored (supports step dependencies)"""
 
     def __init__(
         self,
@@ -150,4 +139,35 @@ class WorkerManifestProperties:
             manifest_properties=manifest_properties,
             local_manifest_paths=data.get("localManifestPaths", []),
             local_root_path=data["localRootPath"],
+        )
+
+    def __eq__(self, other: object) -> bool:
+        """
+        Check equality with another WorkerManifestProperties instance.
+        Args:
+            other: Object to compare with
+        Returns:
+            bool: True if objects are equal, False otherwise
+        """
+        if not isinstance(other, WorkerManifestProperties):
+            return False
+
+        return (
+            self.manifest_properties == other.manifest_properties
+            and self.local_root_path == other.local_root_path
+            and self.local_manifest_paths == other.local_manifest_paths
+        )
+
+    def __hash__(self) -> int:
+        """
+        Generate hash for WorkerManifestProperties instance.
+        Returns:
+            int: Hash value based on immutable attributes
+        """
+        return hash(
+            (
+                self.manifest_properties,
+                self.local_root_path,
+                tuple(self.local_manifest_paths),  # Convert list to tuple for hashing
+            )
         )
--- a/src/deadline_worker_agent/sessions/attachment_models.py
+++ b/src/deadline_worker_agent/sessions/attachment_models.py
@@ -1,0 +1,181 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Worker-specific data structures for job attachment operations.
+This module contains data structures and utilities specifically designed for
+worker agent operations, providing an alternative to hash-based correlation
+while maintaining backward compatibility for CLI operations.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Any
+
+from deadline.job_attachments.models import ManifestProperties, PathMappingRule
+from deadline.job_attachments.asset_manifests import hash_data
+from deadline.job_attachments.asset_manifests.v2023_03_03.asset_manifest import AssetManifest
+
+
+@dataclass
+class WorkerManifestProperties:
+    """
+    Worker-specific manifest properties that extend ManifestProperties with local paths.
+    This class contains the original manifest properties along with worker-specific
+    information about local file system paths for manifest files and root directories.
+    """
+
+    manifest_properties: ManifestProperties
+    """The original manifest properties from job attachment details"""
+
+    local_root_path: str
+    """Local file system root path where attachment files are stored"""
+
+    local_manifest_paths: list[str] = field(default_factory=list)
+    """Local file system paths where the manifest files are stored (supports step dependencies)"""
+
+    def __post_init__(self):
+        """Validate that required fields are set after initialization."""
+        if not self.local_root_path:
+            raise ValueError("local_root_path must be set for WorkerManifestProperties")
+
+    @property
+    def root_path(self) -> str:
+        """Get the original root path from manifest properties."""
+        return self.manifest_properties.rootPath
+
+    @property
+    def root_path_format(self):
+        """Get the root path format from manifest properties."""
+        return self.manifest_properties.rootPathFormat
+
+    @property
+    def file_system_location_name(self) -> Optional[str]:
+        """Get the file system location name from manifest properties."""
+        return self.manifest_properties.fileSystemLocationName
+
+    @property
+    def input_manifest_path(self) -> Optional[str]:
+        """Get the input manifest path from manifest properties."""
+        return self.manifest_properties.inputManifestPath
+
+    @property
+    def input_manifest_hash(self) -> Optional[str]:
+        """Get the input manifest hash from manifest properties."""
+        return self.manifest_properties.inputManifestHash
+
+    @property
+    def output_relative_directories(self) -> Optional[List[str]]:
+        """Get the output relative directories from manifest properties."""
+        return self.manifest_properties.outputRelativeDirectories
+
+    def to_path_mapping_rule(self) -> PathMappingRule:
+        """
+        Convert to an OpenJD-compatible path mapping rule.
+        Returns:
+            PathMappingRule: A path mapping rule using local root path as destination
+        Raises:
+            ValueError: If local_root_path is empty
+        """
+        if not self.local_root_path:
+            raise ValueError("local_root_path must be set to create path mapping rule")
+
+        return PathMappingRule(
+            source_path_format=self.manifest_properties.rootPathFormat.value,
+            source_path=self.manifest_properties.rootPath,
+            destination_path=self.local_root_path,
+        )
+
+    def get_hashed_source_path(self) -> str:
+        """
+        Generate a hashed string from file system location name and root path.
+        Returns:
+            str: A hash string generated from the combination of file_system_location_name
+                 and root_path using the manifest's default hash algorithm
+        Note:
+            This follows the same pattern used in asset_sync.py for manifest name generation.
+        """
+        hash_alg = AssetManifest.get_default_hash_alg()
+        return hash_data(
+            f"{self.file_system_location_name or ''}{self.root_path}".encode(), hash_alg
+        )
+
+    @classmethod
+    def from_manifest_properties(
+        cls,
+        manifest_properties: ManifestProperties,
+        local_root_path: str,
+        local_manifest_paths: Optional[List[str]] = None,
+    ) -> "WorkerManifestProperties":
+        """
+        Create WorkerManifestProperties from ManifestProperties and local paths.
+        Args:
+            manifest_properties: The original manifest properties
+            local_root_path: Local root path for attachment files
+            local_manifest_paths: Optional list of local paths for manifest files (supports step dependencies)
+        Returns:
+            WorkerManifestProperties: A new worker manifest properties instance
+        """
+        return cls(
+            manifest_properties=manifest_properties,
+            local_manifest_paths=local_manifest_paths or [],
+            local_root_path=local_root_path,
+        )
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Convert WorkerManifestProperties to a dictionary for JSON serialization.
+        Returns:
+            Dict[str, Any]: A dictionary representation of the worker manifest properties
+        """
+        # Manually serialize ManifestProperties to avoid dependency on unreleased to_dict method
+        manifest_dict: Dict[str, Any] = {"rootPath": self.manifest_properties.rootPath}
+        manifest_dict["rootPathFormat"] = self.manifest_properties.rootPathFormat.value
+
+        if self.manifest_properties.fileSystemLocationName:
+            manifest_dict["fileSystemLocationName"] = (
+                self.manifest_properties.fileSystemLocationName
+            )
+        if self.manifest_properties.inputManifestPath:
+            manifest_dict["inputManifestPath"] = self.manifest_properties.inputManifestPath
+        if self.manifest_properties.inputManifestHash:
+            manifest_dict["inputManifestHash"] = self.manifest_properties.inputManifestHash
+        if self.manifest_properties.outputRelativeDirectories:
+            manifest_dict["outputRelativeDirectories"] = self.manifest_properties.outputRelativeDirectories
+
+        return {
+            "manifestProperties": manifest_dict,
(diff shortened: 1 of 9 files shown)


src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        return {
+            "manifest_properties": {
+                "root_path": self.manifest_properties.rootPath,
+                "root_path_format": self.manifest_properties.rootPathFormat.value,
+                "file_system_location_name": self.manifest_properties.fileSystemLocationName,
+                "input_manifest_path": self.manifest_properties.inputManifestPath,
+                "input_manifest_hash": self.manifest_properties.inputManifestHash,
+                "output_relative_directories": self.manifest_properties.outputRelativeDirectories,
+            },
```
mwiebe: The `ManifestProperties` class has a `to_dict` member (see https://github.com/aws-deadline/deadline-cloud/blob/67869c6ba09fce023fd37d0ee5e9981f1f4b83eb/src/deadline/job_attachments/models.py#L206). This could call that member to shorten the code and remove the duplication.

In `ManifestProperties.to_dict`, the dict/json form of these objects is using camel case instead of snake case, it would be nice to follow that convention in this `to_dict` as well.

```suggestion
            "manifestProperties": self.manifest_properties.to_dict(),
```

src/deadline_worker_agent/sessions/session.py
```diff
@@ -223,6 +224,7 @@ def __init__(
         self._executor = ThreadPoolExecutor(max_workers=1)
         self._manifest_paths_by_root = dict()
         self._manifest_out_rel_dirs_by_source: dict[str, list[str]] = dict()
+        self._worker_manifest_properties_dict: dict[str, WorkerManifestProperties] = {}
```
mwiebe: From this line, I can't tell what the key values of this dict represent. Below, I can see it using the local root path for the key. Following how the other properties here are named, `_worker_manifest_properties_by_local_root` would be a good name that communicates what the keys are and fits in well.

test/unit/sessions/test_session.py
```diff
@@ -2698,3 +2701,326 @@ def test_session_includes_redacted_env_vars_extension(
...
+        session.set_worker_manifest_properties(worker_props)
+        session.add_local_manifest_path(worker_props.local_root_path, "/first/manifest.json")
+        session.add_local_manifest_path(worker_props.local_root_path, "/second/manifest.json")
+
+        # THEN
+        result = session.get_worker_manifest_properties(worker_props.local_root_path)
+        assert result is not None
+        assert len(result.local_manifest_paths) == 2
+        assert "/first/manifest.json" in result.local_manifest_paths
+        assert "/second/manifest.json" in result.local_manifest_paths
```
mwiebe: Here's a way you can express these three lines, that gives helpful error messages in the logs when you get a test failure. By sorting the list, it normalizes it for comparison. By comparing the list to an expected value instead of checking the length, pytest will show you the difference between the two lists with a helpful diff syntax.

```suggestion
        assert sorted(result.local_manifest_paths) == ["/first/manifest.json", "/second/manifest.json"]
```

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -0,0 +1,175 @@
...
+        if not self.local_root_path:
+            raise ValueError("local_root_path must be set to create path mapping rule")
+
+        return PathMappingRule(
+            source_path_format=self.manifest_properties.rootPathFormat.value,
+            source_path=self.manifest_properties.rootPath,
+            destination_path=self.local_root_path,
+        )
+
+    def get_hashed_source_path(self) -> str:
```
leongdl, line 85: Will we refactor the hash data string formatting so this is clear who owns it and the set standard?

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -0,0 +1,175 @@
...
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Any
+
+from deadline.job_attachments.models import ManifestProperties, PathMappingRule
+from deadline.job_attachments.asset_manifests import hash_data
+from deadline.job_attachments.asset_manifests.v2023_03_03.asset_manifest import AssetManifest
+
+
+@dataclass
+class WorkerManifestProperties:
```
leongdl, line 19: Nit - Would WorkerSessionActionManfiestProperties make sense, or WorkerSessionManifestProperty?

Since this data structure is closely tied to how a Worker's Session works.

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -0,0 +1,175 @@
...
+        Note:
+            This follows the same pattern used in asset_sync.py for manifest name generation.
+        """
+        hash_alg = AssetManifest.get_default_hash_alg()
+        return hash_data(
+            f"{self.file_system_location_name or ''}{self.root_path}".encode(), hash_alg
+        )
+
+    @classmethod
+    def from_manifest_properties(
```
leongdl: Can you link where we will integrate this? I want to see how it will be used.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: Thanks for the feedback. Let me explain my reasoning.

Regarding the naming convention, `ManifestProperties` currently uses camelCase because it's used for parsing API responses. Since there's no specific requirement for camelCase in this context, I opted for snake_case to maintain Python conventions.

I agree with your suggestion about using ManifestProperties.to_dict. The maintainability and consistency with the model outweigh the Python convention benefits. Future API changes would only need updates in one place, which reduces risk of de-synchronization between the model and its usage

I will update to use this approach.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: Yes, I agree this should be part of Job Attachment library. I've been looking for an appropriate place to define this contract, ideally within a data model.

The challenge lies in the mixed sources of the data: the hash algorithm comes from the manifest model, while 'file_system_location_name' and 'root_path' are job-level properties. For now, I've placed it here as a temporary solution.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: The collection of this is kept at session for state management, and written do disk for attachment_upload.py and attachment_download.py. `attachment_download` constructs the data, and `attachment_update` modifies and consumes it.

Here is also prototyping code if you'd like to read for more detail.
https://github.com/aws-deadline/deadline-cloud-worker-agent/compare/mainline...godobyte:deadline-cloud-worker-agent:wa-ja-prototype

test/unit/sessions/test_session.py
godobyte: Thanks for the suggestion. I'll update the test to use this pattern, it'll definitely make troubleshooting easier when tests fail.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: I chose `WorkerManifestProperties` because it clearly describes its purpose: properties related to manifests in the worker context. Since we don't currently have other `Worker*ManifestProperties` classes, adding more specificity (like Session) might be redundant.

However, if the team feels adding Session-related concepts would make the naming more clear, I'm open to that change.

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @property
+    def input_manifest_hash(self) -> Optional[str]:
+        """Get the input manifest hash from manifest properties."""
+        return self.manifest_properties.inputManifestHash
+
+    @property
+    def output_relative_directories(self) -> Optional[List[str]]:
+        """Get the output relative directories from manifest properties."""
+        return self.manifest_properties.outputRelativeDirectories
```
jericht, line 71: Could we inherit from `ManifestProperties` instead of defining these properties for all of the underlying properties? Alternatively, we could just let callers reach into the underlying `manifest_properties` themselves if we want to keep this structure.

src/deadline_worker_agent/sessions/attachment_models.py
jericht: Do we need this function? Couldn't callers just instantiate this dataclass directly?

src/deadline_worker_agent/sessions/attachment_models.py
leongdl: ok!

src/deadline_worker_agent/sessions/session.py
```diff
@@ -299,6 +301,74 @@ def add_manifest_out_rel_dirs(self, source: str, out_rel_dirs: list[str]):
...
+        ----------
+        local_root_path : str
+            The local root path key to look up
+
+        Returns
+        -------
+        WorkerManifestProperties | None
+            The worker manifest properties for the given key, or None if not found
+        """
+        return self._worker_manifest_properties_dict.get(local_root_path)
```
jericht: Do we need these two functions? They are just wrapping around the `worker_manifest_properties_dict` object.

I can see the `set_worker_manifest_properties` being a little convenient because it pulls the `local_root_path` out of the manifest properties to key into the dict for us, but not sure it's worth a dedicated function.

src/deadline_worker_agent/sessions/attachment_models.py
leongdl: ok - this should be in the comment then. Otherwise this code will get recycled and then impossible to remove to a proper interface.

src/deadline_worker_agent/sessions/session.py
```diff
@@ -299,6 +301,74 @@ def add_manifest_out_rel_dirs(self, source: str, out_rel_dirs: list[str]):
...
+
+    def get_worker_manifest_properties_list(self) -> list[WorkerManifestProperties]:
+        """Get the cached worker manifest properties as a list
+
+        Returns
+        -------
+        list[WorkerManifestProperties]
+            The cached list of worker manifest properties
+        """
+        return list(self._worker_manifest_properties_dict.values())
```
jericht: Do we need this? Also just a wrapper around the `worker_manfiest_properties_dict`

src/deadline_worker_agent/sessions/attachment_models.py
leongdl: Where is here?

src/deadline_worker_agent/sessions/session.py
```diff
@@ -299,6 +301,74 @@ def add_manifest_out_rel_dirs(self, source: str, out_rel_dirs: list[str]):
...
+        ------
+        ValueError
+            If the local_root_path is not found in the worker manifest properties dictionary
+        """
+        if worker_props := self._worker_manifest_properties_dict.get(local_root_path):
+            worker_props.local_manifest_paths.append(manifest_path)
+        else:
+            raise ValueError(
+                f"Worker manifest properties not found for local_root_path: {local_root_path}"
+            )
```
jericht: This also looks like a wrapper around the `_worker_manifest_properties_dict` as well. Do we need this?

e.g. a caller could just write this themselves:

```py
manifest_props = ... # a WorkerManifestProperties object
session = ... # a Session object
if manifest.local_root_path not in session.worker_manifest_properties_dict:
    # handle it.. e.g. skip the operation or raise an error
else:
    session.worker_manifest_properties_dict[manifest.local_root_path].append(manifest_path)
```

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: Sorry missed the linked, added to the comment above to indicate the usage. 

> Do we need this function? Couldn't callers just instantiate this dataclass directly?

Yes we'll parse from the API response directly, which contains ManifestProperties.

src/deadline_worker_agent/sessions/session.py
godobyte: Yes it'll be called in attachment download action, and it's not recommended to directly access a private member `self._worker_manifest_properties_dict`.

src/deadline_worker_agent/sessions/session.py
godobyte: Yes same idea as above for `get_worker_manifest_properties`.

src/deadline_worker_agent/sessions/session.py
godobyte: Same idea as above, we don't want to directly manipulate `session._worker_manifest_properties_dict` data via . operations and direct assignment.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: `ManifestProperties` class is defined with camelCase since it parses API result. The current approach is to follow python convention and avoid potential serialization issue. The property forwarding is to make it more clear and explicit.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: Oh just saw the comment, I'll add it in a follow-up PR if I don't get to another rev on this PR.

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -0,0 +1,183 @@
...
+                self.manifest_properties.fileSystemLocationName
+            )
+        if self.manifest_properties.inputManifestPath:
+            manifest_dict["inputManifestPath"] = self.manifest_properties.inputManifestPath
+        if self.manifest_properties.inputManifestHash:
+            manifest_dict["inputManifestHash"] = self.manifest_properties.inputManifestHash
+        if self.manifest_properties.outputRelativeDirectories:
+            manifest_dict["outputRelativeDirectories"] = (
+                self.manifest_properties.outputRelativeDirectories
+            )
```
mwiebe: The `to_dict` method is part of the public interface of the deadline-cloud library, it would be nice to use that directly.

See https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/models.py#L206 to note that the `ManifestProperties` dataclass is available with the name `deadline.job_attachments.models.ManifestProperties`, and `to_dict` is a public method: there are no names prefixed with `_` in that chain.

See also https://github.com/aws-deadline/deadline-cloud/blob/mainline/DEVELOPMENT.md#public-contracts where we document:

> All functions and modules whose name does not begin with an underscore are part of the public contract for this package.

godobyte: Ah you're right. That was my plan. I mixed that with `from_dict` for some reason during the implementation, which is not in the model. Updated to use `to_dict` from `ManifestProperties`.

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -0,0 +1,167 @@
...
+        from deadline.job_attachments.models import PathFormat
+
+        manifest_properties = ManifestProperties(
+            rootPath=manifest_props_data["rootPath"],
+            rootPathFormat=PathFormat(manifest_props_data["rootPathFormat"]),
+            fileSystemLocationName=manifest_props_data.get("fileSystemLocationName"),
+            inputManifestPath=manifest_props_data.get("inputManifestPath"),
+            inputManifestHash=manifest_props_data.get("inputManifestHash"),
+            outputRelativeDirectories=manifest_props_data.get("outputRelativeDirectories"),
+        )
```
godobyte, line 136: As per agreed, I'll add this `from_dict` to the `ManifestProperties` class and referencing it from `WorkerManifestProperties` in follow-up PRs.

src/deadline_worker_agent/sessions/attachment_models.py
mwiebe: I suspect the key will be that a document will own this, not code. I think we'll want to put that in https://github.com/aws-deadline/deadline-cloud/tree/mainline/docs

mwiebe (dismissed):

src/deadline_worker_agent/sessions/attachment_models.py
jusiskin: +1 to what @jericht said. I think we can just use:

```py
WorkerManifestProperties(
  manifest_properties=...,
  local_manifest_paths=...,
  local_root_path=...
)
```

I'm not sure I see the value of having this `WorkerManifestProperties.from_manifest_properties(...)` class function. It is longer to type.

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,518 @@
...
+        assert rule.destination_path == "C:\\local\\root"
+
+    def test_initialization_with_empty_local_root_path_raises_error(self):
+        """Test that empty string local_root_path raises ValueError."""
+        manifest_props = ManifestProperties(
+            rootPath="/source/path", rootPathFormat=PathFormat.POSIX
+        )
+
+        with pytest.raises(ValueError, match="local_root_path must be set"):
+            WorkerManifestProperties(manifest_properties=manifest_props, local_root_path="")
```
jusiskin: nit: this is a duplicate test of lines 40-47

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,518 @@
...
+            manifest_properties=manifest_props1, local_root_path="/local/root1"
+        )
+        worker_props2 = WorkerManifestProperties(
+            manifest_properties=manifest_props2, local_root_path="/local/root2"
+        )
+
+        # Verify each instance has its own empty list
+        assert worker_props1.local_manifest_paths == []
+        assert worker_props2.local_manifest_paths == []
+        assert worker_props1.local_manifest_paths is not worker_props2.local_manifest_paths
```
jusiskin, line 292: This is a good test to have. It prevents a common Python pitfall where the default is a reference to a single list

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,518 @@
...
+        worker_props = WorkerManifestProperties(
+            manifest_properties=manifest_props,
+            local_root_path="/local/root",
+            local_manifest_paths=provided_paths,
+        )
+
+        # Verify the provided list is used
+        assert worker_props.local_manifest_paths == provided_paths
+        # Verify it's the same reference (not a copy)
+        assert worker_props.local_manifest_paths is provided_paths
```
jusiskin: Is this the behaviour we want?

This means that both `WorkerManifestProperties` and the caller can modify the list and affect one another. Usually copy semantics are preferred so that `WorkerManifestProperties` maintains its own separately controlled list and is unaffected by changes made by the caller to the list supplied when creating the instance.

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,518 @@
...
+        assert "manifest" not in result_dict
+
+        # Verify manifestProperties content
+        manifest_data = result_dict["manifestProperties"]
+        assert manifest_data["rootPath"] == "/source/path"
+        assert manifest_data["rootPathFormat"] == "posix"
+        assert manifest_data["fileSystemLocationName"] == "shared_storage"
+        assert manifest_data["inputManifestPath"] == "input.json"
+        assert manifest_data["inputManifestHash"] == "hash123"
+        assert manifest_data["outputRelativeDirectories"] == ["out1", "out2"]
```
jusiskin: nit: we are now testing the behavior of `ManifestProperties.to_dict()` here.

Do we want to guard against changes to the `ManifestProperties` from `deadline-cloud`? Consider the case where a new field is added to it. The test would currently pass, but not have coverage for the new fields. Is that a problem?

Is it better to use a mock here if we defer to the behaviour of `ManifestProperties.to_dict(...)`?

test/unit/sessions/test_attachment_models.py
jusiskin, line 1: nit: the tests in this file do not follow the `GIVEN` / `WHEN` / `THEN` convention of other test modules

jusiskin (changes requested):
Great job with the code here and comprehensive tests. Just a few minor suggestions for improvement.

test/unit/sessions/test_attachment_models.py
godobyte: Good catch, will update.

test/unit/sessions/test_attachment_models.py
godobyte: Absolutely! I was also thinking about that and decided it could be updated later along with the model usage code. Since you also brought it up, I've updated the implementation to use copy semantics for better encapsulation.

test/unit/sessions/test_attachment_models.py
godobyte: Great point! The current test is tightly coupled to the implementation details of `ManifestProperties.to_dict()` from the deadline-cloud. I don't think change in deadline-cloud would easily break this since any change to the function is supposed to be backward compatible, but this test does expose implementation detail from the function. I'll update to use mock.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: ah I might have misread the initial comment, I can update this.

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Initialize WorkerManifestProperties.
+        Args:
+            manifest_properties: The original manifest properties
+            local_root_path: Local root path for attachment files
+            local_manifest_paths: Optional list of local paths for manifest files (supports step dependencies)
+        """
+        self.manifest_properties = manifest_properties
+        self.local_root_path = local_root_path
+        self.local_manifest_paths = list(local_manifest_paths) if local_manifest_paths else []
```
godobyte, line 41: This seems to be causing mypy error that I cannot produce locally 
```
(deadline-cloud-worker-agent) yuanbian@bcd07416ac9b ~/L/C/W/c/t/b/deadline-cloud-worker-agent (sync-data-model)> mypy src test
Success: no issues found in 176 source files
```

```
cmd [3] | mypy src test
/Users/runner/Library/Application Support/hatch/env/virtual/deadline-cloud-worker-agent/O0kGFt_B/deadline-cloud-worker-agent/lib/python3.11/site-packages/pydantic/v1/env_settings.py:23: error: INTERNAL ERROR -- Please try using mypy master on GitHub:
https://mypy.readthedocs.io/en/stable/common_issues.html#using-a-development-mypy-build
If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues
version: 1.18.1
/Users/runner/Library/Application Support/hatch/env/virtual/deadline-cloud-worker-agent/O0kGFt_B/deadline-cloud-worker-agent/lib/python3.11/site-packages/pydantic/v1/env_settings.py:23: : note: please use --show-traceback to print a traceback when reporting a bug
Error: Process completed with exit code 2.
```

I did some research, apparently it's related to mypy and pydantic v1 and v2. I may timebox it and revert to previous implementation if it doesn't work.

godobyte: Turned out to be a breaking change introduced by mypy v1.18.1, fixing in a separate PR https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/736

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,513 @@
...
+
+        worker_props.local_manifest_paths.extend(["/manifest2.json", "/manifest3.json"])
+        assert len(worker_props.local_manifest_paths) == 3
+        assert "/manifest2.json" in worker_props.local_manifest_paths
+        assert "/manifest3.json" in worker_props.local_manifest_paths
+
+        # Test insertion
+        worker_props.local_manifest_paths.insert(0, "/manifest0.json")
+        assert worker_props.local_manifest_paths[0] == "/manifest0.json"
+        assert len(worker_props.local_manifest_paths) == 4
```
jusiskin: nit: this one is still missing the `GIVEN` / `WHEN` / `THEN`.

Looks like a few more tests below too.

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,513 @@
...
+        assert manifest_data["rootPath"] == "/source/path"
+        assert manifest_data["rootPathFormat"] == "windows"
+        # None values are not included in the dictionary by ManifestProperties.to_dict()
+        assert "fileSystemLocationName" not in manifest_data
+        assert "inputManifestPath" not in manifest_data
+        assert "inputManifestHash" not in manifest_data
+        assert "outputRelativeDirectories" not in manifest_data
+
+        assert result_dict["localManifestPaths"] == []
+        assert result_dict["localRootPath"] == "/local/root"
```
jusiskin: nit: Same here. This one is still missing the `GIVEN` / `WHEN` / `THEN`

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,513 @@
...
+        assert worker_props.local_root_path == "/local/root"
+
+    def test_from_dict_missing_required_fields_raises_error(self):
+        """Test that from_dict raises KeyError for missing required fields."""
+        # GIVEN/WHEN/THEN - Missing manifestProperties
+        data_missing_manifest = {
+            "localRootPath": "/local/root",
+        }
+        with pytest.raises(KeyError):
+            WorkerManifestProperties.from_dict(data_missing_manifest)
```
jusiskin: a little unusual in the way pytest does this, but how about this:


```suggestion
        # GIVEN - Missing manifestProperties
        data_missing_manifest = {
            "localRootPath": "/local/root",
        }
        # THEN
        with pytest.raises(KeyError):
            # WHEN
            WorkerManifestProperties.from_dict(data_missing_manifest)
```

same applies to the tests below

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,513 @@
...
+            manifest_properties=manifest_props,
+            local_root_path="/local/minimal",
+        )
+
+        # Serialize to dict and back
+        data_dict = original_worker_props.to_dict()
+        reconstructed_worker_props = WorkerManifestProperties.from_dict(data_dict)
+
+        # Verify they are equal
+        assert reconstructed_worker_props == original_worker_props
```
jusiskin: nit: missing the `GIVEN` / `WHEN` / `THEN`

test/unit/sessions/test_attachment_models.py
```diff
@@ -0,0 +1,513 @@
...
+        data = {
+            "manifestProperties": {
+                "rootPath": "/source/path",
+                "rootPathFormat": "invalid_format",
+            },
+            "localRootPath": "/local/root",
+        }
+
+        with pytest.raises(ValueError):
+            WorkerManifestProperties.from_dict(data)
```
jusiskin: nit: missing the `GIVEN` / `WHEN` / `THEN`

test/unit/sessions/test_attachment_models.py
jusiskin: I still see coverage of the `ManifestProperties.to_dict()` in the tests and do not see any mocks added.

test/unit/sessions/test_attachment_models.py
jusiskin: It appears there are still some without this, but I don't consider these merge-blocking. I've pointed out the ones remaining as new review comments, but they are optional.

jusiskin (changes requested):
Thanks for addressing prior feedback.

The remaining comments are mostly just nitpicks about the `GIVEN` / `WHEN` / `THEN` comments in the unit tests which I don't consider critical.

The one open comment that I think we want to address is about the remaining test coverage of `ManifestProperties.to_dict()` and prior discussion to use a mock instead.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: Added a TODO below on L94 to track the update.

jusiskin (approved):

mwiebe (approved):

src/deadline_worker_agent/sessions/attachment_models.py
leongdl: I suspect the key will be that a document will own this, not code. I think we'll want to put that in https://github.com/aws-deadline/deadline-cloud/tree/mainline/docs

IMO, adding to docs is great but not everyone will read or discover it or read every version of it. Only code and tests can discourage access.

PR #749, opened 2025-09-25
chore: Add S3 output config validation to e2e tests

AlexTranAmz:
Job output, including manifest file and data files, following a specific convention that expose to customer. We need tests to verify and ensure no changes would break the backward compatibility.

Check output manifest follow:
- s3 partition follows farmid/queueid/stepid/taskid/{timestamp}sessionactionid/
- s3 file name follows {hash}_output where hash is rootPath and fileSystemLocationName when present 
- s3 metadata
   - file-system-location-name (optional when it exist)
   - asset-root and asset-root-json (when value is non-ASCII)

We have tighter test which could better caught breaking changes

Running unit tests:
```
===================================================================== slowest 5 durations =====================================================================
1.05s call     test/unit/sessions/test_session.py::TestSessionSyncAssetInputs::test_asset_loading_method[COPIED]
1.01s call     test/unit/test_worker.py::TestMonitorEc2Shutdown::test_asg_termination
1.00s call     test/unit/test_worker.py::TestMonitorEc2Shutdown::test_spot_interruption[spot-shutdown-1-min]
1.00s call     test/unit/test_worker.py::TestMonitorEc2Shutdown::test_spot_interruption[spot-shutdown-15-sec]
0.92s call     test/unit/sessions/test_session.py::TestSessionSyncAssetOutputs::test_sync_asset_outputs_without_manifest_reporting
======================================================= 2628 passed, 62 skipped, 25 warnings in 21.61s ========================================================
```

Running e2e tests:

- Linux
```
===End Flaky Test Report===
================================================================================================================================================== slowest 5 durations ==================================================================================================================================================
184.91s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_enters_stopping_state_while_draining
169.29s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_success
121.92s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux[COPIED]
111.72s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux[VIRTUAL]
91.82s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_reports_canceled_sync_input_actions_as_canceled
==================================================================================================================== 29 passed, 2 skipped, 27 deselected, 1 xpassed, 1 warning in 1699.11s (0:28:19) ====================================================================================================================
```

- Windows
```
===End Flaky Test Report===
================================================================================================================================================== slowest 5 durations ==================================================================================================================================================
377.74s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_enters_stopping_state_while_draining
375.42s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_success
130.87s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_reports_canceled_sync_input_actions_as_canceled
121.93s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_windows
88.98s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_reports_task_progress_and_status_message
========================================================================================================================= 25 passed, 7 skipped, 27 deselected, 1 warning in 1864.85s (0:31:04) ==========================================================================================================================
```

Yes

No

----

### DIFF
--- a/test/e2e/s3_validation_utils.py
+++ b/test/e2e/s3_validation_utils.py
@@ -1,0 +1,377 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Utility functions for validating S3 setup in E2E tests for Deadline Cloud jobs.
+"""
+
+import json
+import logging
+from typing import Dict, List, Optional, Tuple
+
+import boto3
+from botocore.exceptions import ClientError
+
+from deadline.job_attachments._aws.deadline import get_job, get_queue
+from deadline.job_attachments.asset_manifests.hash_algorithms import hash_data
+from deadline.job_attachments.asset_manifests.v2023_03_03.asset_manifest import DEFAULT_HASH_ALG
+from deadline.job_attachments.download import _get_tasks_manifests_keys_from_s3
+from deadline.job_attachments._aws.aws_clients import get_s3_client
+
+from deadline_test_fixtures import DeadlineClient, Job
+from datetime import datetime
+
+LOG = logging.getLogger(__name__)
+
+
+class S3ValidationError(Exception):
+    """Exception raised when S3 validation fails."""
+
+    pass
+
+
+def validate_s3_job_output_manifest(
+    job: Job,
+    deadline_client: DeadlineClient,
+    session: Optional[boto3.Session] = None,
+) -> None:
+    """
+    Validates the S3 setup for a Deadline Cloud job by checking:
+    1. Job attachment manifest and settings to get S3 bucket, rootPath, and fileSystemLocationName
+    2. Lists all steps of the job and constructs S3 keys for each step's output manifests
+    3. Verifies that output directories exist and match expected session actions
+    4. Validates that session action IDs in S3 keys correspond to actual session actions
+    5. Validates that timestamps in S3 keys match session action start times (accurate to the second)
+    6. Validates output file hashes and metadata
+
+    Args:
+        job: The Deadline Cloud job to validate
+        deadline_client: Deadline client for API calls
+        session: Optional boto3 session
+
+    Raises:
+        S3ValidationError: If S3 client operations fail
+        AssertionError: If validation checks fail
+    """
+    farm_id = job.farm.id
+    queue_id = job.queue.id
+    # Get job and queue details
+    job_details = get_job(farm_id=farm_id, queue_id=queue_id, job_id=job.id, session=session)
+    queue_details = get_queue(farm_id=farm_id, queue_id=queue_id, session=session)
+
+    assert job_details.attachments, f"Job {job.id} has no attachments"
+    assert job_details.attachments.manifests, f"Job {job.id} has no manifests"
+
+    LOG.debug(f"Job {job.id} has {len(job_details.attachments.manifests)} manifest(s)")
+
+    assert queue_details.jobAttachmentSettings, f"Queue {queue_id} has no job attachment settings"
+    LOG.debug(f"Queue {queue_id} has job attachment settings configured")
+
+    # Extract S3 configuration
+    s3_bucket = queue_details.jobAttachmentSettings.s3BucketName
+
+    for manifest_properties in job_details.attachments.manifests:
+        root_path = manifest_properties.rootPath
+        file_system_location_name = manifest_properties.fileSystemLocationName
+
+        # Validate each step's output directories for this manifest
+        steps = _get_job_steps(deadline_client, farm_id, queue_id, job.id)
+
+        for step in steps:
+            step_id = step["stepId"]
+            manifest_prefix = f"Deadline/Manifests/{farm_id}/{queue_id}/{job.id}/{step_id}"
+
+            try:
+                manifest_keys = _get_tasks_manifests_keys_from_s3(
+                    manifest_prefix=manifest_prefix, s3_bucket=s3_bucket, session=session
+                )
+
+                if manifest_keys:
+                    LOG.info(
+                        f"Validating {len(manifest_keys)} outputs for step {step_id} with manifest root_path={root_path}"
+                    )
+                    for manifest_key in manifest_keys:
+                        _validate_output_directory(
+                            manifest_key=manifest_key,
+                            s3_bucket=s3_bucket,
+                            deadline_client=deadline_client,
+                            farm_id=farm_id,
+                            queue_id=queue_id,
+                            job_id=job.id,
+                            root_path=root_path,
+                            file_system_location_name=file_system_location_name,
+                            session=session,
+                        )
+            except (ClientError, S3ValidationError, AssertionError) as e:
+                LOG.debug(f"No output manifests for step {step_id} with root_path={root_path}: {e}")
+
+
+def _get_job_steps(
+    deadline_client: DeadlineClient, farm_id: str, queue_id: str, job_id: str
+) -> List[Dict]:
+    """Get all steps for a job."""
+    try:
+        steps = []
+        next_token = None
+
+        while True:
+            params = {"farmId": farm_id, "queueId": queue_id, "jobId": job_id}
+            if next_token:
+                params["nextToken"] = next_token
+
+            response = deadline_client.list_steps(**params)
+            steps.extend(response.get("steps", []))
+
+            next_token = response.get("nextToken")
+            if not next_token:
+                break
+
+        return steps
+    except ClientError as e:
+        raise S3ValidationError(f"Failed to list steps for job {job_id}: {e}")
+
+
+def _validate_output_directory(
+    manifest_key: str,
+    s3_bucket: str,
+    deadline_client: DeadlineClient,
+    farm_id: str,
+    queue_id: str,
+    job_id: str,
+    root_path: str,
+    file_system_location_name: Optional[str],
+    session: Optional[boto3.Session] = None,
+) -> None:
+    """Validate output directory: session action, timestamp, hash, and metadata."""
+    # Parse key for step ID, task ID, session action ID and timestamp
+    parsed_step_id, task_id, session_action_id, timestamp_part = _parse_manifest_key(
+        manifest_key=manifest_key
+    )
+
+    # Validate task ID exists for the step
+    _validate_task_exists(
+        deadline_client=deadline_client,
+        farm_id=farm_id,
+        queue_id=queue_id,
+        job_id=job_id,
+        step_id=parsed_step_id,
+        task_id=task_id,
+    )
+
+    # Validate session action exists and timestamp matches
+    session_action = _get_session_action_details(
+        deadline_client=deadline_client,
+        farm_id=farm_id,
+        queue_id=queue_id,
+        job_id=job_id,
+        session_action_id=session_action_id,
+    )
+    _validate_timestamp_match(timestamp_part=timestamp_part, session_action=session_action)
+
+    # Validate file hash and S3 metadata
+    _validate_file_and_metadata(
+        manifest_key=manifest_key,
+        s3_bucket=s3_bucket,
+        root_path=root_path,
+        file_system_location_name=file_system_location_name,
+        session=session,
+    )
+
+
+def _get_session_action_details(
+    deadline_client: DeadlineClient,
+    farm_id: str,
+    queue_id: str,
+    job_id: str,
+    session_action_id: str,
+) -> Dict:
+    """Get session action details by searching through all sessions."""
+    try:
+        sessions_response = deadline_client.list_sessions(
+            farmId=farm_id, queueId=queue_id, jobId=job_id
+        )
+
+        sessions = sessions_response.get("sessions", [])
+        assert sessions, f"No sessions found for job {job_id}"
+        LOG.debug(f"Found {len(sessions)} session(s) for job {job_id}")
+
+        # Search for the session action across all sessions
+        for session in sessions:
+            try:
+                next_token = None
+
+                while True:
+                    params = {
+                        "farmId": farm_id,
+                        "queueId": queue_id,
+                        "jobId": job_id,
+                        "sessionId": session["sessionId"],
+                    }
+                    if next_token:
+                        params["nextToken"] = next_token
+
+                    actions_response = deadline_client.list_session_actions(**params)
+
+                    for action in actions_response.get("sessionActions", []):
+                        if action["sessionActionId"] == session_action_id:
+                            LOG.debug(f"Found session action {session_action_id}")
+                            return action
+
+                    next_token = actions_response.get("nextToken")
+                    if not next_token:
+                        break
+
+            except ClientError:
+                continue
+
+        # This is a validation failure, not a client error
+        LOG.error(f"Session action {session_action_id} not found for job {job_id}")
+        assert False, f"Session action {session_action_id} not found for job {job_id}"
+
+    except ClientError as e:
+        raise S3ValidationError(f"Failed to list sessions for job {job_id}: {e}")
+
+
+def _parse_manifest_key(manifest_key: str) -> Tuple[str, str, str, str]:
+    """Parse manifest key to extract step ID, task ID, session action ID and timestamp."""
+    key_parts = manifest_key.split("/")
+    step_id = task_id = session_action_id = timestamp_part = None
+
+    for part in key_parts:
+        if part.startswith("step-"):
+            step_id = part
+        elif part.startswith("task-"):
+            task_id = part
+        elif "sessionaction-" in part:
+            if "_" in part:
+                # This is because the file name currently follow the format of {timestamp}_{session_action_id}
+                timestamp_part, session_action_id = part.split("_", 1)
+            else:
+                session_action_id = part
+
+    assert step_id, f"Could not parse step ID from key: {manifest_key}"
+    LOG.debug(f"Parsed step ID: {step_id}")
+
+    assert task_id, f"Could not parse task ID from key: {manifest_key}"
+    LOG.debug(f"Parsed task ID: {task_id}")
+
+    assert session_action_id, f"Could not parse session action ID from key: {manifest_key}"
+    LOG.debug(f"Parsed session action ID: {session_action_id}")
+
+    assert timestamp_part, f"Could not parse timestamp from key: {manifest_key}"
+    LOG.debug(f"Parsed timestamp: {timestamp_part}")
+
+    return step_id, task_id, session_action_id, timestamp_part
+
+
+def _validate_task_exists(
+    deadline_client: DeadlineClient,
+    farm_id: str,
+    queue_id: str,
+    job_id: str,
+    step_id: str,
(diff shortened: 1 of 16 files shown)


test/e2e/utils.py
```diff
@@ -379,3 +381,34 @@ def get_shutdown_on_stop_status_from_toml(
         return cmd_result.stdout.strip()
     else:
         raise Exception(f"Unsupported operating system: {os.environ['OPERATING_SYSTEM']}")
+
+
+def assert_job_s3_setup_valid(
```
godobyte: This appears to be unused code. Could you help me understand its purpose?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+    """Exception raised when S3 validation fails."""
+
+    pass
+
+
+def validate_job_s3_setup(
+    job: Job,
+    deadline_client: DeadlineClient,
+    farm_id: str,
+    queue_id: str,
```
godobyte: I saw this `Job` from `deadline-cloud-test-fixtures/src/deadline_test_fixtures/deadline/resources.py` contains farm and queue, do they need to be passed in again?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+    manifest_properties = job_details.attachments.manifests[0]
+    root_path = manifest_properties.rootPath
+    file_system_location_name = manifest_properties.fileSystemLocationName
+
+    # Validate each step's output directories
+    steps = _get_job_steps(deadline_client, farm_id, queue_id, job.id)
+
+    for step in steps:
+        step_id = step["stepId"]
+        manifest_prefix = f"DeadlineCloud/Manifests/{farm_id}/{queue_id}/{job.id}/{step_id}"
```
godobyte: Could we reuse existing function?

```
 s3_settings.full_output_prefix()
```

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+
+    if not job_details.attachments or not job_details.attachments.manifests:
+        raise S3ValidationError(f"Job {job.id} has no attachments or manifests")
+
+    if not queue_details.jobAttachmentSettings:
+        raise S3ValidationError(f"Queue {queue_id} has no job attachment settings")
+
+    # Extract S3 configuration
+    s3_bucket = queue_details.jobAttachmentSettings.s3BucketName
+    manifest_properties = job_details.attachments.manifests[0]
```
godobyte: `job_details.attachments.manifests` is a list of manifest properties, we need to validate all of them.

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+    step_id = task_id = session_action_id = timestamp_part = None
+
+    for part in key_parts:
+        if part.startswith("step-"):
+            step_id = part
+        elif part.startswith("task-"):
+            task_id = part
+        elif "sessionaction-" in part:
+            if "_" in part:
+                timestamp_part, session_action_id = part.split("_", 1)
```
godobyte, line 247: This could use some comment, an example could help to understand.

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+    # Validate hash
+    file_hash = filename.replace("_output", "")
+    hash_input = f"{file_system_location_name or ''}{root_path}".encode()
+    expected_hash = hash_data(hash_input, DEFAULT_HASH_ALG)
+
+    if file_hash != expected_hash:
+        raise S3ValidationError(f"Hash mismatch: expected {expected_hash}, got {file_hash}")
+
+    # Validate S3 metadata
+    s3_client = boto3.client("s3") if session is None else session.client("s3")
```
godobyte: Could we reuse `get_s3_client` from deadline?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+
+    try:
+        response = s3_client.head_object(Bucket=s3_bucket, Key=manifest_key)
```
godobyte, line 352: Great, this is an API job attachment download needs too, job runners should have access already.

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+        metadata = response.get("Metadata", {})
+
+        # Validate required metadata fields
+        if metadata.get("asset-root") != root_path:
+            raise S3ValidationError(f"Asset root mismatch: expected {root_path}")
```
godobyte: `asset-root-json` is to store the non-ascii root_path as json. This would fail if the root_path is saved as `asset-root-json`, which is to ve validated in L298.

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+
+        if task_id not in task_ids:
+            raise S3ValidationError(f"Task {task_id} not found in step {step_id}")
+
+    except ClientError as e:
+        raise S3ValidationError(f"Failed to list tasks for step {step_id}: {e}")
+
+
+def _validate_timestamp_match(timestamp_part: str, session_action: Dict) -> None:
+    """Validate that key timestamp matches session action start time."""
```
godobyte, line 301: Good catch! The new code path actually does not comply with it, I'd check with stakeholders. Let's keep this validation here for now.

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class S3ValidationError(Exception):
+    """Exception raised when S3 validation fails."""
+
+    pass
+
+
+def validate_job_s3_setup(
```
godobyte: The name can be a bit misleading, let's be explicit. What do you think about `validate_s3_job_output_manifest`?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+
+    # Validate session action exists and timestamp matches
+    session_action = _get_session_action_details(
+        deadline_client, farm_id, queue_id, job_id, session_action_id
+    )
+    _validate_timestamp_match(timestamp_part, session_action)
+
+    # Validate file hash and S3 metadata
+    _validate_file_and_metadata(
+        manifest_key, s3_bucket, root_path, file_system_location_name, session
```
godobyte: nit - do they have to be in helpers?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+    queue_id: str,
+    job_id: str,
+    step_id: str,
+    task_id: str,
+) -> None:
+    """Validate that the task ID exists for the given step."""
+    try:
+        response = deadline_client.list_tasks(
+            farmId=farm_id, queueId=queue_id, jobId=job_id, stepId=step_id
+        )
```
godobyte: We can have many tasks per step, the API is paginated. I don't think we currently have jobs with tasks beyond a page, but we should at least check and fail when that's the case. It reminds us to implement pagination when needed.

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+        except Exception as e:
+            LOG.debug(f"No output manifests for step {step_id}: {e}")
+
+
+def _get_job_steps(
+    deadline_client: DeadlineClient, farm_id: str, queue_id: str, job_id: str
+) -> List[Dict]:
+    """Get all steps for a job."""
+    try:
+        response = deadline_client.list_steps(farmId=farm_id, queueId=queue_id, jobId=job_id)
```
godobyte: We can have many steps per task, the API is paginated. I don't think we currently have jobs with steps beyond a page, but we should at least make sure by checking `nextToken`. It reminds us to implement pagination when needed.

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+def _get_session_action_details(
+    deadline_client: DeadlineClient,
+    farm_id: str,
+    queue_id: str,
+    job_id: str,
+    session_action_id: str,
+) -> Dict:
+    """Get session action details by searching through all sessions."""
+    try:
+        sessions_response = deadline_client.list_sessions(
```
godobyte, line 189: Same for pagination and `nextToken`.

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,306 @@
...
+            )
+        elif start_time is not None:
+            action_start_time = start_time.replace(microsecond=0)
+        else:
+            raise S3ValidationError("Session action has no startedAt timestamp")
+
+        if key_timestamp != action_start_time:
+            raise S3ValidationError(
+                f"Timestamp mismatch: key={key_timestamp}, action start={action_start_time}"
+            )
```
godobyte: Could we use assertion and logs rather than raise the ValidationError when applicable? I noticed this file mixes error (like s3 client error) and value checking by using the same `S3ValidationError`. The later would be cleaner with assert and logs when it fails.

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: Not possible in this case as we just want until `step_id`, but the function require `session_action_id` as well

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: I believe they do, since they just aid the main assert function.

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: Got it

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/s3_validation_utils.py
AlexTranAmz: Yes

test/e2e/s3_validation_utils.py
AlexTranAmz: Fixed

test/e2e/utils.py
AlexTranAmz: fix

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
...
+    """
+    farm_id = job.farm.id
+    queue_id = job.queue.id
+    # Get job and queue details
+    job_details = get_job(farm_id, queue_id, job.id, session=session)
+    queue_details = get_queue(farm_id, queue_id, session=session)
+
+    assert job_details.attachments and job_details.attachments.manifests, (
+        f"Job {job.id} has no attachments or manifests"
+    )
```
AWS-Samuel: We should consider splitting these two checks in order to get more specific error messaging


```
    assert job_details.attachments,  f"Job {job.id} has no attachments"
    assert job_details.attachments.manifests, f"Job {job.id} has no manifests"
```

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
...
+                            s3_bucket,
+                            deadline_client,
+                            farm_id,
+                            queue_id,
+                            job.id,
+                            root_path,
+                            file_system_location_name,
+                            session,
+                        )
+            except Exception as e:
```
AWS-Samuel: Can we narrow down the exception here to something more specific?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
...
+        assert task_id in task_ids, f"Task {task_id} not found in step {step_id}"
+        LOG.debug(f"Task {task_id} validated in step {step_id}")
+
+    except ClientError as e:
+        raise S3ValidationError(f"Failed to list tasks for step {step_id}: {e}")
+
+
+def _validate_timestamp_match(timestamp_part: str, session_action: Dict) -> None:
+    """Validate that key timestamp matches session action start time."""
+    from datetime import datetime
```
AWS-Samuel: we should perform this import at the top of the file with the other imports

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
...
+            )
+        elif start_time is not None:
+            action_start_time = start_time.replace(microsecond=0)
+        else:
+            LOG.error("Session action has no startedAt timestamp")
+            assert False, "Session action has no startedAt timestamp"
+
+        assert key_timestamp == action_start_time, (
+            f"Timestamp mismatch: key={key_timestamp}, action start={action_start_time}"
+        )
```
AWS-Samuel, line 320: I don't know how either of these timestamps are generated, do you know if they are generated independently? (as in, the key timestamp is not generated directly using the action start time)

If they are independently created, is there a chance that they could ever be a second or two off from each other and is that something that we want to fail on or should we add some wiggle room?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
...
+def _validate_file_and_metadata(
+    manifest_key: str,
+    s3_bucket: str,
+    root_path: str,
+    file_system_location_name: Optional[str],
+    session: Optional[boto3.Session] = None,
+) -> None:
+    """Validate file hash and S3 metadata."""
+    filename = manifest_key.split("/")[-1]
+    assert filename.endswith("_output"), f"Invalid filename format: {filename}"
```
AWS-Samuel:
```suggestion
    assert filename.endswith("_output"), f"Nonvalid filename format: {filename}"
```

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
...
+            assert False, "Session action has no startedAt timestamp"
+
+        assert key_timestamp == action_start_time, (
+            f"Timestamp mismatch: key={key_timestamp}, action start={action_start_time}"
+        )
+        LOG.debug(f"Timestamp validation passed: {key_timestamp}")
+
+    except ValueError as e:
+        LOG.error(f"Invalid timestamp format: {timestamp_part}. Error: {e}")
+        assert False, f"Invalid timestamp format: {timestamp_part}. Error: {e}"
```
AWS-Samuel:
```suggestion
        LOG.error(f"Nonvalid timestamp format: {timestamp_part}. Error: {e}")
        assert False, f"Nonvalid timestamp format: {timestamp_part}. Error: {e}"
```

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
```
leongdl, line 3: Really meta comment - why are we coding up such a large chunk of code when we can check in a set of expected test result, and then do a file + folder comparison?

Mark created the job expected output snapshot command so we can test w/o listing and calling so many APIs. Is it possible to add more tests and variations that compare output contents instead of code based comparison?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
...
+        # Validate each step's output directories for this manifest
+        steps = _get_job_steps(deadline_client, farm_id, queue_id, job.id)
+
+        for step in steps:
+            step_id = step["stepId"]
+            manifest_prefix = f"DeadlineCloud/Manifests/{farm_id}/{queue_id}/{job.id}/{step_id}"
+
+            try:
+                manifest_keys = _get_tasks_manifests_keys_from_s3(
+                    manifest_prefix, s3_bucket, session=session
```
godobyte: nit - could we be consistent about using `argName=`?

test/e2e/s3_validation_utils.py
```diff
@@ -0,0 +1,362 @@
...
+            assert metadata.get("file-system-location-name") == file_system_location_name, (
+                f"File system location mismatch: expected {file_system_location_name}, got {metadata.get('file-system-location-name')}"
+            )
+            LOG.debug(f"File system location metadata validated: {file_system_location_name}")
+
+        # Validate JSON encoding for non-ASCII paths
+        if not root_path.isascii():
+            json_root = metadata.get("asset-root-json")
+            assert json_root and json.loads(json_root) == root_path, (
+                f"Asset root JSON mismatch for non-ASCII path: expected {root_path}, got {json_root}"
```
godobyte, line 370: By looking at where this is generated, `asset-root` also exits when it's not ascii, both `asset-root` and `asset-root-json` are set to the same value.

https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/job_attachments/asset_sync.py#L543-L554.

test/e2e/s3_validation_utils.py
AlexTranAmz: I've tested multiple time and all passing with the same second so I think we're good

test/e2e/s3_validation_utils.py
leongdl: Talked offline - we are OK with this test strategy in this PR. Other tests can try the folder + file comparison.

test/e2e/s3_validation_utils.py
AWS-Samuel: Synced offline and confirmed that both of these timestamps pull from the same source and so slight drift is not a concern

AWS-Samuel (approved):

leongdl (approved):
Ok with this approach - but thanks for using a file based comparison for the upcoming tests.

godobyte (approved):

PR #752, opened 2025-10-02
test: add cleanup script for e2e tests

edwards-aws:
We need to be able to clean up old E2E tests resources before running E2E tests so that we can ensure that our tests are running in a clean environment.

When tests create EC2 instances they will tag instances with the test category it'll be running as. Examples of this are LinuxMainline, WindowsMainline, LinuxRelease, WindowsRelease, etc.

I added a script that runs before the E2E tests when running `hatch run e2e-test`, the script will terminate all instances tagged with the appropriate test category (this will be in an environment variable) and cancel all unfinished jobs in the tests queues. The reason for tagging instances is so that a cleanup doesn't mess with test runs that are already going in the same account.

#### Why not a pytest fixture
Good question!

We only need to run this once at test start. All tests require the cleanup to run so we'd have to add the fixture to every test/test class as a session fixture, meaning it'd be easy for a dev to forget in the future. Additionally session fixtures get messy when parallelizing tests. The script avoids all of that. A nice side effect is that devs can also run the script without the tests if they just want to cleanup their environment without running tests.

#### Why can't we just delete instances tagged with the farm and queues of the running tests?
Another great question!

There's an edge case where the tests timeout and exit before the worker starts, so leftover instances may not have the tags that Deadline Cloud puts on the instances.

Less flaky tests.

Ran E2E tests and cancelled them early, ensuring that instances and unfinished jobs would be kept around. I then ran `hatch run e2e-test` and ensured that all appropriate resources were cleaned up and tests succeeded. I made sure to run a Windows test and Linux test at the same time cancelling both randomly and starting them again, ensuring that the only jobs and instances cleaned up were the ones that belonged to the previous test run.

N/A

No
----

NOTE: This can't be merged until https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/210 is in and released. This change depends on a change in that package.

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -11,7 +11,10 @@ test = "pytest test/unit --numprocesses=auto {args}"
 # See: https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html#output-stdout-and-stderr-from-workers
 version = "hatch version"
 metadata = "hatch project metadata {args:}"
-e2e-test= "pytest --no-cov test/e2e {args:}"
+e2e-test= [
+  "python test/e2e/pre_test_cleanup.py",
+  "pytest --no-cov test/e2e {args:}",
+]
 integ-test = "pytest --no-cov test/integ {args:}"
 typing = "mypy {args:src test}"
 style = [
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -5,6 +5,7 @@ import dataclasses
 import logging
 import os
 import pytest
+
 from dataclasses import dataclass, field, InitVar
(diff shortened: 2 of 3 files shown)


edwards-aws:
The mypy checks will probably fail until this change is released -> https://github.com/aws-deadline/deadline-cloud-test-fixtures/pull/210 I've added functionality to the test fixtures for this change.

moorec-aws (dismissed):

miabatta (dismissed):

miabatta (approved):

moorec-aws (approved):

PR #754, opened 2025-10-03
test: add multi output dir and scope down permission to step-step job bundle

AlexTranAmz:
- We want to have e2e test for job with multiple output dir
- We want to have e2e test with scope down permission to verify experimental feature

- Add 2nd `INOUT` dir to the step step dependencies bundle
- For linux job bundle, we also scope down the permission of the second dir if we detect that Asset Sync's feature flag is turn on
- Expose the Asset Sync's feature flag to the run task
- Change the test logic comparing output to account for the newly created dir
- Move the job bundle structure to reflect the 2 `INOUT` dir

Better test coverage with backward compatibility in mind

Running E2E test
- Linux
```
===End Flaky Test Report===
====================================================================================================== slowest 5 durations =======================================================================================================
230.06s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_enters_stopping_state_while_draining
214.92s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_success
140.81s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux[VIRTUAL]
131.08s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux[COPIED]
91.48s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_fails_job_attachment_sync_when_file_does_not_exist_in_bucket
==================================================================================================== short test summary info =====================================================================================================
FAILED test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only - NotImplementedError: Currently do not support more than one root paths, provided dict_keys([])
=================================================================== 1 failed, 28 passed, 2 skipped, 27 deselected, 1 xpassed, 1 warning in 1848.20s (0:30:48) ====================================================================
```
- Windows
```
===End Flaky Test Report===
====================================================================================================== slowest 5 durations =======================================================================================================
335.30s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_success
296.32s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_enters_stopping_state_while_draining
203.08s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_windows
78.44s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_uses_job_attachment_sync[$InputFolder = "{{Param.DataDir}}\\files"\n$OutputFile = "{{Param.DataDir}}\\output_file.txt"\n$combinedContent = ""\n$files = Get-ChildItem -Path $InputFolder -File\nforeach ($file in $files) {\n   $combinedContent += [IO.File]::ReadAllText($file.FullName)\n}\n$sha256 = [System.Security.Cryptography.SHA256]::Create().ComputeHash([System.Text.Encoding]::UTF8.GetBytes($combinedContent))\n$hashString = [System.BitConverter]::ToString($sha256).Replace("-", "").ToLower()\nSet-Content -Path $OutputFile -Value $hashString -NoNewLine]
78.18s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_reports_task_progress_and_status_message
==================================================================================================== short test summary info =====================================================================================================
FAILED test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only - NotImplementedError: Currently do not support more than one root paths, provided dict_keys([])
========================================================================= 1 failed, 24 passed, 7 skipped, 27 deselected, 1 warning in 1747.10s (0:29:07) =========================================================================
```

Yes

No

----


test/e2e/test_job_submissions.py
```diff
@@ -1957,10 +1958,36 @@ def test_worker_job_attachments_dep_data_flow_linux(
...
+            for item in os.listdir(data_dir1):
+                src = os.path.join(data_dir1, item)
+                dst = os.path.join(combined_dir, item)
+                if os.path.isdir(src):
+                    shutil.copytree(src, dst, dirs_exist_ok=True)
+                else:
+                    shutil.copy2(src, dst)
+
+        # Copy contents from data_dir2
+        data_dir2 = output_root_path + "/data_dir2/Test"
```
godobyte: nit - could we update the `Test` dir name to follow python dir naming convention.

test/e2e/job_attachment_bundle/dep_data_flow/linux_bundle/template.yaml
```diff
@@ -299,4 +305,292 @@ steps:
           cp create_job_in.txt $STEP.out
           echo "2. Processed in $STEP" >> $STEP.out
 
+          echo Finished $STEP!
+- name: Step2
```
leongdl: Recommendation - we want to test 2 data directories that interact in a single session, not 2 separate steps with each independently writing to separate outputs. This is needed because each step (1 task) will be run on 1 worker. By splitting the steps it can run on different workers.

godobyte: +1

test/e2e/test_job_submissions.py
```diff
@@ -2000,10 +2027,36 @@ def test_worker_job_attachments_dep_data_flow_windows(
...
+        # Copy contents from data_dir2
+        data_dir2 = output_root_path + "/data_dir2"
+        if os.path.exists(data_dir2):
+            for item in os.listdir(data_dir2):
+                src = os.path.join(data_dir2, item)
+                dst = os.path.join(combined_dir, item)
+                if os.path.isdir(src):
+                    shutil.copytree(src, dst, dirs_exist_ok=True)
+                else:
+                    shutil.copy2(src, dst)
```
godobyte: I'd check why `verify_output_dir_matches` doesn't verify sub-directories than modifying the output dir structures only to work with the function's limitation. That would increase the `verify_output_dir_matches`'s reusability in the future.

test/e2e/job_attachment_bundle/dep_data_flow/linux_bundle/template.yaml
AlexTranAmz: Fixed in new rev

test/e2e/test_job_submissions.py
AlexTranAmz: Fixed in new rev

test/e2e/test_job_submissions.py
AlexTranAmz: Fixed in new rev, compare the entire folder for correct content and structure as well, so no need to combine everything into a dir

src/deadline_worker_agent/sessions/actions/run_step_task.py
```diff
@@ -73,5 +74,9 @@ def start(self, *, session: Session, executor: Executor) -> None:
         session.run_task(
             step_script=self._details.step_template.script,
             task_parameter_values=self._task_parameter_values,
-            os_env_vars={"DEADLINE_SESSIONACTION_ID": self._id, "DEADLINE_TASK_ID": self.task_id},
+            os_env_vars={
+                "DEADLINE_SESSIONACTION_ID": self._id,
+                "DEADLINE_TASK_ID": self.task_id,
+                "ASSET_SYNC_JOB_USER_FEATURE": str(ASSET_SYNC_JOB_USER_FEATURE),
+            },
```
AWS-Samuel, line 79: Why do we need to add this feature flag to the os_env_vars? I don't see it used in any of the other changes in this PR

AlexTranAmz: It was used in the linux_bundle `template.yaml` file to control if we want to scope down the folder or not. This so we can still run test today without failure. The `template.yaml` file current so as a big file so you might need to click on show diff

AWS-Samuel: Ah, I see. 

If possible, we should avoid functional changes which are only used for test code but apply everywhere (such as adding the env var here).

My preference would be to add a job parameter to the job template to handle the chmod logic. The test would set the parameter based on the *default* value for this feature, which would be functionally the same as what we are doing here (since there is currently not a mechanism to turn on feature flags in the[ session-worker](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/test/e2e/conftest.py#L184))

This would allow us to limit our changes to test code and require less clean up when the feature is complete.

test/e2e/utils.py
```diff
@@ -379,3 +379,10 @@ def get_shutdown_on_stop_status_from_toml(
         return cmd_result.stdout.strip()
     else:
         raise Exception(f"Unsupported operating system: {os.environ['OPERATING_SYSTEM']}")
+
+
+def get_feature_value(worker_env_var: Dict[str, str], feature_name: str, default_value: str):
```
AWS-Samuel: I think implementing this function before we made the switch of `List[Dict[str, str]]` -> `Dict[str,str]` made sense, but we no longer need it now that we can use the `.get()` dictionary built-in

```python
worker_env_var.get(feature_name, default_value)
```

test/e2e/test_job_submissions.py
```diff
@@ -1927,19 +1920,33 @@ def test_worker_job_attachments_no_outputs_does_not_fail_job(
         [
             "COPIED",
             # Worker E2E test doesn't run VFS, but this helps verify VIRTUAL fallback and job run successfully
-            "VIRTUAL",
+            # "VIRTUAL",
```
AWS-Samuel: don't think this was supposed to be committed

AWS-Samuel (changes requested):

test/e2e/utils.py
```diff
@@ -379,3 +379,10 @@ def get_shutdown_on_stop_status_from_toml(
...
+    for key, value in worker_env_var.items():
+        if key == feature_name:
+            return value
+    return default_value
```
godobyte: Do we need this util, it can be one line directly being invoked `worker_env_var.get("feature_name", default_value)` right?

test/e2e/test_job_submissions.py
```diff
@@ -1927,19 +1920,33 @@ def test_worker_job_attachments_no_outputs_does_not_fail_job(
...
             os.path.dirname(__file__), "job_attachment_bundle", "dep_data_flow", "linux_bundle"
         )
+        if worker_config.worker_env_var:
+            job_parameters: List[Dict[str, str]] = [
+                {
+                    "name": "AssetSync",
+                    "value": get_feature_value(
+                        worker_env_var=worker_config.worker_env_var,
+                        feature_name="ASSET_SYNC_JOB_USER_FEATURE",
+                        default_value="False",
```
godobyte: What about this to make it explicit?

```suggestion
                    "value": worker_env_var.get("ASSET_SYNC_JOB_USER_FEATURE", "False")
```

test/e2e/test_job_submissions.py
AlexTranAmz: Yes will change in new rev

test/e2e/utils.py
AlexTranAmz: Will change in new rev

test/e2e/test_job_submissions.py
godobyte: Is this a typo?

godobyte (changes requested):

test/e2e/utils.py
AlexTranAmz: as mention above, will fix in new rev!

test/e2e/test_job_submissions.py
AlexTranAmz: Yes :( as mention above, will fix in new rev!

test/e2e/test_job_submissions.py
AlexTranAmz: as mention above, will fix in new rev!

AWS-Samuel (approved):

godobyte (approved):

PR #757, opened 2025-10-07
test: fix s3 validation tests

AlexTranAmz:
Our S3 validation code for better testing assume the S3 prefix start with `Deadline`, but our test account set it up to start with `rootPrefix`

We modify the S3 prefix to start with `rootPrefix`

Mainline E2E passing again

Run against a real job from the test account

No

No

----

### DIFF
--- a/test/e2e/s3_validation_utils.py
+++ b/test/e2e/s3_validation_utils.py
@@ -78,7 +78,7 @@ def validate_s3_job_output_manifest(
 
         for step in steps:
             step_id = step["stepId"]
-            manifest_prefix = f"Deadline/Manifests/{farm_id}/{queue_id}/{job.id}/{step_id}"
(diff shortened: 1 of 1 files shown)


epmog (approved):

erico-aws (approved):

test/e2e/s3_validation_utils.py
```diff
@@ -78,7 +78,7 @@ def validate_s3_job_output_manifest(
 
         for step in steps:
             step_id = step["stepId"]
-            manifest_prefix = f"Deadline/Manifests/{farm_id}/{queue_id}/{job.id}/{step_id}"
+            manifest_prefix = f"rootPrefix/Manifests/{farm_id}/{queue_id}/{job.id}/{step_id}"
```
godobyte, line 81: `queue_details.jobAttachmentSettings` should have `rootPrefex`, could we pass in that value in your next PR rather than hard-code it?

PR #758, opened 2025-10-07
feat: support optional task id in task run session actions

ryyakobe:

WorkerAgent needed to support task chunking workflows where task IDs are not available during session action run. The existing implementation required task IDs for all task run and attachment upload actions, preventing support for chunked steps.

- Made `taskId` optional in `TaskRunAction` and `AttachmentUploadAction` TypedDict definitions
- Implemented dual path format support:
  - New format: `step_id/timestamp_session_action_id/` (when `MANIFEST_REPORTING_FEATURE` feature enabled + no task_id)
  - Old format: `step_id/task_id/timestamp_session_action_id/` (backward compatibility)
- Updated environment variable handling to conditionally set `DEADLINE_TASK_ID` only when task_id is present
- Modified session queue and action classes to handle optional task_id throughout the execution pipeline

- Enables support for task chunking workflows in AWS Deadline Cloud
- Maintains backward compatibility with existing task-based workflows
- No breaking changes to existing API contracts or behavior when feature flag is disabled

- Added comprehensive unit tests covering both optional and required task_id scenarios
- Created 5 new test files
- Tests validate correct path format selection, environment variable handling, and action creation
- Verified backward compatibility with existing task_id workflows

No additional documentation required as this is an internal implementation change that maintains existing API compatibility.

No - this change is backward compatible. Existing workflows continue to function unchanged when the feature flag is disabled or when task_id is provided.

----

### DIFF
--- a/src/deadline_worker_agent/boto/shim.py
+++ b/src/deadline_worker_agent/boto/shim.py
@@ -156,9 +156,10 @@ class DeadlineClient:
             mapped_action = TaskRunAction(
                 sessionActionId=action_id,
                 actionType="TASK_RUN",
-                taskId=action["taskId"],
                 stepId=action["stepId"],
             )
+            if action.get("taskId"):
+                mapped_action["taskId"] = action["taskId"]
             if parameters := action.get("parameters", None):
                 mapped_action["parameters"] = parameters
             return mapped_action
--- a/src/deadline_worker_agent/config/config_file.py
+++ b/src/deadline_worker_agent/config/config_file.py
@@ -116,7 +116,7 @@ line below:
     SESSION_ROOT_DIR = ModifiableSettingData(
         setting_name="session_root_dir",
         table_name="worker",
-        preceding_comment="""
+        preceding_comment=r"""
 The session root directory is a parent directory where worker agent creates per-session
 subdirectories under. This value is overridden when the DEADLINE_WORKER_SESSION_ROOT_DIR environment
 variable is set or the --session-root-dir command-line argument is specified.
--- a/src/deadline_worker_agent/sessions/session.py
+++ b/src/deadline_worker_agent/sessions/session.py
@@ -1276,9 +1276,10 @@ class Session:
                     sessionActionId=current_action.definition._id,
                     actionType="SYNC_OUTPUT_JOB_ATTACHMENTS",
                     stepId=current_action.definition.step_id,
-                    taskId=current_action.definition.task_id,
                     startTime=current_action.start_time.timestamp(),
                 )
+                if current_action.definition.task_id is not None:
+                    action["taskId"] = current_action.definition.task_id
 
                 if not self._action_output_log_filter:
                     self._action_output_log_filter = ActionOutputCaptureFilter(
@@ -1489,6 +1490,7 @@ class Session:
         from .actions import RunStepTaskAction
 
         assert isinstance(current_action.definition, RunStepTaskAction)
+        assert current_action.definition.task_id is not None
 
         upload_summary_statistics: SummaryStatistics = self._asset_sync.sync_outputs(
             s3_settings=s3_settings,
--- a/src/deadline_worker_agent/api_models.py
+++ b/src/deadline_worker_agent/api_models.py
@@ -80,7 +80,7 @@ class ChunkIntParameter(TypedDict):
 class TaskRunAction(TypedDict):
     sessionActionId: str
     actionType: StepActionType
-    taskId: str
+    taskId: NotRequired[str]
     stepId: str
     parameters: NotRequired[
         dict[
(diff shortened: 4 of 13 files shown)


src/deadline_worker_agent/sessions/session.py
```diff
@@ -1489,6 +1490,7 @@ def _sync_asset_outputs(
         from .actions import RunStepTaskAction
 
         assert isinstance(current_action.definition, RunStepTaskAction)
+        assert current_action.definition.task_id is not None
```
erico-aws, line 1493: I'm guessing that this is a safeguard and that conditionals further up the call stack choose a different path if task_id isn't set. Is that right?

ryyakobe: Exactly! This function is only called in old code paths (when `ASSET_SYNC_JOB_USER_FEATURE` is false), where `task_id` should always be present.

The assertion helps keep mypy happy (since `task_id` changed from `str` to `Optional[str]`) and also helps catch potential bugs early by failing fast, in case the old logic is accidentally modified.

src/deadline_worker_agent/boto/shim.py
```diff
@@ -156,9 +156,10 @@ def parse_task_run_action(action: dict, action_id: str) -> TaskRunAction:
             mapped_action = TaskRunAction(
                 sessionActionId=action_id,
                 actionType="TASK_RUN",
-                taskId=action["taskId"],
                 stepId=action["stepId"],
             )
+            if action.get("taskId"):
```
leongdl, line 161: Purely a question, can this just be `mapped_action["taskId"] = action["taskId"]` ? Yes it assigns a `None` element, but we are probably doing `in` checks, or `.get` in a bunch of places already.

src/deadline_worker_agent/sessions/session.py
```diff
@@ -1276,9 +1276,10 @@ def _action_updated_impl(
                     sessionActionId=current_action.definition._id,
                     actionType="SYNC_OUTPUT_JOB_ATTACHMENTS",
                     stepId=current_action.definition.step_id,
-                    taskId=current_action.definition.task_id,
                     startTime=current_action.start_time.timestamp(),
                 )
+                if current_action.definition.task_id is not None:
```
leongdl, line 1281: Same nit as before, since we are using `.get` in many places.

src/deadline_worker_agent/boto/shim.py
ryyakobe: I think I could just change `api_models.py` to use `Optional` instead of `taskId: NotRequired[str]` as we have now. This might simplify the code.  `NotRequired` looked fancy and I thought it communicates the intent a bit better.

test/e2e/utils.py
```diff
@@ -3,7 +3,7 @@
 import logging
 import filecmp
 import json
-import yaml
+import yaml  # type: ignore[import-untyped]
```
godobyte: nit - is this change intended?

src/deadline_worker_agent/config/config_file.py
```diff
@@ -116,7 +116,7 @@ class ModifiableSetting(Enum):
     SESSION_ROOT_DIR = ModifiableSettingData(
         setting_name="session_root_dir",
         table_name="worker",
-        preceding_comment="""
+        preceding_comment=r"""
```
godobyte, line 119: Curious what lead to this addition, do we need one for L133?

ryyakobe: The build and tests succeeded, I just noticed a few warnings in the output:
```
<unknown>:119: DeprecationWarning: invalid escape sequence \P
.../deadline-cloud-worker-agent/src/deadline_worker_agent/config/config_file.py:119: DeprecationWarning: invalid escape sequence \P
  preceding_comment="""```
  ```
  
It's because the string contains `\P` in `"C:\ProgramData\Amazon\OpenJD"` which Python interprets as an escape sequence, so I just decided to fix it.

`WINDOWS_JOB_USER` on L133 doesn't have backslashes, so we should be good there.

src/deadline_worker_agent/boto/shim.py
ryyakobe: I tried to address both of your comments and went back to the original version. Both options are not ideal and I don't think it really matters which we choose. I didn't want to update tests to assert that `taskId` exists and is equal to `None`. Current assertion in tests that `taskId` is not present is a bit cleaner, even though we have those extra `if` statements.

Unfortunately, some of those `get` checks are needed to keep mypy happy.

test/e2e/utils.py
ryyakobe: Removed, thanks. Weird. I think at some point this was giving me build failures, but everything works as is now.

erico-aws (approved):

godobyte (approved):

src/deadline_worker_agent/boto/shim.py
leongdl: OK sounds good. sorry just curious since we have ifs everywhere.

leongdl (approved):

PR #755, opened 2025-10-06
feat(experimental): enhance download process by using WorkerManifestProperties and download_files_from_manifests

godobyte:
### What was the problem/requirement? (What/Why)
The job attachment download needed better tracking of manifest properties including local paths during the download process.

### What was the solution? (How)
1. **Enhanced WorkerManifestProperties model**: Added a new `local_input_manifest_path` field to the `WorkerManifestProperties` class to store the local path to the input manifest file during download operations.

2. **Updated attachment download logic**: Modified the `run_attachment_download.py` to properly set and utilize the WorkerManifestProperties.

3. **Refactored download script**: Enhanced the `attachment_download.py` script to handle the new manifest path information more effectively. Use `download_files_from_manifests` to replace `attachment_download` API.

### What is the impact of this change?
- **Improved tracking**: Better visibility and control over local input manifest paths during job attachment downloads
- **Enhanced reliability**: More robust download process with proper manifest path management  

### How was this change tested?
1. hatch run test
2. run job submission e2e tests

### Was this change documented?
Documentation is tracked separately.

### Is this a breaking change?
No

----

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

PR #762, opened 2025-10-08
feat(experimental): enhance upload process by using WorkerManifestProperties and upload_assets

godobyte:
The attachment upload functionality needed improvements for better reliability and efficiency. By passing list of `WorkerManifestProperties` around and use upload_assets function directly, upload process gets sufficient information to deterministically process all output assets.

Refactored core attachment handling components and added comprehensive test coverage:
- Enhanced `attachment_upload.py` script with improved manifest processing and S3 upload logic
- Updated `run_attachment_upload.py` and `run_attachment_download.py` action handlers
- Expanded `attachment_models.py` with better data structures for attachment management
- Simplified session management code in `session.py` by removing the legacy code path
- Added and streamlined unit tests for attachment upload functionality

Improved reliability and code maintainability for upload process

1. hatch run test
2. run job submission e2e tests

Added docstring. Adding more doc to README is tracked separately.

No

----


src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
```diff
@@ -71,54 +71,55 @@ def __init__(
 
     def set_step_script(
         self,
-        manifest_paths_by_root: dict[str, list[str]],
-        out_rel_dirs_by_root: dict[str, list[str]],
         s3_settings: JobAttachmentS3Settings,
+        worker_manifest_properties_list: list,
```
jericht: nit: can we type the expected objects in the list?
```suggestion
        worker_manifest_properties_list: list[WorkerManifestProperties],
```

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
+        "-wp",
+        "--worker-properties",
+        type=str,
+        help="Path to JSON file containing worker manifest properties configuration",
+        required=True,
+    )
     return parser.parse_args(args)
 
 
 def main(args=None):
```
jericht: nit: can we move main to the bottom of the file just above the `if __name__ == "__main__"` check?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
+        ValueError: If required environment variables are missing
+    """
+
+    # Helper function to get environment variable or raise error if missing
+    def get_env_or_raise(name):
+        """Get required environment variable or raise ValueError if not set."""
+        value = os.environ.get(name)
+        if value is None:
+            raise ValueError(f"Required environment variable '{name}' is not set")
+        return value
```
jericht: Can't we just do `os.environ[...]`?

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
```diff
@@ -8,7 +8,7 @@
 import sys
 import json
 
-from pathlib import Path, PurePosixPath, PureWindowsPath
+from pathlib import Path
```
leongdl, line 11: nit on the commit title. This seems more like "refactor job attachments upload using WorkerManifestProperties to exchange data between worker agent and job-user script"

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -73,6 +75,69 @@ def output_relative_directories(self) -> Optional[List[str]]:
         """Get the output relative directories from manifest properties."""
         return self.manifest_properties.outputRelativeDirectories
 
+    def local_output_relative_directories(self) -> Optional[List[str]]:
+        """
+        Get the output relative directories converted to the local path format.
```
leongdl: This description is hard to grok.

leongdl: This looks like this is normalizing the output directory to host operating system format.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
+from deadline_worker_agent.sessions.attachment_models import (
+    WorkerManifestProperties,
+)
+
+
+def merge(worker_manifest_properties: list[WorkerManifestProperties]) -> dict[str, Optional[str]]:
+    """
+    Merge multiple manifest files for each worker manifest property.
+
+    This function takes a list of worker manifest properties and merges their
```
leongdl: This description is probably "merge multiple input manifest by root right" ? I see it has a for loop that does `_manifest_merge`, but what is is merging with?

leongdl: https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/5504ff38fba53664295a14617e781c64dd76323f/src/deadline_worker_agent/sessions/attachment_models.py#L19

To clarify, maybe state each WorkerManifestProperties represents all manifests within a root.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
 
-python attachment_upload.py \
-    -pm /sessions/session-f63c206fb5f04c04aa17821001aa3847fajfm5x4/path_mapping.json \
-    -s3 s3://test-job-attachment/DeadlineCloud \
-    -mp '{"/sessions/session-e0317487a6cd470084b1c6fd85c789e6ank4lmh5/assetroot-a7714e87e776d9f1c179": "/sessions/session-e0317487a6cd470084b1c6fd85c789e6ank4lmh5/manifests/0bb7eb91fdf8780c4a7e6174de6dfc5e_manifest"}'
-    -od '{"/sessions/session-e0317487a6cd470084b1c6fd85c789e6ank4lmh5/assetroot-a7714e87e776d9f1c179": ["out_dir"]}'
-"""
+        if output:
+            # Store the path to the merged manifest
+            root_path_to_local_manifest[manifest_props.root_path] = output.local_manifest_path
```
AWS-Samuel, line 80: is `manifest_props.root_path` guaranteed to be unique? wondering if this could end up overwriting another root path in some weird edge case

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: Sure `os.environ[]`'s error message is explicit enough, I can make the change.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
godobyte: The current title also captured the fact we're changing from `attachment_upload` to `upload_assets`, I think it's worth mentioning. Thoughts?

src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -73,6 +75,69 @@ def output_relative_directories(self) -> Optional[List[str]]:
...
+                    output_dir = str(Path(PureWindowsPath(output_dir)))
+                elif source_path_format == PathFormat.POSIX:
+                    # Convert Windows path to fit the current platform format
+                    output_dir = str(Path(PurePosixPath(output_dir)))
+
+            local_output_relative_directories.append(output_dir)
+
+        return local_output_relative_directories
+
+    def as_output_metadata(self) -> dict[str, dict[str, str]]:
```
godobyte, line 111: Will cut over to function defined at `ManifestProperties` class once deadline-cloud is released.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
+        Only includes entries where changes were detected.
+    """
+    # Create directory for snapshot manifest files
+    output_path = "manifest_snapshot"
+    root_path_to_output_manifest = {}
+
+    # Process each worker manifest property
+    for manifest_props in worker_manifest_properties:
+        local_root_path = manifest_props.local_root_path
+        output_relative_directories = manifest_props.local_output_relative_directories() or []
```
AWS-Samuel, line 116: since we create a list anyway, is it worth changing the return value of `local_output_relative_directories()` to return an empty list if none are defined?

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: Will reword it in a rev.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
+        if output_manifest:
+            # Log the snapshot information for worker agent processing
+            # ja_snapshot: is a keyword detected by the worker agent log filter
+            print(
+                f"ja_snapshot: {json.dumps({'root': local_root_path, 'manifest': output_manifest.manifest})}"
+            )
+            root_path_to_output_manifest[manifest_props.root_path] = output_manifest.manifest
+        else:
+            # No changes detected - no files to upload
+            print(f"No change detected for root {manifest_props.root_path}")
```
AWS-Samuel, line 140: Is the reason we aren't using the [structured logging classes](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/src/deadline_worker_agent/log_messages.py) here because this runs outside the worker agent process as a separate session

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
-        if manifest:
-            manifests.append(manifest.manifest)
-            print(f"ja_snapshot: {json.dumps({'root': root, 'manifest': manifest.manifest})}")
+    Raises:
+        ValueError: If the file cannot be read or contains invalid JSON
+    """
+    try:
+        with open(file_path, "r") as f:
+            data = json.load(f)
+    except (IOError, OSError, json.JSONDecodeError) as e:
```
AWS-Samuel: If the file doesn't exist will this raise a `FileNotFoundError` ?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
+    # Convert each JSON object to WorkerManifestProperties
+    return [WorkerManifestProperties.from_dict(item) for item in data]
 
 
 def parse_args(args):
-    parser = argparse.ArgumentParser()
-    parser.add_argument("-pm", "--path-mapping", type=str, help="", required=True)
-    parser.add_argument("-s3", "--s3-uri", type=str, help="", required=True)
-    parser.add_argument("-mp", "--manifest-paths-by-root", type=json.loads, required=True)
-    parser.add_argument("-od", "--out-rel-dirs-by-root", type=json.loads, required=False)
```
AWS-Samuel, line 136: Is this CLI only used internally? If it's called from outside the WA then this seems like a breaking change

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,349 @@
...
-            path_mapping_rules=parsed_args.path_mapping,
+        manifest_infos = upload_output_assets(
+            s3_uri=parsed_args.s3_uri,
+            worker_manifest_properties=worker_manifest_properties,
+            root_path_to_output_manifest=root_path_to_output_manifest,
+        )
+
+        # Check if manifest reporting feature is enabled via environment variable
+        manifest_reporting_enabled = (
+            os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
```
AWS-Samuel: Will this work if this process is run in a session as the job user? (I'm not familiar with where this feature flag is set)

src/deadline_worker_agent/sessions/attachment_models.py
AWS-Samuel, line 111: Is it worth creating a `TypedDict` for this?

godobyte: I wouldn't invest more effort here, this function is defined at `ManifestProperties` class in deadline-cloud, I'll use that one when it gets released.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: That's a very good question. This is passed in via CreateJob. If the job is submitted by bundle_submit, we can guarantee it to be unique. However I don't see it in CreateJob API spec to guarantee the uniqueness.

I'd like to dive deeper and take this as a follow-up, the expected behaviour is unclear at the moment.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: Right, similar to customer task run logs, this is run in openjd sub-process and printed out to session logs.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: Good catch, I'll catch generic errors to capture.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: It's only meant to be used internally. We don't vend it for external usage.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: This process is run in a session as the job user and the feature flag is passed in from agent user to the job user already. 

Here is the reference, it's passed in via `run_task` call 
https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/ae885cd8e65ef9f79ee76db1725ce710db18286a/src/deadline_worker_agent/sessions/actions/run_attachment_upload.py#L196

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: The function is designed to convert output directories to local path format. If there are no output directories defined (None), then there's nothing to convert. Returning None preserves the "no output directories" state. 

I prefer keep the consistency unless you have strong opinion about it.

src/deadline_worker_agent/sessions/actions/run_attachment_download.py
leongdl: > The current title also captured the fact we're changing from attachment_upload to upload_assets, I think it's worth mentioning. Thoughts?

That's a good thing to put in the PR description body?

leongdl (dismissed):
thanks for addressing comments.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -1,172 +1,348 @@
...
+            root=manifest_props.to_path_mapping_rule().source_path,
+            # paths to manifest files to be merged
+            manifest_files=manifest_props.local_manifest_paths,
+            # directory to put the generated merged manifests
+            destination=manifest_path,
+            name="merge",
+        )
 
-The manifest snapshot and attachment upload commands are available in deadline-cloud as python API and AWS Deadline Cloud CLI.
+        if not root_path_to_local_manifest.get(manifest_props.root_path):
```
AWS-Samuel: I think we want

```suggestion
        if root_path_to_local_manifest.get(manifest_props.root_path) is not None:
```

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
AWS-Samuel: Synced offline that this is the same as previous behavior, and that other logic should prevent this edge case. Agreed to add a log statement to get some insight just in case

AWS-Samuel (approved):

jericht (approved):

godobyte:
Keep the title as-is to be symmetrical to 

[feat(experimental): enhance download process by using WorkerManifestProperties and download_files_from_manifests (](https://github.com/aws-deadline/deadline-cloud-worker-agent/commit/0272a468157b893165d5f855d52b5c50e7d43e96)https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/755)]

PR #766, opened 2025-10-10
chore(deps): update to latest deadline-cloud version (0.53.1)

justinsaws:
There was a release of deadline-cloud
Update to the latest version
We are targeting the latest version of Deadline Cloud
`hatch build && hatch run all:test`
No.
No.
----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ license = "Apache-2.0"
 dependencies = [
     "requests ~= 2.31",
     "boto3 >= 1.34.75",
-    "deadline >= 0.50,< 0.53",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -13,7 +13,7 @@ license = "Apache-2.0"
 dependencies = [
     "requests ~= 2.31",
     "boto3 >= 1.34.75",
-    "deadline >= 0.50,< 0.53",
+    "deadline >= 0.53.1",
```
evanspearman-a, line 16: Was the change in the way we're pinning the version intentional?

PR #768, opened 2025-10-10
test: Add E2E submit job using createJob API

AlexTranAmz:
We need an end-to-end test that covers many cases as one job by directly calling `CreateJob`. `CreateJob` gives more control over the input over `deadline bundle submit`. It also fills the testing gap that customer directly calling `CreateJob`

Add Linux and Windows tests that using `CreateJob` with a complex job bundle include:
- Have more than one manifests and root path
- Have step-step dependency
- Have output

And for Linux, test also setup storage profile
Better E2E test coverage

Running E2E test:
- Windows
```
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================================================================================================== slowest 5 durations ===================================================================================================================================================
315.92s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_create_job_API_call_windows
217.28s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_create_job_API_call_windows
6.34s teardown test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_create_job_API_call_windows
================================================================================================================================ 1 passed, 59 deselected, 1 warning in 539.66s (0:08:59) =================================================================================================================================
```
- Linux
```
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================================================================================================== slowest 5 durations ===================================================================================================================================================
210.09s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_create_job_API_call_linux[COPIED]
131.65s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_create_job_API_call_linux[VIRTUAL]
102.69s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_create_job_API_call_linux[COPIED]
6.40s teardown test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_create_job_API_call_linux[VIRTUAL]

(1 durations < 0.005s hidden.  Use -vv to show these durations.)
================================================================================================================================ 2 passed, 60 deselected, 1 warning in 450.95s (0:07:30) =================================================================================================================================
```

----

### DIFF
--- a/src/deadline_worker_agent/api_models.py
+++ b/src/deadline_worker_agent/api_models.py
@@ -80,7 +80,7 @@ class ChunkIntParameter(TypedDict):
 class TaskRunAction(TypedDict):
     sessionActionId: str
     actionType: StepActionType
-    taskId: str
+    taskId: NotRequired[str]
     stepId: str
     parameters: NotRequired[
         dict[
@@ -106,7 +106,7 @@ class AttachmentUploadAction(TypedDict):
     sessionActionId: str
     actionType: AttachmentUploadActionType
     stepId: str
-    taskId: str
+    taskId: NotRequired[str]
     startTime: float
 
 
--- a/src/deadline_worker_agent/boto/shim.py
+++ b/src/deadline_worker_agent/boto/shim.py
@@ -156,9 +156,10 @@ class DeadlineClient:
             mapped_action = TaskRunAction(
                 sessionActionId=action_id,
                 actionType="TASK_RUN",
-                taskId=action["taskId"],
                 stepId=action["stepId"],
             )
+            if action.get("taskId"):
+                mapped_action["taskId"] = action["taskId"]
             if parameters := action.get("parameters", None):
                 mapped_action["parameters"] = parameters
             return mapped_action
--- a/src/deadline_worker_agent/config/config_file.py
+++ b/src/deadline_worker_agent/config/config_file.py
@@ -116,7 +116,7 @@ line below:
     SESSION_ROOT_DIR = ModifiableSettingData(
         setting_name="session_root_dir",
         table_name="worker",
-        preceding_comment="""
+        preceding_comment=r"""
 The session root directory is a parent directory where worker agent creates per-session
 subdirectories under. This value is overridden when the DEADLINE_WORKER_SESSION_ROOT_DIR environment
 variable is set or the --session-root-dir command-line argument is specified.
--- a/src/deadline_worker_agent/scheduler/session_queue.py
+++ b/src/deadline_worker_agent/scheduler/session_queue.py
@@ -483,7 +483,7 @@ class SessionActionQueue:
                 action_queue_entry = cast(TaskRunQueueEntry, action_queue_entry)
                 action_definition = action_queue_entry.definition
                 step_id = action_definition["stepId"]
-                task_id = action_definition["taskId"]
+                task_id = action_definition.get("taskId")
                 try:
                     step_details = self._job_entities.step_details(step_id=step_id)
                 except UnsupportedSchema as e:
@@ -509,15 +509,14 @@ class SessionActionQueue:
                     id=action_id,
                     details=step_details,
                     task_parameter_values=task_parameters,
-                    task_id=action_definition["taskId"],
+                    task_id=action_definition.get("taskId"),
                 )
             elif action_type == "SYNC_OUTPUT_JOB_ATTACHMENTS":
                 action_queue_entry = cast(AttachmentUploadActionQueueEntry, action_queue_entry)
                 action_definition = action_queue_entry.definition
                 step_id = action_definition["stepId"]
-                task_id = action_definition["taskId"]
+                task_id = action_definition.get("taskId")
                 start_time = action_definition["startTime"]
-
                 next_action = AttachmentUploadAction(
                     id=action_id,
                     session_id=self._session_id,
--- a/src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
+++ b/src/deadline_worker_agent/sessions/actions/run_attachment_upload.py
@@ -51,7 +51,7 @@ class AttachmentUploadAction(OpenjdAction):
 
     _step_script: Optional[StepScript_2023_09]
     _step_id: str
-    _task_id: str
+    _task_id: Optional[str]
     _start_time: float
 
     def __init__(
@@ -60,7 +60,7 @@ class AttachmentUploadAction(OpenjdAction):
         id: str,
         session_id: str,
         step_id: str,
-        task_id: str,
+        task_id: Optional[str] = None,
         start_time: float,
     ) -> None:
         super(AttachmentUploadAction, self).__init__(
@@ -183,15 +183,20 @@ class AttachmentUploadAction(OpenjdAction):
         )
 
         assert self._step_script is not None
+
+        env_vars = {
+            "DEADLINE_SESSIONACTION_ID": self._id,
+            "DEADLINE_SESSIONACTION_START_TIME": str(self._start_time),
+            "DEADLINE_STEP_ID": self._step_id,
+            "MANIFEST_REPORTING_FEATURE": str(MANIFEST_REPORTING_FEATURE),
+        }
+
+        if self._task_id is not None:
+            env_vars["DEADLINE_TASK_ID"] = self._task_id
+
         session.run_task(
             step_script=self._step_script,
             task_parameter_values=dict[str, ParameterValue](),
-            os_env_vars={
-                "DEADLINE_SESSIONACTION_ID": self._id,
-                "DEADLINE_SESSIONACTION_START_TIME": str(self._start_time),
-                "DEADLINE_STEP_ID": self._step_id,
-                "DEADLINE_TASK_ID": self._task_id,
-                "MANIFEST_REPORTING_FEATURE": str(MANIFEST_REPORTING_FEATURE),
-            },
+            os_env_vars=env_vars,
             log_task_banner=False,
         )
--- a/src/deadline_worker_agent/sessions/actions/run_step_task.py
+++ b/src/deadline_worker_agent/sessions/actions/run_step_task.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 from concurrent.futures import Executor
-from typing import Any, TYPE_CHECKING
+from typing import Any, Optional, TYPE_CHECKING
 
 from openjd.model import TaskParameterSet
 
@@ -25,13 +25,13 @@ class RunStepTaskAction(OpenjdAction):
         The unique step identifier
     details : StepDetails
         The environment details
-    task_id : str
+    task_id : Optional[str]
         The unique task identifier
     task_parameter_values : TaskParameterSet
         The task parameter values
     """
 
-    task_id: str
+    task_id: Optional[str]
     _details: StepDetails
     _task_parameter_values: TaskParameterSet
 
@@ -40,7 +40,7 @@ class RunStepTaskAction(OpenjdAction):
         *,
         id: str,
         details: StepDetails,
-        task_id: str,
+        task_id: Optional[str] = None,
         task_parameter_values: TaskParameterSet,
     ) -> None:
         super(RunStepTaskAction, self).__init__(
@@ -70,11 +70,12 @@ class RunStepTaskAction(OpenjdAction):
         executor : Executor
             An executor for running futures
         """
+        env_vars = {"DEADLINE_SESSIONACTION_ID": self._id}
+        if self.task_id is not None:
+            env_vars["DEADLINE_TASK_ID"] = self.task_id
+
         session.run_task(
             step_script=self._details.step_template.script,
             task_parameter_values=self._task_parameter_values,
-            os_env_vars={
-                "DEADLINE_SESSIONACTION_ID": self._id,
-                "DEADLINE_TASK_ID": self.task_id,
-            },
+            os_env_vars=env_vars,
         )
--- a/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
+++ b/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
@@ -269,15 +269,7 @@ def upload_output_assets(
     """
 
     # Generate S3 upload path using session context from environment variables
-    s3_upload_path = JobAttachmentS3Settings.partial_session_action_manifest_prefix(
-        farm_id=os.environ["DEADLINE_FARM_ID"],
-        queue_id=os.environ["DEADLINE_QUEUE_ID"],
-        job_id=os.environ["DEADLINE_JOB_ID"],
-        step_id=os.environ["DEADLINE_STEP_ID"],
-        task_id=os.environ["DEADLINE_TASK_ID"],
-        session_action_id=os.environ["DEADLINE_SESSIONACTION_ID"],
-        time=float(os.environ["DEADLINE_SESSIONACTION_START_TIME"]),
-    )
+    s3_upload_path = build_s3_manifest_path()
 
     # Create S3 settings from the provided URI
     s3_settings = JobAttachmentS3Settings.from_s3_root_uri(s3_uri)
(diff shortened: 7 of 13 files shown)


test/e2e/test_job_submissions.py
```diff
@@ -3282,3 +3284,185 @@ def test_worker_job_attachments_output_only(
...
+                job=job,
+                deadline_client=deadline_client,
+            )
+            LOG.info("S3 validation completed successfully")
+
+            # Get job output path
+            output_root_path = tempfile.mkdtemp(
+                dir=self.JOB_OUTPUT_PATH, prefix=f"dep_data_flow_linux-{file_system}"
+            )
+            output_path: dict[str, list[str]] = wait_for_job_output(
```
leongdl: Does this timeout?

test/e2e/test_job_submissions.py
```diff
@@ -3282,3 +3284,185 @@ def test_worker_job_attachments_output_only(
...
+
+            # Create Job API
+            job = submit_job_from_create_job_API(
+                deadline_client=deadline_client,
+                deadline_resources=deadline_resources,
+                bundle_path=job_bundle_path,
+                debug_snapshot_dir=snap_dir,
+            )
+
+            job.wait_until_complete(client=deadline_client)
```
leongdl: Does this timeout?

test/e2e/test_job_submissions.py
leongdl: Or i guess it is the same as all other tests?

test/e2e/test_job_submissions.py
```diff
@@ -3282,3 +3284,185 @@ def test_worker_job_attachments_output_only(
...
+
+            set_setting("defaults.farm_id", deadline_resources.farm.id, config)
+            set_setting("defaults.queue_id", deadline_resources.queue_a.id, config)
+            set_setting(
+                "settings.storage_profile_id",
+                deadline_resources.queue_a_job_storage_profile_id,
+                config,
+            )
+
+            submit_job_from_bundle(
```
godobyte: Do we need to call submit from bundle too, or this is called to generate the content in `snap_dir`? Maybe worth adding some comments to explain.

test/e2e/test_job_submissions.py
```diff
@@ -3282,3 +3284,185 @@ def test_worker_job_attachments_output_only(
...
+                debug_snapshot_dir=snap_dir,
+            )
+
+            # Create Job API
+            job = submit_job_from_create_job_API(
+                deadline_client=deadline_client,
+                deadline_resources=deadline_resources,
+                bundle_path=job_bundle_path,
+                debug_snapshot_dir=snap_dir,
+            )
```
AWS-Samuel: It looks like we submit the job twice here but only reference and check the second submission. Is the first one necessary? (Happens below too for windows)

test/e2e/test_job_submissions.py
```diff
@@ -3282,3 +3284,185 @@ def test_worker_job_attachments_output_only(
...
+                job=job,
+                deadline_client=deadline_client,
+                deadline_resources=deadline_resources,
+                output_root_path=output_root_path,
+            )
+            LOG.info(f"output_path dict is: {output_path}")
+
+            # Verify the final output file exists and contains expected content
+            verify_output_dir_matches(
+                reference_dir_path=f"{os.path.dirname(__file__)}/job_attachment_bundle/dep_data_flow/correct_output",
```
godobyte: Is the existing step-dep bundle job test using the same output dir, could you check whether that's cleaned up so the results don't interfere each other?

pyproject.toml
```diff
@@ -13,7 +13,7 @@ license = "Apache-2.0"
 dependencies = [
     "requests ~= 2.31",
     "boto3 >= 1.34.75",
-    "deadline >= 0.50,< 0.53",
```
AWS-Samuel: If this is to bump the deadline version, we shouldn't pin to a patch version


```suggestion
    "deadline >= 0.50,< 0.54",
```

test/e2e/utils.py
```diff
@@ -379,3 +385,115 @@ def get_shutdown_on_stop_status_from_toml(
...
+            job_template = json.loads(f.read())
+    else:
+        LOG.error(
+            f"Was expecting to find either {template_file_name}.yaml or {template_file_name}.json in directory {bundle_path} but found none."
+        )
+        raise FileNotFoundError
+
+    queue_details = get_queue(
+        farm_id=deadline_resources.farm.id, queue_id=deadline_resources.queue_a.id
+    )
```
godobyte: This assumes the job submitted to `queue_a`, which may not always be true. Could we pass the farm and queue in, similar to `submit_job_from_bundle`? That'd make it explicit and improve reusability.

test/e2e/utils.py
```diff
@@ -379,3 +385,115 @@ def get_shutdown_on_stop_status_from_toml(
...
+        with open(yaml_path) as f:
+            job_template = yaml.safe_load(f.read())
+    elif os.path.isfile(json_path):
+        with open(json_path) as f:
+            job_template = json.loads(f.read())
+    else:
+        LOG.error(
+            f"Was expecting to find either {template_file_name}.yaml or {template_file_name}.json in directory {bundle_path} but found none."
+        )
+        raise FileNotFoundError
```
godobyte: Do we need these, I thought the template file is captured under `debug_snapshot_dir`?

I saw the only usage of `﻿﻿job_template` is a print statement, did I miss other usage?

test/e2e/utils.py
```diff
@@ -379,3 +385,115 @@ def get_shutdown_on_stop_status_from_toml(
...
+
+    # Load parameters file
+    with open(f"{debug_snapshot_dir}/parameters_param.json", "r") as f:
+        parameters = json.load(f)
+
+    response = deadline_client.create_job(
+        farmId=deadline_resources.farm.id,
+        queueId=deadline_resources.queue_a.id,
+        template=template_content,
+        templateType="YAML",
```
godobyte: Is it always `YAML`?

test/e2e/utils.py
```diff
@@ -379,3 +385,115 @@ def get_shutdown_on_stop_status_from_toml(
...
+        attachments = json.load(f)
+
+    # Load parameters file
+    with open(f"{debug_snapshot_dir}/parameters_param.json", "r") as f:
+        parameters = json.load(f)
+
+    response = deadline_client.create_job(
+        farmId=deadline_resources.farm.id,
+        queueId=deadline_resources.queue_a.id,
+        template=template_content,
```
godobyte: I believe this is the actual template used for submission, which is read from `f"{debug_snapshot_dir}/template_param.data"`.

test/e2e/utils.py
```diff
@@ -379,3 +385,115 @@ def get_shutdown_on_stop_status_from_toml(
...
+
+    return Job(
+        farm=deadline_resources.farm,
+        queue=deadline_resources.queue_a,
+        template=job_template,
+        **job_details,
+    )
+
+
+def upload_directory_to_s3(s3_client: Any, local_dir: str, bucket: str, s3_prefix: str) -> None:
```
godobyte: Is there any existing function that we could reuse?

godobyte (changes requested):

test/e2e/test_job_submissions.py
AlexTranAmz: use to generate snap dir, but remove in next rev, in favour of manually create a job bundle

pyproject.toml
AlexTranAmz: Fix in next rev

test/e2e/test_job_submissions.py
AlexTranAmz: The first one is to generate content using snap-dir to later use for createjob API. But we decided to remove this confusion and just manually create a job bundle

test/e2e/test_job_submissions.py
AlexTranAmz: It does not, behave the same as other test, only changes is method of job submission

test/e2e/test_job_submissions.py
AlexTranAmz: It does not, behave the same as other test, only changes is method of job submission

test/e2e/test_job_submissions.py
AlexTranAmz: In the next rev, moved to a diff folder

test/e2e/utils.py
AlexTranAmz: That's a good catch, since I copy over from the submit job bundle. Removed in the new rev

test/e2e/utils.py
AlexTranAmz: Fixed in next rev

test/e2e/utils.py
AlexTranAmz: Yes, removed the not needed part

test/e2e/utils.py
AlexTranAmz: Yes in this case, it's always YAML

test/e2e/utils.py
AlexTranAmz: Try to take a look but from my research, not :(

test/e2e/utils.py
```diff
@@ -52,14 +54,18 @@ def wait_for_job_output(
...
+            job_output_downloader.set_root_path(
+                list(output_paths_by_root.keys())[0], os.path.abspath(output_root_path)
             )
-
-        job_output_downloader.set_root_path(
-            list(output_paths_by_root.keys())[0], os.path.abspath(output_root_path)
-        )
+        else:
+            # Multiple root paths - process each one with a subdirectory
+            for i, root_path in enumerate(output_paths_by_root.keys()):
```
AWS-Samuel: Is the order of `.keys()` deterministic? Should we sort this first?

test/e2e/utils.py
```diff
@@ -168,6 +176,8 @@ def submit_job_from_bundle(
         create_job_args["job_parameters"] = job_parameters  # type: ignore
 
     job_id = create_job_from_job_bundle(**create_job_args)  # type: ignore
+    if debug_snapshot_dir:
+        return  # type: ignore
```
AWS-Samuel: I don't think we should `type: ignore` this. By returning `None`, callers do need to handle the case where there is no return value.

test/e2e/test_job_submissions.py
```diff
@@ -3282,3 +3283,109 @@ def test_worker_job_attachments_output_only(
...
+    )
+    def test_worker_create_job_API_call_linux(
+        self,
+        deadline_resources: DeadlineResources,
+        deadline_client: DeadlineClient,
+        session_worker: EC2InstanceWorker,
+    ) -> None:
+        job_bundle_path: str = os.path.join(
+            os.path.dirname(__file__), "job_attachment_bundle", "complex_bundle", "linux"
+        )
```
AWS-Samuel: bit a of a nit that `complex` isn't very descriptive. Would like if we could capture this bit from your PR description in a doc string here so that future readers can quickly grasp what makes the bundle complex

> Add Linux and Windows tests that using CreateJob with a complex job bundle include:
> 
> Have more than one manifests and root path
> Have step-step dependency
> Have output
> And for Linux, test also setup storage profile

test/e2e/utils.py
AlexTranAmz: Good catch! Fixed in next rev

test/e2e/utils.py
AlexTranAmz: Remove this as it's not needed anymore

test/e2e/test_job_submissions.py
AlexTranAmz: Added in the next rev

test/e2e/utils.py
```diff
@@ -52,14 +54,18 @@ def wait_for_job_output(
 
     # Download file and place it into the output_paths_by_root
     if output_root_path is not None:
-        if len(output_paths_by_root) != 1:
-            raise NotImplementedError(
-                f"Currently do not support more than one root paths, provided {output_paths_by_root.keys()}"
+        if len(output_paths_by_root) == 1:
```
leongdl: nit: Can we just collapse this if else so the output is always downloading to `root_output_path = os.path.join(output_root_path, f"root_{i}")` ? Simpler test code understanding.

test/e2e/utils.py
```diff
@@ -379,3 +384,105 @@ def get_shutdown_on_stop_status_from_toml(
         return cmd_result.stdout.strip()
     else:
         raise Exception(f"Unsupported operating system: {os.environ['OPERATING_SYSTEM']}")
+
+
+def submit_job_from_create_job_API(
```
leongdl: Super nit, a comment for this function is good. This function does not use deadline cloud lib. It manually performs and copies the data, manifest files then calls the CreateJob API.

test/e2e/utils.py
```diff
@@ -379,3 +384,105 @@ def get_shutdown_on_stop_status_from_toml(
...
+
+
+def upload_directory_to_s3(s3_client: Any, local_dir: str, bucket: str, s3_prefix: str) -> None:
+    """Upload a directory recursively to S3"""
+    local_path = Path(local_dir)
+    if not local_path.exists():
+        LOG.info(f"Warning: {local_dir} does not exist, skipping upload")
+        return
+
+    for file_path in local_path.rglob("*"):
```
leongdl: Nit for future, this can be a nice executor to parallelize since files can get larger. Its ok for now since the files are small.

leongdl (dismissed):
OK - but think about the suggestion to simplify the single root and multi root test case.

AWS-Samuel (dismissed):

pyproject.toml
godobyte, line 16: You'll need to merge the recent changes as this is already updated in mainline.

test/e2e/utils.py
godobyte: +1, a description for on the args would be useful too.

test/e2e/utils.py
```diff
@@ -379,3 +384,105 @@ def get_shutdown_on_stop_status_from_toml(
...
+    deadline_resources: DeadlineResources,
+    farm: Farm,
+    queue: Queue,
+    bundle_path: str,
+    debug_snapshot_dir: str,
+    template_file_name: str = "template",
+    storage_profile: bool = False,
+) -> Job:
+    bundle_path = os.path.normpath(bundle_path)
+    LOG.info(f"Submitting bundle {bundle_path} to farm {farm.id} and queue {queue.id}")
```
godobyte: I thought submission content is taken from the `debug_snapshot_dir`, what do you need this `bundle_path` for? Apparently it's only being used in the next log line.

test/e2e/test_job_submissions.py
```diff
@@ -3282,3 +3283,128 @@ def test_worker_job_attachments_output_only(
...
+        os.makedirs(name=self.JOB_OUTPUT_PATH, exist_ok=True)
+
+        # Create Job API
+        job = submit_job_from_create_job_API(
+            deadline_client=deadline_client,
+            deadline_resources=deadline_resources,
+            farm=deadline_resources.farm,
+            queue=deadline_resources.queue_a,
+            bundle_path=job_bundle_path,
+            debug_snapshot_dir=job_bundle_path,
```
godobyte: Only one of the params should be needed here, they're both passing in `job_bundle_path`.

pyproject.toml
godobyte, line 16: The version has already been updated in mainline, you may get conflict from here.

godobyte (changes requested):

pyproject.toml
AlexTranAmz: Pull in the changes

test/e2e/test_job_submissions.py
AlexTranAmz: Fixed in next rev

test/e2e/utils.py
AlexTranAmz: good catch, remove the not need field in the next rev

pyproject.toml
AlexTranAmz: Merged

test/e2e/utils.py
AlexTranAmz: Got it! Tks

test/e2e/utils.py
AlexTranAmz: I will change it in another PR as that will introduce more file path changes to existing job bundle that do not in scope of this PR

leongdl: ok

PR #769, opened 2025-10-10
chore(experimental): add metrics for attachment_upload

jericht:
We need metrics for attachment_upload, the experimental implementation for syncing assets as the job user

Add metrics for success/fail, detailed error reasons, and latencies

We have above metrics

- Added unit tests
- Manually calling the CLI to invoke each telemetry event

N/A

No

----

### DIFF
--- a/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
+++ b/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
@@ -13,11 +13,13 @@ to handle job output file uploads after task completion.
 
 #! /usr/bin/env python3
 import argparse
+import dataclasses
+import functools
 import sys
 import time
 import os
 import json
-from typing import Optional
+from typing import cast, Any, Callable, Optional, TypeVar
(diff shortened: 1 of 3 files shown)


leongdl (dismissed):

AWS-Samuel (approved):

godobyte (approved):

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -32,11 +34,49 @@
...
+
+@dataclasses.dataclass
+class AttachmentUploadLatences:
+    """Stores metrics for function latencies in this script"""
+
+    merge: int = 0
+    snapshot: int = 0
+    parse_worker_manifest_properties: int = 0
+    upload_output_assets: int = 0
+    total: int = 0
```
godobyte, line 76: nit - apparently the int is representing nano-seconds from the implementation below, maybe that can be in the name or comment to be explicit.

PR #776, opened 2025-10-20
test: add e2e test to verify empty outputRelativeDirectories lead to no output

godobyte:
ouputRelativeDirectories is an optional field in [ManifestProperties](https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_ManifestProperties.html) as part of CreateJob API model. If not provided upon calling CreateJob, the value will be None. When the list is None, nothing gets recognize as upload for upload.

This can be easily overlooked, we should have a test to guard the expected behaviour.

Add tests to guard the expected behaviour when `ouputRelativeDirectories` is None.

Better test coverage.

#### Verified both windows and linux tests pass.
```
test/e2e/test_job_submissions.py::TestJobSubmission::test_job_attachments_no_output_relative_directories
=================== 1 passed, 1 warning in 342.44s (0:05:42) ===================
```
#### Verified no job output

windows
```
 ~> deadline job download-output --job-id job-1bd1c08da4f94ec3a991d8e6c6fc9572
Downloading output from Job 'empty-outputRelativeDirectories-test-windows'
There are no output files available for download at this moment. Please verify that the Job/Step/Task you are trying to download output from has completed successfully.
```
linux
```
 ~> deadline job download-output --job-id job-c1fc3dee3df54c6f9f60a90cb7a00272
Downloading output from Job 'empty-outputRelativeDirectories-test-linux'
There are no output files available for download at this moment. Please verify that the Job/Step/Task you are trying to download output from has completed successfully.
```

----

### DIFF
--- a/test/e2e/test_job_submissions.py
+++ b/test/e2e/test_job_submissions.py
@@ -3406,3 +3406,156 @@ with open(output_path, "w") as f:
             reference_dir_path=f"{os.path.dirname(__file__)}/job_attachment_bundle/complex_bundle/windows/correct_output",
             output_dir_path=output_root_path,
         )
+
+    def test_job_attachments_no_output_relative_directories(
+        self,
+        deadline_resources: DeadlineResources,
+        deadline_client: DeadlineClient,
+        session_worker: EC2InstanceWorker,
+        test_runner_identity: dict[str, str],
+    ) -> None:
+        """
+        Tests that a job created with multiple manifests that have the same rootPath and both manifests have a file with the same path
+        results in the file from the second manifest taking precedence over the first manifest
+        """
+        # Get S3 settings from queue using boto3 directly
+        queue_info = deadline_client._real_client.get_queue(
+            farmId=deadline_resources.farm.id, queueId=deadline_resources.queue_a.id
+        )
+        s3_bucket = queue_info["jobAttachmentSettings"]["s3BucketName"]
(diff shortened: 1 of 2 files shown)


test/e2e/test_job_submissions.py
```diff
@@ -3406,3 +3406,156 @@ def test_worker_create_job_API_call_windows(
...
+        # Upload manifests
+        s3.put_object(
+            Bucket=s3_bucket,
+            Key=f"{s3_prefix}/Manifests/manifest1hash",
+            Body=json.dumps(manifest),
+            ExpectedBucketOwner=test_runner_identity["Account"],
+        )
+
+        # Create and submit job using boto3 directly
+        template = {
```
leongdl, line 3467: Is there anyway to put this job template to a file and just source it in code?

leongdl: This would make a good job bundle to submit.

godobyte: We could, but I'd track that task separately because this test currently uses CreateJob (instead of bundle submit) to also test any s3 input path, which is not applicable with `bundle submit`.

AWS-Samuel (approved):

test/e2e/test_job_submissions.py
leongdl: ok no problem.

leongdl (approved):

PR #778, opened 2025-10-21
test(experimental): add job attachments E2E tests for storage profile path mapping

jericht:
We need E2E tests cases for Job Attachments behaviour when storage profiles are being used for Windows <-> Windows, Windows <-> Linux, Linux <-> Windows

- Add above test cases
- Move JA specific test cases into dedicated test file since the main file was getting large and it will be easier to have core Worker tests and JA tests in separate files.

We have test coverage for above

- Ran the E2E tests in my dev account with both Linux and Windows workers
- Deploy the CFN and run the tests against the resources deployed by the CFN with both Linux and Windows workers

No

No

----


scripts/e2e_testing_infrastructure.yaml
```diff
@@ -664,13 +667,25 @@ Resources:
           Path: "/tmp/storageprofiletest"
           Type: "LOCAL"
 
+  WindowsJobStorageProfile:
```
godobyte, line 676: Do we need a windows Fleet storage profile?

test/e2e/test_job_attachments.py
```diff
@@ -0,0 +1,252 @@
...
+        submission_os: str,
+        deadline_resources: DeadlineResources,
+        deadline_client: DeadlineClient,
+        session_worker: EC2InstanceWorker,
+        tmp_path: pathlib.Path,
+    ) -> None:
+        # Test that submits a job that has storage profile and confirm that the final output path and content is as we expect
+
+        if submission_os == "linux":
+            storage_profile_id = deadline_resources.queue_a_job_storage_profile_id
```
godobyte: nit - we could have named this SP more explicit with linux to differentiate the windows one.

test/e2e/test_job_attachments.py
```diff
@@ -0,0 +1,252 @@
...
+        # Create a unique root directory to not conflict with other test runs
+        test_disambiguator = uuid.uuid4().hex
+
+        # Create the initial input file
+        input_asset = Asset(path=os.path.join(test_disambiguator, "files", "test_input_file"))
+
+        asset_s3_key = f"{s3_prefix}/Data/{input_asset.hash}"
+        LOG.info(f"Uploading input file to s3://{s3_bucket}/{asset_s3_key}: {input_asset}")
+        s3 = boto3.client("s3")
+        s3.put_object(
```
godobyte, line 120: This may lead to security finding that requires to add `ExpectedBucketOwner`, this attribute ensures it's from expected account.

test/e2e/test_job_attachments.py
```diff
@@ -0,0 +1,252 @@
...
+            Bucket=s3_bucket,
+            Key=asset_s3_key,
+            Body=input_asset.content,
+        )
+
+        # Create asset manifest
+        manifest = {
```
godobyte, line 127: nit - do we have existing data model from client lib to construct manifest?

test/e2e/test_job_attachments.py
```diff
@@ -0,0 +1,252 @@
...
+            "manifestVersion": "2023-03-03",
+            "paths": [
+                input_asset.to_dict(),
+            ],
+            "totalSize": input_asset.size,
+        }
+
+        manifest_s3_key = f"{s3_prefix}/Manifests/manifesthash"
+        LOG.info(f"Uploading manifest file to s3://{s3_bucket}/{manifest_s3_key}: {manifest}")
+        s3.put_object(
```
godobyte, line 138: ditto

test/e2e/test_job_attachments.py
```diff
@@ -0,0 +1,252 @@
...
+            f"Expected exactly one output root, but got: {output_root_to_file_mappings}"
+        )
+        output_root_path, output_file_rel_paths = list(output_root_to_file_mappings.items())[0]
+        output_file_paths = [p for p in output_file_rel_paths if p.endswith("output_file")]
+        assert len(output_file_paths) == 1, (
+            f"Expected exactly one output file, but got: {output_file_paths}"
+        )
+        output_file_path = os.path.join(output_root_path, output_file_paths[0])
+        try:
+            with open(output_file_path, "r", encoding="utf-8-sig") as output_file:
```
godobyte, line 264: I see it's being used in old code that's being migrated to this new test, curious whether you know why we use `utf-8-sig` encoding here.

scripts/e2e_testing_infrastructure.yaml
jericht: Added

test/e2e/pre_test_cleanup.py
```diff
@@ -228,8 +228,9 @@ def get_farm() -> str:
 def cleanup_test_environment() -> None:
     config = Config(retries={"mode": "adaptive"})
 
-    ec2_client = boto3.client("ec2", config=config)
-    cleanup_ec2_instances(ec2_client)
+    if os.getenv("KEEP_WORKER_AFTER_FAILURE", "").lower() != "true":
+        ec2_client = boto3.client("ec2", config=config)
+        cleanup_ec2_instances(ec2_client)
```
jericht, line 233: This was cleaning up my worker to iterate faster on my failing tests

test/e2e/test_job_attachments.py
jericht: That's fine, it's for tests. We can dismiss it. The expected owner will change depending on the account the tests are run in.

test/e2e/test_job_attachments.py
jericht: We do in Job Attachments, but serialization to dict/JSON is not built in. Need to manually do it anyway IIRC.

Plus, I think it's good to test with our public interface here rather than our implementation classes

test/e2e/test_job_attachments.py
jericht: Not sure 🤷

jericht: Looks like this is for compatibility with files created in Microsoft's Notepad: https://docs.python.org/3/library/codecs.html

> To increase the reliability with which a UTF-8 encoding can be detected, Microsoft invented a variant of UTF-8 (that Python calls "utf-8-sig") for its Notepad program: Before any of the Unicode characters is written to the file, a UTF-8 encoded BOM (which looks like this as a byte sequence: 0xef, 0xbb, 0xbf) is written.

test/e2e/test_job_attachments.py
```diff
@@ -0,0 +1,270 @@
...
+                                "args": ["{{ Task.File.runScript }}"],
+                            },
+                        },
+                        "embeddedFiles": [
+                            {
+                                "name": "runScript",
+                                "type": "TEXT",
+                                "runnable": True,
+                                "data": rf"""
+import json
```
leongdl, line 188: nit, refactor after only. Anyway we can make this script an asset? It is hard to maintain embedded like this.

test/e2e/test_job_attachments.py
```diff
@@ -0,0 +1,270 @@
...
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "path": self.path,
+            "hash": self.hash,
+            "size": self.size,
+            "mtime": self.mtime,
+        }
+
+
+class TestJobAttachments:
```
leongdl, line 51: Looks like largely refactoring test code from the file below.

leongdl (dismissed):

test/e2e/test_job_attachments.py
jericht: It's a good idea, but I think for now let's keep it here. It has OpenJD symbols in it, so it would be weird to save it as a standalone `.py` file in its current state. I'd want to refactor it to be a proper script that accepts CLI or env var arguments for all the files in needs filled in

jericht: I want to move all the E2E tests that specifically target Job Attachments  functionality into this file in a future PR, maybe that's a good time to do the cleanup in follow-up PR to that.

leongdl: sounds good, follow up only. Lets get the tests in first.

scripts/get_e2e_test_ids_from_cfn.sh
```diff
@@ -81,4 +81,7 @@ export FLEET_ID=$(echo "${STACK_OUTPUTS}" | jq -r ".${OS}ManualFleetX86Id")
 export SCALING_QUEUE_ID=$(echo "${STACK_OUTPUTS}" | jq -r '.ScalingQueueId')
 export SCALING_FLEET_ID=$(echo "${STACK_OUTPUTS}" | jq -r ".${OS}AutoFleetX86Id")
 export JOB_STORAGE_PROFILE_ID=$(echo "${STACK_OUTPUTS}" | jq -r '.LinuxJobStorageProfileId')
+export WINDOWS_JOB_STORAGE_PROFILE_ID=$(echo "${STACK_OUTPUTS}" | jq -r '.WindowsJobStorageProfileId')
+export FLEET_STORAGE_PROFILE_ID=$(echo "${STACK_OUTPUTS}" | jq -r '.LinuxFleetStorageProfileId')
```
godobyte, line 85: nit - could be more explicit as `LINUX_FLEET_STORAGE_PROFILE_ID`

test/e2e/conftest.py
```diff
@@ -99,7 +102,10 @@ def deadline_resources() -> Generator[DeadlineResources, None, None]:
         FLEET_ID: ID of a non scaling Deadline fleet to use for tests.
         SCALING_QUEUE_ID: ID of the Deadline scaling queue to use.
         SCALING_FLEET_ID: ID of the Deadline scaling fleet to use.
-        JOB_STORAGE_PROFILE_ID: ID of the Deadline storage profile to use
+        JOB_STORAGE_PROFILE_ID: ID of the Deadline storage profile to use for Linux jobs
```
godobyte, line 105: I see so the ones not prefixed with `windows` are for linux.

test/e2e/test_job_attachments.py
```diff
@@ -0,0 +1,270 @@
...
+    @pytest.mark.usefixtures("session_worker")
+    @pytest.mark.parametrize(
+        "submission_os",
+        [
+            "linux",
+            "windows",
+        ],
+    )
+    def test_worker_job_attachment_storage_profile_path_mapping(
```
godobyte, line 60: Are you planing to merge this after our infrastructure is ready, or mark it as xfail for now?

scripts/get_e2e_test_ids_from_cfn.sh
jericht: We can do a cleanup pass later

test/e2e/test_job_attachments.py
jericht: Let's mark as xfail for now. I'll prep another PR once the infrastructure is in place.

godobyte (dismissed):

moorec-aws (approved):

godobyte (approved):

PR #779, opened 2025-10-22
fix(experimental): not reporting manifests in progress updates

ryyakobe:

The WorkerAgent was incorrectly including the `manifests` field in UpdateWorkerSchedule API when reporting progress updates. This occurred because the code automatically initialized `manifests = []` for all RunStepTaskAction instances. Manifest reporting code is hidden under `MANIFEST_REPORTING_FEATURE` feature flag, so the issue is not customer-facing.

1. Removed automatic manifest initialization (`manifests = []`) for all RunStepTaskAction instances in `_handle_action_update`
2. Changed manifest initialization in attachment upload completion from `manifests_list = None` to `manifests_list = []` to ensure empty list is reported for successful tasks without job attachments
3. Added proper conditional logic to only include manifests when:
   - `MANIFEST_REPORTING_FEATURE` is enabled
   - `completed_status == "SUCCEEDED"`
   - Action is a `RunStepTaskAction`

The WorkerAgent now correctly differentiates between old workers (never report manifests) and new workers (always report empty list for successful task runs, even if no outputs). Progress updates and failed actions no longer incorrectly include the manifests field in API requests.

Added three new unit tests:
- `test_progress_update_excludes_manifests` - Verifies progress updates don't include manifests
- `test_failed_task_excludes_manifests` - Verifies failed tasks don't include manifests  
- `test_successful_task_without_job_attachments_includes_empty_manifests` - Verifies successful tasks without attachments include empty list

Updated existing test expectations to expect `manifests=None` for failed actions instead of `manifests=[]`.

No user-facing documentation changes required. This is an internal bug fix in the API request construction. Manifest reporting code is still under feature flag, so this PR doesn't introduce any no customer-facing changes.

No. This fix ensures the worker agent correctly implements the intended behavior of the manifest reporting feature.

### DIFF
--- a/test/unit/sessions/test_session.py
+++ b/test/unit/sessions/test_session.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import os
 from collections.abc import Generator, Iterable
 from concurrent.futures import wait
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from pathlib import Path, PurePosixPath, PureWindowsPath
 from threading import Event, RLock
 from types import TracebackType
@@ -60,6 +60,9 @@ from deadline_worker_agent.sessions.actions import (
     EnterEnvironmentAction,
     ExitEnvironmentAction,
     RunStepTaskAction,
+)
+from deadline_worker_agent.sessions.actions.run_attachment_upload import (
+    AttachmentUploadAction as AttachmentUploadActionClass,
 )
 from deadline_worker_agent.sessions.job_entities import (
     EnvironmentDetails,
@@ -1437,7 +1440,7 @@ class TestSessionActionUpdatedImpl:
                 start_time=action_start_time,
                 completed_status="FAILED",
                 end_time=action_complete_time,
-                manifests=[],
+                manifests=None,
             )
         else:
             expected_action_update = SessionActionStatus(
@@ -1516,7 +1519,7 @@ class TestSessionActionUpdatedImpl:
                 start_time=action_start_time,
                 completed_status="FAILED",
                 end_time=action_complete_time,
-                manifests=[],
+                manifests=None,
             )
         else:
             expected_action_update = SessionActionStatus(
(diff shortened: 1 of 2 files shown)


test/unit/sessions/test_session.py
```diff
@@ -2180,6 +2183,203 @@ def test_action_output_capture_filter_ja_upload_callback_valid(
         assert len(session._upload_manifest_list) == 1
         assert all(isinstance(item, UploadManifestInfo) for item in session._upload_manifest_list)
 
+    @pytest.mark.skipif(
+        not MANIFEST_REPORTING_FEATURE or not ASSET_SYNC_JOB_USER_FEATURE,
+        reason="This test only runs when both MANIFEST_REPORTING_FEATURE and ASSET_SYNC_JOB_USER_FEATURE are enabled",
+    )
```
jusiskin, line 2189: Rather than skipping the tests, can we mock the feature flags as part of the test instead?

Otherwise we have to run the test suite with the feature flags turned on to get CI/CD coverage.

After looking at earlier tests in this file, it seems there is precedent for tests being conditional on the feature flag. Unfortunately, right now it means our CI/CD is not running tests for the feature flags.

Open to discuss breaking this out to a future PR, but what are your thoughts?

test/unit/sessions/test_session.py
```diff
@@ -2180,6 +2183,203 @@ def test_action_output_capture_filter_ja_upload_callback_valid(
...
+        # Successful action status for the upload action
+        success_action_status = ActionStatus(state=ActionState.SUCCESS, exit_code=0)
+
+        # WHEN - Upload action completes successfully
+        with (
+            patch.object(
+                session_mod,
+                "OPENJD_ACTION_STATE_TO_DEADLINE_COMPLETED_STATUS",
+                {ActionState.SUCCESS: "SUCCEEDED"},
+            ),
```
jusiskin, line 2369: nit: why do we need to mock this attribute?

jusiskin (approved):
Looks good. Just an optional request to change the testing approach of feature flags in unit tests.

erico-aws (approved):

PR #780, opened 2025-10-22
fix(experimental): attachment uploads all as output when no output relative directories defined for a root

godobyte:
`ouputRelativeDirectories` is an optional field in `ManifestProperties` as part of `CreateJob` API model. If not provided upon calling CreateJob, the value will be None. Old Job Attachment code path identify output file under the output dir by iterating through ouputRelativeDirectories. When the list is None or empty, nothing gets recognize as upload for upload. This is the correct intended behaviour.

The new code path for `ASSET_SYNC_JOB_USER_FEATURE` treated the empty and None as uploading all, which is incorrect.

Skip snapshot and upload when ouputRelativeDirectories is not specified.

Correctly upload for specified output directories.

1. Added and ran unit tests
2. Ran E2E tests

```
attachments:
  manifests:
  - rootPath: /test/assets1
    rootPathFormat: posix
    inputManifestPath: manifest1hash
    inputManifestHash: manifest1hash
  - rootPath: /test/assets2
    rootPathFormat: posix
    outputRelativeDirectories: []
    inputManifestPath: manifest1hash
    inputManifestHash: manifest1hash
  fileSystem: COPIED
```

Logs
```
2025/10/22 13:51:36-07:00 Output:2025/10/22 13:51:37-07:00 No output directories specified for /test/assets1, skipping upload
2025/10/22 13:51:37-07:00 No output directories specified for /test/assets2, skipping upload
```

N/A

No
----

### DIFF
--- a/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
+++ b/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
@@ -158,6 +158,7 @@ def snapshot(
         local_root_path = manifest_props.local_root_path
         output_relative_directories = manifest_props.local_output_relative_directories()
 
+        # Skip when output relative directories is None or []
         if not output_relative_directories:
             print(
                 f"No output directories specified for {manifest_props.root_path}, skipping upload"
@@ -173,6 +174,7 @@ def snapshot(
             # if the base is None, meaning all changes are output
             diff=root_path_to_base_manifest.get(manifest_props.root_path),
             # include patterns for output directories (with recursive wildcard)
+            # when the code reaches here, it's guaranteed output_relative_directories contains value
             include=[subdir + "/**" for subdir in output_relative_directories],
             name="output",
         )
--- a/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
+++ b/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
@@ -156,7 +156,13 @@ def snapshot(
     # Process each worker manifest property
     for manifest_props in worker_manifest_properties:
         local_root_path = manifest_props.local_root_path
-        output_relative_directories = manifest_props.local_output_relative_directories() or []
+        output_relative_directories = manifest_props.local_output_relative_directories()
+
+        if not output_relative_directories:
+            print(
+                f"No output directories specified for {manifest_props.root_path}, skipping upload"
+            )
+            continue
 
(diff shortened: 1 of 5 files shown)


src/deadline_worker_agent/sessions/attachment_models.py
```diff
@@ -88,7 +88,7 @@ def local_output_relative_directories(self) -> Optional[List[str]]:
                                or None if no output directories are defined
         """
 
-        if not self.output_relative_directories:
+        if self.output_relative_directories is None:
```
leongdl, line 91: [] and false both evaluates to NOT right? So why explicitly None now? Any comments why? 

If its none, should it be Optional instead? Typing and code flow would matter a lot.

jericht (dismissed):

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -156,7 +156,13 @@ def snapshot(
     # Process each worker manifest property
     for manifest_props in worker_manifest_properties:
         local_root_path = manifest_props.local_root_path
-        output_relative_directories = manifest_props.local_output_relative_directories() or []
+        output_relative_directories = manifest_props.local_output_relative_directories()
+
+        if not output_relative_directories:
```
leongdl, line 162: But we use `not` here, while below we use `None`.

src/deadline_worker_agent/sessions/attachment_models.py
godobyte: The function already returns `Optional`. The previous code would return None when `output_relative_directories` is None and `[]`, which is not clear for downstream to differentiate if needed.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
godobyte: Yea it's expected behaviour to skip when `output_relative_directories` is both `None` and `[]`. The code change below is not required, I kept it since it's an improvement.

leongdl: Got it, maybe add a comment? Otherwise we'll have to think about it again. What are the cases to think about?

AlexTranAmz (dismissed):

leongdl (approved):

AlexTranAmz (approved):

jericht (approved):

PR #790, opened 2025-10-27
test(experimental): parametrize ASSET_SYNC_JOB_USER_FEATURE for e2e tests

godobyte:
We want to guard the experimental asset sync run as job user feature via E2E tests, making sure there is no regression.

Parametrize ASSET_SYNC_JOB_USER_FEATURE for e2e tests that submits jobs

- Add asset_sync_worker_config fixture with parametrized feature flag
- Add asset_sync_session_worker fixture using parametrized config, ensure feature flag is set before worker initialization
- Replace all session_worker usage with asset_sync_session_worker for job submission tests
- Remove hardcoded ASSET_SYNC_JOB_USER_FEATURE from base worker_config

All tests that involve job submission will be run as part of CI/CD to catch regressions.

#### Linux
```
261.28s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_success[False]
260.52s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_enters_stopping_state_while_draining
214.51s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_success[True]
150.92s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_dep_data_flow_linux[True-COPIED]
136.57s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_reports_canceled_sync_input_actions_as_canceled[False]
============ 61 passed, 6 skipped, 1 warning in 3829.93s (1:03:49) =============

```

#### Windows
```
=========================== short test summary info ============================
FAILED test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only[True]
======= 1 failed, 52 passed, 14 skipped, 1 warning in 3310.99s (0:55:10) =======


351.73s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only[False]
326.70s setup    test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only[True]
45.38s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only[True]
23.57s call     test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only[False]
6.59s teardown test/e2e/test_job_submissions.py::TestJobSubmission::test_worker_job_attachments_output_only[False]
=================== 2 passed, 1 warning in 755.73s (0:12:35) ===================


378.95s setup    test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_job_attachment_storage_profile_path_mapping[True-linux]
367.49s setup    test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_job_attachment_storage_profile_path_mapping[False-linux]
33.48s call     test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_job_attachment_storage_profile_path_mapping[True-windows]
23.39s call     test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_job_attachment_storage_profile_path_mapping[True-linux]
23.29s call     test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_job_attachment_storage_profile_path_mapping[False-linux]
=================== 4 passed, 1 warning in 857.24s (0:14:17) ===================
```

N/A

No
----

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -9,7 +9,6 @@ import pytest
 from dataclasses import dataclass, field, InitVar
 from typing import Callable, Generator, Type
 from contextlib import contextmanager
-from typing import Dict
 
 from deadline_test_fixtures import (
     DeadlineWorker,
@@ -25,7 +24,6 @@ from deadline_test_fixtures import (
     Ec2Tag,
 )
 import pytest
-from deadline_worker_agent.feature_flag import ASSET_SYNC_JOB_USER_FEATURE
 
 LOG = logging.getLogger(__name__)
 
@@ -192,9 +190,50 @@ def worker_config(
     Returns:
         DeadlineWorkerConfiguration: Configuration for use by DeadlineWorker.
     """
-    worker_env_var: Dict[str, str] = {
-        "ASSET_SYNC_JOB_USER_FEATURE": str(ASSET_SYNC_JOB_USER_FEATURE),
+
+    return dataclasses.replace(
+        worker_config,
+        job_users=[
+            posix_job_user,
+            posix_config_override_job_user,
+            posix_env_override_job_user,
+        ],
+        windows_job_users=windows_job_users,
+    )
+
+
+@pytest.fixture(scope="session")
+def session_worker(
+    request: pytest.FixtureRequest,
+    worker_config: DeadlineWorkerConfiguration,
+    ec2_worker_type: Type[EC2InstanceWorker],
+) -> Generator[DeadlineWorker, None, None]:
+    with create_worker(worker_config, ec2_worker_type, request) as worker:
+        yield worker
+
+    stop_worker(request, worker)
+
+
+@pytest.fixture(scope="session", params=[True, False])
+def asset_sync_worker_config(
+    request: pytest.FixtureRequest,
+    posix_job_user: PosixSessionUser,
+    posix_env_override_job_user: PosixSessionUser,
+    posix_config_override_job_user: PosixSessionUser,
+    worker_config: DeadlineWorkerConfiguration,
+    windows_job_users: list[str],
+) -> DeadlineWorkerConfiguration:
+    """
+    Parametrized worker configuration fixture that tests ASSET_SYNC_JOB_USER_FEATURE with both True and False values.
+
+    This fixture ensures the environment variable is set before worker initialization,
+    allowing proper testing of the asset sync feature flag behavior.
+    """
+    asset_sync_feature = request.param
+    worker_env_var = {
+        "ASSET_SYNC_JOB_USER_FEATURE": str(asset_sync_feature),
     }
+    LOG.info(f"worker_env_var: {worker_env_var}")
 
     return dataclasses.replace(
         worker_config,
@@ -209,12 +248,12 @@ def worker_config(
 
 
 @pytest.fixture(scope="session")
-def session_worker(
+def asset_sync_session_worker(
     request: pytest.FixtureRequest,
-    worker_config: DeadlineWorkerConfiguration,
+    asset_sync_worker_config: DeadlineWorkerConfiguration,
     ec2_worker_type: Type[EC2InstanceWorker],
 ) -> Generator[DeadlineWorker, None, None]:
-    with create_worker(worker_config, ec2_worker_type, request) as worker:
+    with create_worker(asset_sync_worker_config, ec2_worker_type, request) as worker:
         yield worker
 
     stop_worker(request, worker)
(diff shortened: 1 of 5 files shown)


test/e2e/conftest.py
```diff
@@ -192,9 +190,6 @@ def worker_config(
     Returns:
         DeadlineWorkerConfiguration: Configuration for use by DeadlineWorker.
     """
-    worker_env_var: Dict[str, str] = {
-        "ASSET_SYNC_JOB_USER_FEATURE": str(ASSET_SYNC_JOB_USER_FEATURE),
-    }
```
AWS-Samuel, line 197: it might be worth keeping this as an empty dictionary to make it clear/easy for future changes to modify the environment

test/e2e/conftest.py
```diff
@@ -220,6 +214,51 @@ def session_worker(
...
+@pytest.fixture(scope="session")
+def asset_sync_session_worker(
+    request: pytest.FixtureRequest,
+    asset_sync_worker_config: DeadlineWorkerConfiguration,
+    ec2_worker_type: Type[EC2InstanceWorker],
+) -> Generator[DeadlineWorker, None, None]:
+    with create_worker(asset_sync_worker_config, ec2_worker_type, request) as worker:
+        yield worker
+
+    stop_worker(request, worker)
```
AWS-Samuel, line 262: Are there any other workers in the queue/fleet that this worker picks work from? Just checking that there won't be a chance of another worker without the asset sync configuration picking up a job meant for this one.

test/e2e/test_job_submissions.py
jericht, line 1: I wonder if there's a cleaner way we can do this without explicitly requesting a different fixture for these tests. Not a huge deal, but when we eventually turn on this feature permanently, we will want to get rid of this new fixture otherwise we are just duplicating tests

test/e2e/conftest.py
```diff
@@ -434,7 +473,10 @@ def pytest_collection_modifyitems(items):
     sorted_list = list(items)
     for item in items:
         # Run session scoped tests last to prevent Worker conflicts with class and function scoped tests.
-        if "session_worker" in item.fixturenames:
+        if (
+            "session_worker" in item.fixturenames
+            or "asset_sync_session_worker" in item.fixturenames
+        ):
```
jusiskin: There may be further code modifications needed here. The intention behind this `pytest_collection_modifyitems` pytest hook is to sort tests in an optimal way so that we only ever have a single worker online for the fleet to guarantee that the correct worker is picking up the relevant job for the test.

We would want the `session_worker` tests to be grouped together in the sort order and the `asset_sync_session_worker` tests to be grouped together. Currently, the code will the tests using either of the fixtures together which could cause multiple workers to be online and contend for jobs.

jusiskin (changes requested):

test/e2e/test_job_submissions.py
godobyte: Another way I can think of is to modify the `session_worker` fixture to check test class name, which brings hidden logic. I'd prefer keeping the current approach with separate asset_sync_session_worker fixture. It's more explicit, maintainable, and follows pytest best practices of making dependencies clear in test signatures.

test/e2e/conftest.py
godobyte: I believe this is related to @jusiskin's comment below, I'll modify `pytest_collection_modifyitems` to ensure sequential execution of the different session worker test groups.

test/e2e/conftest.py
godobyte: Thanks for calling it out, I'll address this.

test/e2e/conftest.py
godobyte: Sure I was actually thinking along that line too, it'll be required when we need to remove the `asset_sync_worker_config` along with the feature flag. I'll add it now.

jericht (dismissed):

test/e2e/conftest.py
```diff
@@ -471,13 +474,21 @@ def operating_system() -> OperatingSystem:
...
-            or "asset_sync_session_worker" in item.fixturenames
-        ):
+        # Separate session worker tests into two groups to prevent worker conflicts
+        if "asset_sync_session_worker" in item.fixturenames:
             sorted_list.remove(item)
-            sorted_list.append(item)
+            asset_sync_session_worker_tests.append(item)
+        elif "session_worker" in item.fixturenames:
+            sorted_list.remove(item)
+            session_worker_tests.append(item)
```
AWS-Samuel: can we raise an error here if for some reason a test requests both fixtures?

godobyte: Good catch! I don't think this fixture will live for too long, but it's always good to code defensive and fail fast. Addressed in the rev.

AWS-Samuel (approved):

jusiskin (approved):
LGTM.

PR #818, opened 2025-11-14
test: parameterize job names with asset sync configuration

godobyte:
E2E test jobs were using generic, hard-coded names that made it difficult to identify which specific test case or configuration was being tested when looking at job execution logs or debugging test failures. This lack of descriptive naming hindered troubleshooting and test result analysis.

- Added dynamic job naming that includes the asset sync feature flag value (e.g., "True_StorageProfilePathMappingJob" or "False_AppendStringJob")
- Extracted asset sync feature configuration into a reusable pattern across all test methods
- Made job template names parameterizable by adding JobName parameter definitions

### What is the impact of this change
- Better visibility and debuggability via DCM.
- Help to gather job data for analysis.

n/a

No

----

### DIFF
--- a/test/e2e/test_job_attachments.py
+++ b/test/e2e/test_job_attachments.py
@@ -67,6 +67,7 @@ class Asset:
 
 class TestJobAttachments:
     JOB_OUTPUT_PATH = os.path.join(os.getcwd(), "job_output")
+    ASSET_SYNC_JOB_USER_FEATURE = "ASSET_SYNC_JOB_USER_FEATURE"
 
     @pytest.mark.usefixtures("asset_sync_class_worker")
     @pytest.mark.parametrize(
@@ -81,6 +82,7 @@ class TestJobAttachments:
         submission_os: str,
         deadline_resources: DeadlineResources,
         deadline_client: DeadlineClient,
+        asset_sync_worker_config: DeadlineWorkerConfiguration,
         tmp_path: pathlib.Path,
     ) -> None:
         # Test that submits a job that has storage profile and confirm that the final output path and content is as we expect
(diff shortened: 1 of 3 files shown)


test/e2e/test_job_attachments.py
```diff
@@ -169,10 +171,17 @@ def test_worker_job_attachment_storage_profile_path_mapping(
             "inputManifestHash": test_disambiguator,
         }
 
+        asset_sync_feature = (
+            asset_sync_worker_config.worker_env_var.get(self.ASSET_SYNC_JOB_USER_FEATURE, "False")
+            if asset_sync_worker_config.worker_env_var
+            else "False"
+        )
+        job_name = f"{asset_sync_feature}_StorageProfilePathMappingJob"
```
jusiskin: How about using the f-string expression syntax ([docs](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals) here? The docs say:

>   To display both the expression text and its value after evaluation, (useful in debugging), an equal sign '=' may be added after the expression.

So if we put

```suggestion
        job_name = f"StorageProfilePathMappingJob[{asset_sync_feature=}]"
```

It should produce job names of the form:

*   `StorageProfilePathMappingJob[asset_sync_feature=True]`
*   `StorageProfilePathMappingJob[asset_sync_feature=False]`

This will have the benefit of making searching for jobs by prefix easier.

Same suggestion applies to the job name format changes below&hellip;

leongdl (dismissed):
Thank you - this is a good debugging change!

jusiskin (changes requested):
Great ideas for improving debugging and diagnosis. I have one suggestion to improve the job names slightly

test/e2e/test_job_attachments.py
godobyte: I chose not to do that from DCM user experience perspective. Adding that to the end meaning
1. Most of the time the information would be hidden unless expanding job name column
2. If there is a switch from other view, the column width is immediately reset
3. Having it in the beginning looks more explicit since the value is short
4. Job names' length varies, having it in the end may not be immediately obvious for comparison

Since it's a short-lived feature flag and the only feature flag we parameterize, I prefer keeping it as-is unless you have strong opinions against it.

jusiskin: My thinking was that using search will be more common than reading job names in the table. Do you anticipate that developers would read the names in the table more often than using search?

jusiskin: Also, when reading job names in the job table, the feature flag is more secondary than the other information encoded in the name of the job

godobyte: I don't see how this format is not suitable for searching, only jobs run with this feature has the value and follow the pattern. 

This is a 2-way door, we can let it deploy and gather feedback before next iteration. Personally I don't like always need to resize the column width for job name, but I can disagree and commit to get this change through.

jusiskin (approved):

leongdl (approved):

PR #822, opened 2025-11-18
test: ignore non-cancellable jobs in cleanup

edwards-aws:
Jobs in the test queue can sometimes get in a state where they can't be cancelled. This causes the tests to fail to start as the cleanup script crashes and the tests don't run.

Ignore any error when trying to cancel the job. This script is a best effort and there's no reason for it to block the tests running.

Tests can run again if jobs are in a bad state

I ran the script against a farm/queue with `CREATE_FAILED` jobs and the script succeeded.

N/A

No

----

### DIFF
--- a/test/e2e/pre_test_cleanup.py
+++ b/test/e2e/pre_test_cleanup.py
@@ -13,22 +13,24 @@ This script will automatically run before the tests, if the tests are run with
 hatch run e2e-test
 """
 
-from typing import Any
-import time
-import boto3
-import boto3.session
-
 import logging
 import os
 import sys
+import time
 from dataclasses import dataclass
+from typing import Any
+
+import boto3
+import boto3.session
 from botocore.client import BaseClient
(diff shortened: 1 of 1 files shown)


test/e2e/pre_test_cleanup.py
```diff
@@ -13,22 +13,24 @@
...
+import boto3
+import boto3.session
 from botocore.client import BaseClient
-from botocore.exceptions import ClientError
 from botocore.config import Config
+from botocore.exceptions import ClientError
+from deadline_test_fixtures.deadline.resources import (
+    COMPLETE_TASK_STATUSES,
+    TaskStatus,
+)
```
edwards-aws, line 31: My LSP decided to organize the imports.

epmog (dismissed):

epmog (approved):

mwiebe (approved):

PR #828, opened 2025-11-22
fix(experimental): escape special characters in output directory patterns when generate diff for upload

godobyte:
The caller for manifest snapshot in attachment_upload.py was not properly escaping these  special characters when creating include patterns for file detection, causing the glob pattern matching to fail or match unintended files.

Added glob.escape() function to properly escape glob special characters in output directory names before creating include patterns for calling _manifest_snapshot.

Output directories with special characters in their names are now properly detected for upload.

1. Added unit test
2. Updated E2E test to cover the case
3. Manually submitted a test job that failed previously and passed with the change

N/A

No

----


leongdl (approved):
Darn, glob.escape.

test/e2e/job_attachment_bundle/dep_data_flow/linux_bundle/template.yaml
```diff
@@ -17,7 +17,7 @@ parameterDefinitions:
   type: PATH
   objectType: DIRECTORY
   dataFlow: INOUT
-  default: ./data_dir2
+  default: "./[data_]dir2"
```
godobyte, line 20: Verified `./[data_]*dir2?` works as expected but github CI would fail on Windows.

https://github.com/aws-deadline/deadline-cloud-worker-agent/actions/runs/19588337219?pr=828

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -176,7 +177,7 @@ def snapshot(
             diff=root_path_to_base_manifest.get(manifest_props.root_path),
             # include patterns for output directories (with recursive wildcard)
             # when the code reaches here, it's guaranteed output_relative_directories contains value
-            include=[subdir + "/**" for subdir in output_relative_directories],
+            include=[glob.escape(subdir) + "/**" for subdir in output_relative_directories],
```
leongdl, line 180: Can we please add this comment to the Attachments API so users know the input include should be blob escaped.

leongdl: Side follow up to improve our APIs

godobyte: Sure, that would help other manifest API/CLI users. I can take that separately.

test/e2e/job_attachment_bundle/dep_data_flow/linux_bundle/template.yaml
jusiskin: Thanks for manually testing this!

nit: I have an idea to work-around the git checkout issue. This suggestion can be a follow-up if preferred. I see error is happening on git checkout and git complains that Windows does not allow `*` in the filename.

What if the test code conditionally creates the directory if not on Windows?

jusiskin (approved):
Great work fixing this bug!

test/e2e/job_attachment_bundle/dep_data_flow/linux_bundle/template.yaml
godobyte: Yea that certainly can be done as a follow-up. I actually had some in-progress code for this approach, and decided to get the fix PR out first as it's more urgent.

PR #843, opened 2025-12-10
feat: add telemetry event on what filesystem is used

npmacl:

we would like to have stats on the usage of COPIED vs VIRTUAL for filesystem

added a telemetry event when we are syncing the filesystem that emits whether the session is using COPIED vs VIRTUAL

we will have data on usage of COPIED vs VIRTUAL

tested submitting jobs using both COPIED and VIRTUAL  to an SMF that used this change via host configuration script. Verified that I saw the expected telemetry events being created in our backend.

no
----

### DIFF
--- a/src/deadline_worker_agent/aws/deadline/__init__.py
+++ b/src/deadline_worker_agent/aws/deadline/__init__.py
@@ -835,7 +835,7 @@ def record_uncaught_exception_telemetry_event(exception_type: str) -> None:
     )
 
 
-def record_asset_sync_filesystem_event(queue_id: str, file_system: str) -> None:
+def _record_asset_sync_filesystem_event(queue_id: str, file_system: str) -> None:
     """Calls the telemetry client to record what filesystem was used"""
     details: Dict[str, Any] = {"queue_id": queue_id, "filesystem": file_system}
     _get_deadline_telemetry_client().record_event(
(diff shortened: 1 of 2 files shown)


src/deadline_worker_agent/aws/deadline/__init__.py
```diff
@@ -835,6 +835,15 @@ def record_uncaught_exception_telemetry_event(exception_type: str) -> None:
     )
 
 
+def record_asset_sync_filesystem_event(queue_id: str, file_system: str) -> None:
```
leongdl: This should be an internal function, please prefix it with `_`

leongdl (dismissed):

src/deadline_worker_agent/aws/deadline/__init__.py
```diff
@@ -835,6 +835,15 @@ def record_uncaught_exception_telemetry_event(exception_type: str) -> None:
     )
 
 
+def _record_asset_sync_filesystem_event(queue_id: str, file_system: str) -> None:
+    """Calls the telemetry client to record what filesystem was used"""
+    details: Dict[str, Any] = {"queue_id": queue_id, "filesystem": file_system}
+    _get_deadline_telemetry_client().record_event(
+        event_type="com.amazon.rum.deadline.worker_agent.sync_inputs_filesystem",
```
godobyte:
```suggestion
        event_type="com.amazon.rum.deadline.worker_agent.attachment_download_filesystem",
```

What about attachment_download_filesystem to match other attachment download events? `sync_inputs` is the term for the deprecating code path.

leongdl (approved):

godobyte (approved):

epmog (approved):

PR #855, opened 2026-01-06
feat: report output job attachment manifest locations to the service

ryyakobe:

The MANIFEST_REPORTING_FEATURE flag should now be enabled by default. We don't have to set MANIFEST_REPORTING_FEATURE flag to True but can instead remove all the guardrails related to it. The functionality should now be always available (when another flag ASSET_SYNC_JOB_USER_FEATURE is also enabled). 

Removed the MANIFEST_REPORTING_FEATURE flag from feature_flag.py and cleaned up all related code paths in the scheduler, session management, and attachment upload functionality.

- Manifest reporting is now enabled by default
- Simplified codebase by removing 226 lines of unused code
- Eliminated feature flag complexity around manifest reporting
- Reduced maintenance burden by removing dead code paths
- Cleaned up test suite by removing tests for removed functionality

Existing test suite should continue to pass with the removal of the feature flag code and its associated tests.

No documentation changes required as this removes internal feature flag code that was not part of the public API.

No, this is not a breaking change as it removes internal feature flag code that was not part of the public interface.

----

### DIFF
--- a/src/deadline_worker_agent/feature_flag.py
+++ b/src/deadline_worker_agent/feature_flag.py
@@ -7,6 +7,3 @@ import os
 ASSET_SYNC_JOB_USER_FEATURE = (
     os.environ.get("ASSET_SYNC_JOB_USER_FEATURE", "false").lower() == "true"
 )
-
-# Feature flag for output manifest reporting
-MANIFEST_REPORTING_FEATURE = os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
--- a/src/deadline_worker_agent/scheduler/scheduler.py
+++ b/src/deadline_worker_agent/scheduler/scheduler.py
@@ -69,7 +69,6 @@ from ..log_messages import (
     WorkerLogEvent,
     WorkerLogEventOp,
 )
-from ..feature_flag import MANIFEST_REPORTING_FEATURE
 from .log import LOGGER
 from .session_cleanup import SessionUserCleanupManager
 from .session_queue import SessionActionQueue, SessionActionStatus
@@ -609,8 +608,8 @@ class WorkerScheduler:
         if action_updated.end_time:
             updated_action["endedAt"] = action_updated.end_time
 
-        # Add manifest information if available and feature flag is enabled
-        if MANIFEST_REPORTING_FEATURE and action_updated.manifests is not None:
+        # Add manifest information if available
+        if action_updated.manifests is not None:
             updated_action["manifests"] = action_updated.manifests
 
         # Truncate message to max bytes allowed by UpdateWorkerSchedule API
--- a/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
+++ b/src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
@@ -366,14 +366,9 @@ def upload_output_assets(
     return output_manifest_info_list
 
 
-def is_manifest_reporting_enabled() -> bool:
-    """Check if manifest reporting feature is enabled."""
-    return os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
-
-
 def should_use_task_chunking_format() -> bool:
     """Determine if task chunking format (without task_id) should be used."""
-    return is_manifest_reporting_enabled() and not os.environ.get("DEADLINE_TASK_ID")
+    return not os.environ.get("DEADLINE_TASK_ID")
 
 
 def build_s3_manifest_path() -> str:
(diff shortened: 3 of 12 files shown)


src/deadline_worker_agent/scheduler/scheduler.py
```diff
@@ -609,8 +608,8 @@ def _exit_code_to_32bit_signed(exitcode: int) -> int:
         if action_updated.end_time:
             updated_action["endedAt"] = action_updated.end_time
 
-        # Add manifest information if available and feature flag is enabled
-        if MANIFEST_REPORTING_FEATURE and action_updated.manifests is not None:
+        # Add manifest information if available
+        if action_updated.manifests is not None:
             updated_action["manifests"] = action_updated.manifests
```
jusiskin, line 613: The PR description says:

>   The MANIFEST_REPORTING_FEATURE code was no longer needed and was cluttering the codebase with **unused feature flag logic** and associated test code.

Looking at this logic, it seems to change the logic to work as-if the feature flag was turned on by default.

Am I following this correctly?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -366,14 +366,9 @@ def upload_output_assets(
...
 
-def is_manifest_reporting_enabled() -> bool:
-    """Check if manifest reporting feature is enabled."""
-    return os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
-
-
 def should_use_task_chunking_format() -> bool:
     """Determine if task chunking format (without task_id) should be used."""
-    return is_manifest_reporting_enabled() and not os.environ.get("DEADLINE_TASK_ID")
+    return not os.environ.get("DEADLINE_TASK_ID")
```
jusiskin, line 365: Same feedback as before. It seems like we are changing the logic to behave as-if the feature flag were always enabled.

src/deadline_worker_agent/scheduler/scheduler.py
ryyakobe: Updated the PR description. Let me know if it's clearer now.

 We do not need this feature flag anymore, as we want the manifest reporting functionality to be always enabled.

jusiskin: Okay that makes more sense. Can we also update the PR title to reflect the change more accurately? This should probably be a `feat:` since it was previously an opt-in feature flag that most customers would not have used.

ryyakobe: Changed to `feat: enable output manifest reporting by default`. Let me know if that works.

src/deadline_worker_agent/feature_flag.py
```diff
@@ -7,6 +7,3 @@
 ASSET_SYNC_JOB_USER_FEATURE = (
     os.environ.get("ASSET_SYNC_JOB_USER_FEATURE", "false").lower() == "true"
 )
-
-# Feature flag for output manifest reporting
-MANIFEST_REPORTING_FEATURE = os.environ.get("MANIFEST_REPORTING_FEATURE", "false").lower() == "true"
```
godobyte, line 12: I noticed the PR includes unit tests to validate the feature, which is a great start. Do we have a plan for E2E testing of manifest reporting? This would help catch any integration issues that unit tests might miss.

ryyakobe: There's no plan yet to add those tests in this codebase. I agree this is a great idea and we could add them as a follow up. We do have some E2E tests as part of the service.

godobyte: Could we confirm the existing service E2E tests exercise this feature? WA rollbacks via SMF have historically been painful and slow - having representative E2E coverage before merge would help us catch integration issues early.

It could be something like adding verification to existing E2E tests by
1. verify Job response, check whether it's a chunked job
2. if it's a chunked job, verify manifest is reported, otherwise it's not reported

godobyte (dismissed):

src/deadline_worker_agent/scheduler/scheduler.py
jusiskin, line 1: One last request that we update the worker API protocol [here](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/docs/dev/worker_api_protocol.md#main-control-loop) to specify that we should be providing the output manifests when completing task runs.

jusiskin (changes requested):

jusiskin (approved):

godobyte (approved):

PR #865, opened 2026-01-13
feat: run attachment download and upload as job user, report output job attachment manifest locations to the service

godobyte:
1. The ASSET_SYNC_JOB_USER_FEATURE flag was still disabled by default and controlled via environment variable, even though the feature (running attachment download/upload as job user) has been released to Service Managed Fleets.
2. The MANIFEST_REPORTING_FEATURE code was previously reverted but needed to be re-removed since the feature is now stable.

1. Enabled ASSET_SYNC_JOB_USER_FEATURE by default (hardcoded to True) and removed the environment variable toggle. Updated comments to indicate the legacy code path is deprecated. Updated e2e test fixture asset_sync_worker_config to only test with True since the feature is now always enabled.
1. Reverted the revert of the MANIFEST_REPORTING_FEATURE removal, effectively re-removing the feature flag and its associated code paths.

- The ASSET_SYNC_JOB_USER_FEATURE environment variable is no longer respected. Attachment downloads and uploads will now always run as the job user. Reduced test matrix by removing the False parameter from asset sync tests.
- Removing the MANIFEST_REPORTING_FEATURE flag and related conditional logic.

- hatch run test
- e2e tests `hatch run pytest -- test/e2e/test_job_attachments.py`

```
235.42s setup    test/e2e/test_job_attachments.py::TestJobAttachments::test_job_submission_many_small_files_download_does_not_cancel
195.33s setup    test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_job_attachment_storage_profile_path_mapping[True-linux]
158.46s call     test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_create_job_API_call_linux[True]
122.49s call     test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_job_attachments_dep_data_flow_linux[True-VIRTUAL]
113.76s call     test/e2e/test_job_attachments.py::TestJobAttachments::test_worker_job_attachments_dep_data_flow_linux[True-COPIED]
============ 14 passed, 3 skipped, 1 warning in 1281.26s (0:21:21) =============
```

N/A

No

----

### DIFF
--- a/test/unit/sessions/actions/test_attachment_upload_path_format.py
+++ b/test/unit/sessions/actions/test_attachment_upload_path_format.py
@@ -73,7 +73,7 @@ class TestAttachmentUploadPathFormat:
     )
     @patch("deadline_worker_agent.sessions.actions.scripts.attachment_upload.decode_manifest")
     @patch("builtins.open")
-    def test_old_format_when_feature_disabled(
+    def test_old_format_when_task_id_present(
         self,
         mock_open,
         mock_decode_manifest,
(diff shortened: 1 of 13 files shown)


test/unit/sessions/actions/test_attachment_upload_path_format.py
```diff
@@ -220,7 +173,7 @@ def test_default_behavior_is_old_format(
         mock_root_path_to_output_manifest,
     ):
         """Test that old format is used by default when feature flag not set."""
-        # Arrange: No MANIFEST_REPORTING_FEATURE in environment
+        # Arrange
```
jusiskin, line 176: nit: I'm not sure this comment has any value

test/unit/sessions/actions/test_attachment_upload_path_format.py
```diff
@@ -170,9 +125,7 @@ def test_new_format_when_feature_enabled_task_id_not_present(
         mock_worker_manifest_properties,
         mock_root_path_to_output_manifest,
     ):
-        """Test that new format is used when MANIFEST_REPORTING_FEATURE=true."""
-        # Arrange
-        mock_env_vars["MANIFEST_REPORTING_FEATURE"] = "true"
+        """Test that new format is used."""
```
jusiskin, line 128: nit: can we elaborate a bit more here?

As it sits now, the comment doesn't seem to add much beyond the test function name (`test_new_format_when_feature_enabled_task_id_not_present`).

Also, the test appears to assert that a function is called, but it doesn't confirm how the return value of that function is used.

jusiskin (approved):

diane-hj (approved):
Thanks for updating it!

leongdl (approved):
Looks good on the revert and turning job asset run as user to TRUE.

test/unit/sessions/actions/test_attachment_upload_path_format.py
godobyte: Probably not, I can remove it when work around this part of code.

test/unit/sessions/actions/test_attachment_upload_path_format.py
godobyte: I believe this test is trying to validate when DEADLINE_TASK_ID env var is not present, the function calls `test_new_format_when_feature_enabled_task_id_not_present`, otherwise `partial_session_action_manifest_prefix`.

The value the function is used is via `upload_output_assets` logging, that can probably be verified if not already covered. We can look into that as a follow-up.

PR #880, opened 2026-02-18
chore(e2e): Verify Windows Powershell file replacement successful

ttblanchard:

The `test_config_file_user_override` test can fail with a misleading assertion error when the PowerShell `-replace` command doesn't find a match. The test expects `I am: config-override` but gets `I am: job-user`.

PowerShell's `-replace` operator silently returns the original string when the pattern isn't found - it doesn't throw an error or return a non-zero exit code. The existing assert `cmd_result.exit_code == 0` check only verifies the command ran, not that it actually modified anything.

Additionally, while the test fixtures use `$ErrorActionPreference = 'Stop'` for worker configuration commands (see `configure_worker_common`), this isn't applied to arbitrary `send_command` calls from tests. And even if it were, `-replace` not finding a match isn't a terminating error in PowerShell.

Add a verification step after the replacement to confirm the config was actually modified. This provides a clear, actionable error message pointing to the actual problem rather than a downstream assertion failure about unexpected user output.

This change adds a fail-fast verification to the `test_config_file_user_override` e2e test. If the config file replacement silently fails (e.g., due to template format changes), the test will now fail
immediately with a clear error message showing the actual config contents, rather than failing later with a confusing assertion about unexpected user output.

I ran the e2e test:

```
===================================================================================================================== slowest 5 durations ======================================================================================================================
318.11s setup    test/e2e/test_override_job_user.py::TestWindowsJobUserOverride::test_config_file_user_override
61.22s call     test/e2e/test_override_job_user.py::TestWindowsJobUserOverride::test_config_file_user_override
8.22s teardown test/e2e/test_override_job_user.py::TestWindowsJobUserOverride::test_config_file_user_override
=========================================================================================================== 1 passed, 1 warning in 387.60s (0:06:27) ===========================================================================================================
```

No documentation needed - this is an internal test improvement.

No. This only affects test behavior and does not change any production code or public APIs.

----

### DIFF
--- a/test/e2e/test_override_job_user.py
+++ b/test/e2e/test_override_job_user.py
@@ -14,6 +14,7 @@ import os
 import logging
 
 from e2e.conftest import DeadlineResources
+from e2e.utils import windows_replace_and_verify
 from deadline_test_fixtures import (
     Job,
     Farm,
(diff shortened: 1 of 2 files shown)


leongdl (dismissed):

joel-wong-aws (approved):
Synced with Trevor. This will make the test fail with a more useful error message (but won't make it less flaky). The more useful error message is helpful by itself. Approving.

epmog (approved):

leongdl (approved):

PR #888, opened 2026-02-26
feat: pass host AWS_ENDPOINT_URL env variables to job attachment sync actions

### LINKED ISSUE #882: Feature request: Support for AWS_ENDPOINT_URL_S3
### Describe the problem

We want to use https://repost.aws/articles/ARaWiUYarLQSanDgHfMAfbsA/introducing-s3-hybrid-cache-an-intelligent-caching-proxy-for-amazon-s3 as `AWS_ENDPOINT_URL_S3` but it does not seem like it is respected. Is there some way to configure deadline cloud worker agent to use a custom ENDPOINT_URL for all of the S3 interactions?

Related to https://github.com/aws-deadline/deadline-cloud-worker-agent/issues/746

### Proposed Solution

https://repost.aws/articles/ARaWiUYarLQSanDgHfMAfbsA/introducing-s3-hybrid-cache-an-intelligent-caching-proxy-for-amazon-s3 and working AWS_ENDPOINT_URL_S3

### Example Use Cases

Caching S3 files between runs (different machines in the same office)

epmog:
Fixes #882 

With the new "sync as job user" feature, users are unable to pass along any endpoint url overrides to the sync action since it explicitly disregards the session env.

Allowlist `AWS_ENDPOINT_URL` env variables so that they're passed along.

Users should be able to set these endpoint overrides on their worker host and have them work in the job attachment sync action.

Added some tests, but need to test manually

This might be something we want to document for users. Not currently sure the best location.

N/A

----

### DIFF
--- a/src/deadline_worker_agent/sessions/session.py
+++ b/src/deadline_worker_agent/sessions/session.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import os
 from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
@@ -1191,6 +1192,16 @@ class Session:
         os_env_vars: Optional[dict[str, str]] = None,
         log_task_banner: bool = True,
     ) -> None:
+        # Forward any AWS endpoint URL overrides from the host environment so that
+        # customers can direct S3 (or other service) traffic to custom endpoints.
+        endpoint_overrides = {
+            k: v for k, v in os.environ.items() if k.startswith("AWS_ENDPOINT_URL")
+        }
+        if endpoint_overrides:
+            if os_env_vars is None:
+                os_env_vars = {}
+            os_env_vars.update(endpoint_overrides)
+
         self._session._run_task_without_session_env(
             step_script=step_script,
             task_parameter_values=task_parameter_values,
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -189,6 +189,13 @@ def logs_client() -> MagicMock:
 
 
 @pytest.fixture(autouse=True)
+def clear_aws_endpoint_url_env_vars(monkeypatch: pytest.MonkeyPatch):
+    for key in list(os.environ):
+        if key.startswith("AWS_ENDPOINT_URL"):
+            monkeypatch.delenv(key)
+
+
+@pytest.fixture(autouse=True)
 def patch_windows_session_user_validate():
     with patch.object(WindowsSessionUser, "_validate_username_password"):
         yield
--- a/test/unit/sessions/test_session.py
+++ b/test/unit/sessions/test_session.py
@@ -2524,3 +2524,51 @@ class TestRunAttachmentSyncTask:
             )
 
         assert exc_info.value is expected_exception
+
+    def test_forwards_aws_endpoint_url_overrides(
+        self,
+        session: Session,
+        mock_openjd_session: MagicMock,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        """Tests that AWS_ENDPOINT_URL* env vars from the host are forwarded."""
+        # GIVEN
+        monkeypatch.setenv("AWS_ENDPOINT_URL_S3", "https://custom-s3-endpoint")
+        monkeypatch.setenv("AWS_ENDPOINT_URL", "https://custom-global-endpoint")
+        monkeypatch.setenv("UNRELATED_VAR", "should-not-appear")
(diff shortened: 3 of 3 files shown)


src/deadline_worker_agent/sessions/session.py
```diff
@@ -1191,6 +1192,16 @@ def _run_attachment_sync_task(
         os_env_vars: Optional[dict[str, str]] = None,
         log_task_banner: bool = True,
     ) -> None:
+        # Forward any AWS endpoint URL overrides from the host environment so that
+        # customers can direct S3 (or other service) traffic to custom endpoints.
+        endpoint_overrides = {
```
leongdl, line 1197: This is great for CMF setup and configuring the workers. No need for Queue Env overrides. So yeah, exporting it in 1 place is good.

One side thought, would a customer ever want different endpoint URLs by queue ? Probably too niche right?

komapa (approved):

komapa:
Would love to see this get merged :)

src/deadline_worker_agent/sessions/session.py
leongdl: I would take the thumbs down as no to: "want different endpoint URLs by queue ? Probably too niche right?"

leongdl (approved):

leongdl:
> Would love to see this get merged :)

Working on getting reviews.

sakshie95:
The change looks good overall. Two things - 1. Needs a rebase with mainline. 2. The description says "Added some tests, but need to test manually" - is there some testing that needs to be performed for the feature so it doesn't break asset sync as job user?

sakshie95 (approved):
Approved with a comment

epmog:
> Would love to see this get merged :)

Heh sorry that's on me. I put something up quickly and floated this around the team but it's not quite right (it's a little weird to special case these environments from the worker host in general) and I have to do some testing about the below.

### My understanding

1. The user's environment block (windows) or profile (unix) is loaded and available to the session actions.
    1. This will be the the `job-user` or `worker-user` based on how the worker-agent is configured.
2. The `no session env variable` function for job attachments sync is strictly about not propagating variables set in the session (ie. `openjd_env: `), it'll still have the user's already set environment (changes based on service vs. not service)

### What that means for you
So. If you update your existing worker agent's `deadline-cloud` install to `0.54.1`, STS/S3 endpoint url overrides are available by modifying the worker-agent's environment via the registry we talked about in the issue since by default it's a special action that's run as the `worker-user`.

On the next worker agent release, this behaviour will change (there'll be a release note indicating this) where the job sync attachment action will swap to run as the job-user (if it's configured). In which case, for the endpoint url override to work you'll have to have it set as the job-user.

### What we're going to do

So we shouldn't need this code change and I'm trying to figure out how we can document this type of configuration so it's not so confusing for users because it all depends on if actions are being run as the job-user or worker-user, and if it's running as a service or not.

Let me know if that helps.

komapa:
> > Would love to see this get merged :)
> 
> Heh sorry that's on me. I put something up quickly and floated this around the team but it's not quite right (it's a little weird to special case these environments from the worker host in general) and I have to do some testing about the below.
> 
> ### My understanding
> 1. The user's environment block (windows) or profile (unix) is loaded and available to the session actions.
>    
>    1. This will be the the `job-user` or `worker-user` based on how the worker-agent is configured.
> 2. The `no session env variable` function for job attachments sync is strictly about not propagating variables set in the session (ie. `openjd_env: `), it'll still have the user's already set environment (changes based on service vs. not service)
> 
> ### What that means for you
> So. If you update your existing worker agent's `deadline-cloud` install to `0.54.1`, STS/S3 endpoint url overrides are available by modifying the worker-agent's environment via the registry we talked about in the issue since by default it's a special action that's run as the `worker-user`.
> 
> On the next worker agent release, this behaviour will change (there'll be a release note indicating this) where the job sync attachment action will swap to run as the job-user (if it's configured). In which case, for the endpoint url override to work you'll have to have it set as the job-user.
> 
> ### What we're going to do
> So we shouldn't need this code change and I'm trying to figure out how we can document this type of configuration so it's not so confusing for users because it all depends on if actions are being run as the job-user or worker-user, and if it's running as a service or not.
> 
> Let me know if that helps.

I generally do not have preference on how this would/should work. We control the host this is running on so we can set ENV variables at any level, as long as it is possible to set the S3 endpoint, it would not matter. I can try `deadline-cloud`  version `0.54.1` and see if this unblocks us for now. 

I personally would also prefer as little change as possible around this so please take your time if you need to get it right and if the `deadline-cloud` already unblocked us.

PR #934, opened 2026-05-12
fix: Reconfigure stdout/stderr to UTF-8 in attachment_upload.py

### LINKED ISSUE #928: UnicodeEncodeError in attachment_upload.py when file paths contain non-ASCII characters
## Bug Description

The `attachment_upload.py` script crashes with a `UnicodeEncodeError` when output file paths contain non-ASCII characters (e.g., Cyrillic, Unicode curly quotes). This happens even though the render itself completes successfully — the job fails only during the output upload phase.

This is particularly frustrating because the customer pays for the render compute time, the frame renders correctly, but the output is never uploaded.

## Error

```
File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\sessions\actions\scripts\attachment_upload.py", line 472, in <module>
    main()
  File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\aws\deadline\__init__.py", line 976, in wrapper
    raise e
  File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\aws\deadline\__init__.py", line 973, in wrapper
    return function(*args, **kwargs)
  File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\sessions\actions\scripts\attachment_upload.py", line 436, in main
    root_path_to_output_manifest = snapshot(
  File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\sessions\actions\scripts\attachment_upload.py", line 67, in wrapper
    raise e
  File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\sessions\actions\scripts\attachment_upload.py", line 61, in wrapper
    return function(*args, **kwargs)
  File "C:\Program Files\Python311\Lib\site-packages\deadline_worker_agent\sessions\actions\scripts\attachment_upload.py", line 166, in snapshot
    print(
  File "C:\Program Files\Python311\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 41-47: character maps to <undefined>
```

## Root Cause

On Windows, the default console encoding is `cp1252`. The `print()` call at line 166 in `attachment_upload.py` (inside the `snapshot` function) attempts to print a path that contains characters not representable in `cp1252`. The script does not set `PYTHONIOENCODING=utf-8` or use `encoding='utf-8'` when writing to stdout.

## Reproduction

1. Submit a 3ds Max (or any DCC) job where the scene references files with non-ASCII characters in their paths (e.g., Cyrillic filenames, Unicode curly quotes)
2. The render completes successfully
3. During "Job Attachments Upload for Task", the script crashes when trying to print the manifest/snapshot path

Example filenames that trigger this:
- `tmekdbzbw_8K_Albedo — копия.jpg` (em dash + Cyrillic)
- `Фон.jpg` (Cyrillic)
- `Arctostaphylos 'Sunset'.cgeo` (Unicode curly quotes U+2018/U+2019)

## Environment

- Windows Server (EC2)
- Python 3.11
- deadline-cloud-worker-agent (latest as of 2026-05-11)
- Render completed successfully with Corona renderer in 3ds Max 2025

## Suggested Fix

Either:
1. Set `PYTHONIOENCODING=utf-8` in the subprocess environment when launching `attachment_upload.py`
2. Or wrap stdout with a UTF-8 writer at the top of the script: `sys.stdout.reconfigure(encoding='utf-8')`
3. Or use `errors='replace'` / `errors='backslashreplace'` in the print statement to gracefully handle unencodable characters

## Workaround

Set `PYTHONIOENCODING=utf-8` as a system environment variable on the worker fleet AMI/instances.

rickrams:
On Windows, the default console encoding is cp1252 which cannot encode non-ASCII characters (Cyrillic, Unicode quotes, etc.) in file paths. This causes UnicodeEncodeError when print() is called with paths containing such characters during the output upload phase.

Add sys.stdout.reconfigure(encoding='utf-8') and
sys.stderr.reconfigure(encoding='utf-8') in the __main__ block before calling main() to ensure all output is encoded as UTF-8 regardless of the system's default encoding.

Closes: https://github.com/aws-deadline/deadline-cloud-worker-agent/issues/928

Tests added

N/A

No
----


test/unit/sessions/actions/scripts/test_attachment_upload.py
```diff
@@ -541,6 +542,67 @@ def test_main_no_output(
...
+            # Simulate a non-UTF-8 default (like Windows cp1252)
+            sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='cp1252')
+            sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding='cp1252')
+            # Apply the same fix as the __main__ block
+            sys.stdout.reconfigure(encoding='utf-8')
+            sys.stderr.reconfigure(encoding='utf-8')
+            # Verify it works with non-ASCII
+            print("Фон — копия.jpg")
+            print(sys.stdout.encoding)
+        """)
```
godobyte: This test spawns a subprocess that manually applies the fix and verifies it works. This proves reconfigure works in general but doesn't test the actual script entry point.

test/unit/sessions/actions/scripts/test_attachment_upload.py
```diff
@@ -541,6 +542,67 @@ def test_main_no_output(
...
+
+        args = ["-s3", "s3://test-bucket/path", "-wp", "/path/to/worker.json"]
+
+        # WHEN/THEN - should not raise UnicodeEncodeError
+        main(args)
+
+    def test_script_reconfigures_stdout_stderr(self):
+        """Test that the __main__ block reconfigures stdout and stderr to UTF-8."""
+        import subprocess
+        import textwrap
```
godobyte: nit - could we follow the convention to import at the top?

leongdl (dismissed):

test/unit/sessions/actions/scripts/test_attachment_upload.py
```diff
@@ -541,6 +542,67 @@ def test_main_no_output(
         )
 
 
+class TestUTF8Reconfiguration:
+    """Test that stdout/stderr are reconfigured to UTF-8 for non-ASCII path support."""
+
+    @patch("deadline_worker_agent.sessions.actions.scripts.attachment_upload.upload_output_assets")
+    @patch("deadline_worker_agent.sessions.actions.scripts.attachment_upload.snapshot")
```
godobyte: This test mocks everything and doesn't actually exercise the `print()` path that crashes. The mock for snapshot returns `{}` so the problematic print inside snapshot is never reached. This test doesn't reproduce the bug.

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
```diff
@@ -475,4 +475,10 @@ def main(args=None):
 
 
 if __name__ == "__main__":
+    # Reconfigure stdout/stderr to use UTF-8 encoding to prevent UnicodeEncodeError
+    # on Windows where the default console encoding (cp1252) cannot encode non-ASCII
+    # characters in file paths (e.g., Cyrillic, Unicode quotes).
+    # See: https://github.com/aws-deadline/deadline-cloud-worker-agent/issues/928
+    sys.stdout.reconfigure(encoding="utf-8")  # type: ignore[union-attr]
+    sys.stderr.reconfigure(encoding="utf-8")  # type: ignore[union-attr]
```
godobyte: I think we should consider setting PYTHONIOENCODING=utf-8 in the [subprocess environment when launching attachment_upload.py](https://github.com/rickrams/deadline-cloud-worker-agent/blob/cc01ddd317ed58f40c1835713f182e05880dc130/src/deadline_worker_agent/sessions/actions/run_attachment_upload.py#L186-L190). That would fix the encoding for the entire subprocess regardless of what print() calls exist inside.

Similarly, would be nice to apply the same fix for [attachment_download.py](https://github.com/rickrams/deadline-cloud-worker-agent/blob/cc01ddd317ed58f40c1835713f182e05880dc130/src/deadline_worker_agent/sessions/actions/run_attachment_download.py#L380-L382).

test/unit/sessions/actions/scripts/test_attachment_upload.py
rickrams: Done — moved subprocess and textwrap to the top-level imports

test/unit/sessions/actions/scripts/test_attachment_upload.py
rickrams: Good catch. Replaced with test_snapshot_prints_non_ascii_paths which only mocks _manifest_snapshot and actually exercises the print() calls inside snapshot() with Cyrillic and Unicode quote characters in the paths.

test/unit/sessions/actions/scripts/test_attachment_upload.py
rickrams: Fair point — kept this as a secondary test that validates the reconfigure approach works in a cp1252 environment (which we can't easily simulate in-process on Linux/macOS CI). The primary regression coverage is now in test_snapshot_prints_non_ascii_paths which exercises the real code path.

test/unit/sessions/actions/scripts/test_attachment_upload.py
```diff
@@ -541,6 +541,35 @@ def test_main_no_output(
         )
 
 
+class TestUTF8Reconfiguration:
+    """Test that PYTHONIOENCODING=utf-8 is set for non-ASCII path support."""
+
+    @patch("deadline_worker_agent.sessions.actions.scripts.attachment_upload._manifest_snapshot")
+    def test_snapshot_prints_non_ascii_paths(self, mock_manifest_snapshot: Mock):
```
godobyte: This test is not clearly reflecting the latest change. The mocking defeats the purpose of testing the print in `_manifest_snapshot`. Based on the previous reply, I think you meant to keep the other test?

src/deadline_worker_agent/sessions/actions/scripts/attachment_upload.py
rickrams: Good call — switched to setting PYTHONIOENCODING=utf-8 in the subprocess environment at both launch sites (run_attachment_upload.py and run_attachment_download.py). Dropped the reconfigure call from the script. Cleaner and covers the entire subprocess from startup.

test/unit/sessions/actions/scripts/test_attachment_upload.py
rickrams: You're right — replaced with test_attachment_upload_handles_non_ascii_with_pythonioencoding which spawns a subprocess with PYTHONIOENCODING=utf-8 in the env (matching what run_attachment_upload.py now does) and verifies it can print Cyrillic/Unicode paths without error. The launch-site tests in test_run_attachment_upload.py and test_run_attachment_download.py also assert the env var is passed.

godobyte (dismissed):

godobyte (approved):

leongdl (approved):

PR #939, opened 2026-05-19
feat!: report task runs as INTERRUPTED if interrupted during input asset sync or env enter

jericht:
The Worker Agent protocol currently has task runs being reported as `NEVER_ATTEMPTED` in the event of a spot interruption during input asset attachment sync or env enter phases. We have received customer feedback that it would be better to have the task runs reported as `INTERRUPTED` in these cases so it is easier to tell at a glance from the Deadline Cloud monitor which tasks were not attempted due to spot interruptions during these phases.

- Update the worker protocol so that task runs are reported as `INTERRUPTED` if a spot interruption occurs during input asset sync or env enter
- Update the Worker agent to report task runs as `INTERRUPTED` instead of `NEVER_ATTEMPTED` in the above scenario

Task runs are now reported as `INTERRUPTED` rather than `NEVER_ATTEMPTED` if a spot interruption occurs during input asset sync or env enter

- Unit tests
- [WIP] E2E test
- Manual e2e testing via code in this branch: https://github.com/jericht/deadline-cloud-worker-agent/tree/jericht/spot_int

No

Yes

----

### DIFF
--- a/docs/dev/worker_api_protocol.md
+++ b/docs/dev/worker_api_protocol.md
@@ -360,14 +360,26 @@ reported as `NEVER_ATTEMPTED` must not report `startedAt` or `endedAt` times. Th
 must not queue additional Session Actions to the Session except for `envExit` Session Actions that
 correspond to already completed `envEnter` Session Actions for the same Session.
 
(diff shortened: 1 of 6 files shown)


leongdl:
Stale PR - closing. We can re-open and recreate as needed.

PR #940, opened 2026-05-20
feat: support domain users

### LINKED ISSUE #458: Feature request: Install worker agent to run as a domain user on Windows
### Use Case

To be able to setup the worker agent to run as a domain user. This is important for compliance and consistency when setting up worker hosts in certain organizations. Some organizations want all user accounts to be centrally managed in a directory/domain.

### Proposed Solution

Modify the worker agent installer to support supplying a `--user` argument that specifies a domain user. This should support specifying the user using "user principal name" and/or "down-level logon name" formats (see https://learn.microsoft.com/en-us/windows/win32/secauthn/user-name-formats).

### LINKED ISSUE #652: Bug: Down-level usernames throw an error
### Describe the bug

When the customer used Down-Level Logon Name format for entering queue-users ( e.g. DOMAIN\USER) this causes error in the worker agent:
```
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\startup\entrypoint.py", line 262, in entrypoint
    worker_sessions.run()
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\worker.py", line 261, in run
    future.result()
  File "C:\Python311\Lib\concurrent\futures\_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception
  File "C:\Python311\Lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\scheduler\scheduler.py", line 305, in run
    interval = self._sync(interruptable=True)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\scheduler\scheduler.py", line 468, in _sync
    self._update_sessions(response=response)
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\scheduler\scheduler.py", line 627, in _update_sessions
    created_session_ids = self._create_new_sessions(assigned_sessions=assigned_sessions)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\scheduler\scheduler.py", line 924, in _create_new_sessions
    job_details = job_entities.job_details()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\sessions\job_entities\job_entities.py", line 337, in job_details
    self._windows_credentials_resolver.get_windows_session_user(
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\windows\win_credentials_resolver.py", line 186, in get_windows_session_user
    cache_entry = get_windows_credentials(user, password)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\deadline_worker_agent\windows\win_logon.py", line 83, in get_windows_credentials
    logon_token = LogonUser(
                  ^^^^^^^^^^
pywintypes.error: (1326, 'LogonUser', 'The user name or password is incorrect.')
```

### Expected Behaviour

It should support down-level usernames or if a Down-Level Logon Name in the console is added ... you get told you can't enter that.

### Current Behaviour

The crash happens on the Worker agent service when the job attempts to run - you don't get any errors feedback in the job task ( the monitor is blank, and the task its starts to run is red with an error) - you only see the error in the Worker logs. And because we're testing on a CMF , we noticed that the service had stopped.

Can a sanity check be added to the console so that when a user is entered, and it is deemed to be Down-Level Logon Name , the user is preventing from doing so?

The alternate is a bug that can only caught when looking at worker logs.

And - I'm not sure if you want to flag this as another bug - when this happens the agent services crashes, and needs to be restarted on the worker.

### Reproduction Steps

Add a `domain\user` format user as a queue user. 

I think adding below code should resolve the issue: 

change [win_logon.py on line#79](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/75eda8eb3add4561bd135e6973a9437e372903f6/src/deadline_worker_agent/windows/win_logon.py#L81) like below:
(file lives here deadline-cloud-worker-agent/src/deadline_worker_agent/windows)
```python
    try:
        # Handle domain\user format by splitting into domain and username
        domain = None
        user = username
        if '\\' in username:
            domain, user = username.split('\\', 1)
            
        # https://timgolden.me.uk/pywin32-docs/win32profile__LoadUserProfile_meth.html
        logon_token = LogonUser(
            Username=user,
            Domain=domain,
            LogonType=LOGON32_LOGON_INTERACTIVE,
            LogonProvider=LOGON32_PROVIDER_DEFAULT,
            Password=password,
        )
```

### Environment

At minimum:
1. Operating system Windows CMF
2. Version of this package

Please share other details about your environment that you think might be relevant to reproducing the bug.

epmog:
Fixes: #458, #652

  ~~Can't be merged until 0.10.9 openjd-sessions is released:
  https://github.com/OpenJobDescription/openjd-sessions-for-python/pull/311~~ 0.10.9 released

  The worker agent needed to support Active Directory domain users as both the agent service identity and the job
  run-as user. This includes supporting both username formats: Down-Level Logon Name (`DOMAIN\user`) and User
  Principal Name (`user@domain.com`).

  Additionally, we need to ensure that this works as expected with permissions (job attachments) and GPU access.

  The config-level `windows_job_user` override also needed to work for domain users, and local users needed an option
  to use externally managed passwords (via Secrets Manager) instead of the agent resetting them on every startup.

  - **UPN handling in installer**: `is_user_in_group()` and `get_effective_user_rights()` now handle UPN format
  (resolving to DDL where Win32 APIs require it).
  - **Session cleanup**: `_extract_username()` handles both DDL (split on `\`) and UPN (split on `@`) when matching
  process owners for cleanup.
  - **Config-level domain user override**: Added `WindowsUserSettings` dataclass to hold user + password ARN for
  deferred credential resolution. The scheduler resolves credentials via `WindowsCredentialsResolver` at session
  creation time, taking precedence over the queue-configured user.
  - **Password ARN support for local users**: When `windows_job_user_password_arn` is set for a local user, the agent
  fetches the password from Secrets Manager instead of resetting it via `NetUserSetInfo`. This allows externally
  managed local user passwords.
  - **Config file parsing**: Added `windows_job_user_password_arn` to `OsConfigSection` so it can be read from the
  TOML config file (previously only worked via environment variable).
  - **CLI flag**: Added `--windows-job-user-password-arn` command-line argument.
  - **E2E tests**: wrote domain user tests to use a job bundle (`domain_user_whoami`) that validates identity via
  `whoami` and verifies job attachments (input read + output write). Tests are parameterized across DDL and UPN
  formats. Added `test_job_runs_as_domain_config_override_user` to verify the config-level override path.
  `test_service_identity` now verifies the exact format is preserved in the service configuration.
  - **Test infrastructure**: Added `job_failure_message()` helper for detailed failure diagnostics with console/log
  links. Fixed queue-fleet association teardown to properly stop scheduling before deletion. Config override test uses
  `try/finally` to ensure cleanup.
  - **Dependency**: Pinned `openjd-sessions == 0.10.9` which includes domain user support (domain parsing in
  `LogonUserW`/`CreateProcessWithLogonW` and SID-based `is_process_user()`).
  - **Entrypoint**: Guarded `sys.stderr.write` for Windows service context where stderr may be None.
  - **GPU bundles**: Added `nvidia_smi` and `blender_gpu_render` job bundle templates for future GPU testing.

  Enables the worker agent to run in Active Directory environments where both the agent and job users are domain
  accounts. Both DDL and UPN username formats work end-to-end. Job attachments (S3 sync) work correctly under domain
  user sessions.

  Local Windows users can now optionally use Secrets Manager for password management instead of having the agent reset
  their password on every startup.

  - E2E tests (`test_domain_user.py`) deploy a Windows Server EC2 instance, promote it to a domain controller, create
  domain users, install the agent as a domain user, and submit jobs that verify identity and job attachments.
  Parameterized across DDL and UPN formats.
  - Unit tests for config loading (domain user + ARN, local user + ARN, domain user without ARN raises error),
  scheduler domain user resolution, config file parsing of `windows_job_user_password_arn`, and CLI argument parsing.
  - Manually verified on Windows Server 2022 with `openjd-sessions == 0.10.9`.

  Yes — `docs/dev/domain_users.md` updated with session cleanup changes, `is_process_user()` SID comparison details,
  password management table (including local user + ARN path), and Kerberos S4U future work section.
  `worker.toml.example` updated with corrected documentation for `windows_job_user_password_arn`.

  No. Existing local user workflows are unaffected (password reset remains the default when no ARN is specified).
  Domain user support is additive. The `openjd-sessions` pin to `== 0.10.9` is a dependency constraint change but does
  not alter public API.

  ----

  *By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution,
  under the terms of your choice.*


docs/dev/domain_users.md
```diff
@@ -0,0 +1,243 @@
...
+
+1. The Deadline service provides a `user` and `passwordArn` (configured on the queue's `jobRunAsUser.windows`)
+2. The agent fetches the password from Secrets Manager via `GetSecretValue`
+3. The agent calls `LogonUser` with the credentials and caches the logon token
+
+**Password:** Stored in AWS Secrets Manager as `{"password": "value"}`. The AD admin sets/rotates the password in AD, and keeps the Secrets Manager secret in sync (manually or via automation). The agent re-fetches on cache miss or credential failure.
+
+The only difference: local queue users can fall back to `reset_user_password()` (via `NetUserSetInfo`) if no `passwordArn` is configured. Domain users cannot — they always require a Secrets Manager secret.
+
+## Changes Required
```
leongdl: Does this section need to be part of the dev docs? like intended to be kept for ever?

docs/dev/domain_users.md
```diff
@@ -0,0 +1,243 @@
...
+
+### openjd-sessions
+
+| File | Change | Status |
+|------|--------|--------|
+| `_win32/_helpers.py` | `logon_user()` parses domain from username and passes it to `LogonUserW` | ✅ Done |
```
leongdl: This looks like a kiro to spec :)

src/deadline_worker_agent/config/config.py
```diff
@@ -180,18 +189,38 @@ def __init__(
...
+                # Domain users cannot have their password reset locally.
+                # They must provide a password ARN to fetch credentials from Secrets Manager.
+                if not settings.windows_job_user_password_arn:
+                    raise ConfigurationError(
+                        f"Domain user '{settings.windows_job_user}' requires 'windows_job_user_password_arn' "
+                        "to be set. Domain user passwords must be managed externally via Secrets Manager."
+                    )
+                # For domain users, we defer credential resolution to runtime via the credentials resolver.
+                # Store the settings so the scheduler can use WindowsCredentialsResolver.
+                from ..sessions.job_entities.job_details import JobRunAsWindowsUser
```
leongdl: nit, inteneded inline import?

leongdl (changes requested):
nit comments only, can we clean up the docs/dev?

src/deadline_worker_agent/config/config.py
epmog: Yea, it's cyclical otherwise

docs/dev/domain_users.md
epmog: It's not super required to stay forever was just an easy way to track/update as it chugged along

src/deadline_worker_agent/config/config.py
epmog: I'm gonna remove this though since it's not the right class. that's specifically for job_details api.

leongdl (dismissed):

src/deadline_worker_agent/windows/win_user.py
```diff
@@ -0,0 +1,17 @@
...
+def is_domain_user(username: str) -> bool:
+    """Returns True if the username is in a domain format.
+
+    Domain users can be specified in two formats:
+    - Down-Level Logon Name (DDL): DOMAIN\\username
+    - User Principal Name (UPN): username@domain.com
+
+    See https://learn.microsoft.com/en-us/windows/win32/secauthn/user-name-formats
+    """
+    return "\\" in username or "@" in username
```
seant-aws, line 17: I'm not an expert on Windows, but this seems like it could potentially be flimsy. Would this cause issues? Or is there more of a user side guard that protects usernames from being all funky

seant-aws (dismissed):

test/e2e/utils.py
```diff
@@ -538,6 +538,48 @@ def windows_replace_and_verify(
...
+
+
+def job_failure_message(
+    job: Job,
+    deadline_client: DeadlineClient,
+    queue: Queue,
+    deadline_resources: DeadlineResources,
+) -> str:
+    """Build a detailed failure message with job identifiers, console links, and logs."""
+    import boto3
```
godobyte: nit - should we do import at the top of file?

test/unit/config/test_cli_args.py
```diff
@@ -178,6 +178,26 @@ def test_windows_job_user(self, arg_parser: ArgumentParser, windows_job_user: st
         # THEN
         assert result.windows_job_user == windows_job_user
 
+    @pytest.mark.skipif(os.name != "nt", reason="Windows-only argument")
```
godobyte, line 181: I saw both this and `sys.platform == "win32"` are used, are they serving different purposes?

src/deadline_worker_agent/windows/win_user.py
epmog: This check is just moved from another location. In this case it's actually pretty safe, you can follow the link. In which case we're just opening it up to more users.

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,442 @@
...
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        try:
+            response = ssm_client.describe_instance_information(
+                Filters=[{"Key": "InstanceIds", "Values": [instance_id]}]
+            )
+            instances = response.get("InstanceInformationList", [])
+            if instances and instances[0].get("PingStatus") == "Online":
+                return
+        except Exception:
```
godobyte: nit - should we add some log visibility for the exception to help debugging?

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,442 @@
...
+DOMAIN_JOB_USER = "domain-job-user"
+WINDOWS_PASSWORD_SECRET = "WindowsPasswordSecret"
+
+AGENT_USER_DDL = f"{DOMAIN_NETBIOS}\\{DOMAIN_AGENT_USER}"
+AGENT_USER_UPN = f"{DOMAIN_AGENT_USER}@{DOMAIN_NAME}"
+
+
+def wait_for_ssm_online(ssm_client, instance_id: str, timeout: int = 600) -> None:
+    """Wait for SSM agent to report the instance as online."""
+    deadline = time.time() + timeout
```
godobyte: `deadline` means something different in my vocabulary 🙈

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,442 @@
...
+        try:
+            response = ssm_client.describe_instance_information(
+                Filters=[{"Key": "InstanceIds", "Values": [instance_id]}]
+            )
+            instances = response.get("InstanceInformationList", [])
+            if instances and instances[0].get("PingStatus") == "Online":
+                return
+        except Exception:
+            pass
+        time.sleep(10)
```
godobyte: Q - how long this usually takes, would this test lead to CI test timeout?

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,442 @@
...
+        LOG.info("Waiting for instance to come back online after DC promotion...")
+        time.sleep(120)
+        wait_for_ssm_online(boto3.client("ssm"), worker.instance_id, timeout=600)
+
+        LOG.info("Verifying AD Domain Services are ready...")
+        cmd_result = worker.send_command("Import-Module ActiveDirectory; Get-ADDomain")
+        assert cmd_result.exit_code == 0, f"AD not ready after promotion: {cmd_result}"
+
+        # Install NVIDIA GRID drivers if on a GPU instance
+        if os.environ.get("WORKER_INSTANCE_TYPE", "").startswith("g"):
```
godobyte: Where is this configured, are we going to run this in automated worker e2e tests?

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,442 @@
...
+    os.environ.get("OPERATING_SYSTEM") != "windows",
+    reason="Domain user tests are Windows-only",
+)
+@pytest.mark.parametrize(
+    "agent_user_format",
+    [AGENT_USER_DDL, AGENT_USER_UPN],
+    ids=["ddl", "upn"],
+    scope="class",
+)
+class TestDomainUser:
```
godobyte, line 167: Did a quick analysis with AI about test run time, do they look correct to you? How long does it takes to run this test class locally, do we need to increase test workflow execution time?

```
 Normal case estimate:

   - EC2 instance launch + SSM ready: ~3–5 min (from class_worker fixture, not shown here)
   - AD DS install: ~1–2 min
   - DC promotion + reboot: 120s sleep + ~30s SSM poll = ~2.5 min
   - Create users + grant rights: ~10s
   - GPU driver install (if GPU instance): ~2–3 min
   - Install agent (DDL): ~60s
   - Run 4 tests × ~2–3 min each (job submit + wait): ~8–12 min
   - Reinstall agent (UPN): ~60s
   - Run 4 tests again: ~8–12 min
   - Teardown (stop QFA + delete): ~30s

   Normal total: ~25–35 minutes

   Worst case:

   - SSM online timeout maxes out: +600s
   - Jobs take longer or retry: +several minutes per job
   - QFA teardown polls full 60s

   Worst case total: ~45–50 minutes
```

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,442 @@
...
+
+    @staticmethod
+    def submit_whoami_job(
+        test_name: str,
+        deadline_client: DeadlineClient,
+        farm: Farm,
+        queue: Queue,
+        expected_user: str,
+    ) -> Job:
+        from e2e.utils import submit_job_from_bundle
```
godobyte: nit - could we move functional it to the top, unless there are specific reasons (i.e. platform-guarded or circular deps)?

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,442 @@
...
+        return submit_job_from_bundle(
+            deadline_client=deadline_client,
+            farm=farm,
+            queue=queue,
+            bundle_path=bundle_path,
+            job_parameters=[{"name": "ExpectedUser", "value": expected_user}],
+            max_retries_per_task=0,
+        )
+
+    def test_job_runs_as_local_queue_user(
```
godobyte, line 308: I thought we wanted to start off as flaky with multiple runs, gather data, and remove when confirmed otherwise. Should we add the flaky decorators for tests introduced?

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,442 @@
...
+    def test_job_runs_as_domain_config_override_user(
+        self,
+        deadline_resources: DeadlineResources,
+        domain_controller: EC2InstanceWorker,
+        deadline_client: DeadlineClient,
+        region: str,
+    ) -> None:
+        """Config-level domain user override resolves credentials and runs jobs as that user."""
+        secretsmanager_client = boto3.client("secretsmanager", region_name=region)
+        secret = secretsmanager_client.describe_secret(SecretId=WINDOWS_PASSWORD_SECRET)
```
godobyte: Does it exist in the current infra, or we need to add that?

test/e2e/job_bundles/blender_gpu_render/template.yaml
```diff
@@ -0,0 +1,22 @@
+specificationVersion: jobtemplate-2023-09
```
godobyte: Nit: The three GPU job bundles (nvidia_smi, blender_gpu_render, blender_benchmark) aren't referenced by any test in this PR. Could you either wire them into a test (even if skipped by default without a GPU instance), add a brief README in the job_bundles/ directory explaining their intended use, or split them into a follow-up PR? As-is they're easy to forget about and go stale.

src/deadline_worker_agent/config/config.py
```diff
@@ -143,6 +158,10 @@ def __init__(
             settings_kwargs["posix_job_user"] = parsed_cli_args.posix_job_user
         if parsed_cli_args.windows_job_user is not None:
             settings_kwargs["windows_job_user"] = parsed_cli_args.windows_job_user
+        if getattr(parsed_cli_args, "windows_job_user_password_arn", None) is not None:
```
godobyte: nit - should we be consistent with other args checking

```suggestion
        if parsed_cli_args.windows_job_user_password_arn is not None:
```

src/deadline_worker_agent/config/config.py
```diff
@@ -180,18 +199,45 @@ def __init__(
...
+                self.job_run_as_user_overrides = JobsRunAsUserOverride(
+                    run_as_agent=settings.run_jobs_as_agent_user,
+                    windows_user_settings=WindowsUserSettings(
+                        user=settings.windows_job_user,
+                        password_arn=settings.windows_job_user_password_arn,
+                    ),
+                )
+            else:
+                if settings.windows_job_user_password_arn:
+                    # Local user with externally managed password — defer resolution like domain users.
```
godobyte, line 220: Curious, what's the customer use case for this?

src/deadline_worker_agent/installer/win_installer.py
```diff
@@ -277,7 +276,12 @@ def is_user_in_group(group_name: str, user_name: str) -> bool:
         logging.error(f"Failed to get group members of '{group_name}': {e}")
         raise
 
-    return any(group_member["name"] == user_name for group_member in group_members_info[0])
+    return any(
+        group_member["name"].lower() == user_name.lower()
+        or group_member["name"].lower() == user_name.split("\\")[-1].lower()
+        or group_member["name"].lower() == user_name.split("@")[0].lower()
```
godobyte: Could this matching logic cause false-positive when a local user and domain user share the same short name? For example, if local bob is in Administrators and you check `DOMAIN\bob`, the `split("\\")[-1]` extracts bob and matches the local member.

godobyte (dismissed):

test/e2e/job_bundles/blender_gpu_render/template.yaml
epmog: I might just remove them. They were more for investigative purposes rather than testing something we have to ensure we don't break.

test/e2e/test_domain_user.py
epmog: it's actually 12-15min that we're adding here. Current timeout is 3hours. and we're around 1h45m on windows before these additions

test/e2e/test_domain_user.py
epmog: yeaa, I'll update that :)

test/unit/config/test_cli_args.py
epmog: effectively the same, I'll make sure it matches within the file

test/e2e/test_domain_user.py
epmog: Yup, already exists: https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/mainline/scripts/e2e_testing_infrastructure.yaml#L633

src/deadline_worker_agent/config/config.py
epmog: We added the configuration option, so I'm just making it consistent behaviour for a local user, since it takes very few lines. It's a lot less important, but yea if you want to manage the password and be able to login to the user, this will be helpful.

test/e2e/test_domain_user.py
epmog: We can, I'll check in a week later to see if they've failed.

test/e2e/test_domain_user.py
epmog: The timeout is 10min but it's closer to like 2ish min, just so we're not spamming the API. No need for backoff stuff here since we know we gotta wait.

pyproject.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
     "boto3 >= 1.34.75",
     "deadline-job-attachments == 0.1.0",
     # Pinned to patch version due to Host Config Script runner usage of private OpenJD Sessions API.
-    "openjd-sessions == 0.10.8",
+    "openjd-sessions == 0.10.9",
```
godobyte, line 18: Closed overlapping dependent bot PR
 https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/946

test/e2e/test_domain_user.py
```diff
@@ -0,0 +1,456 @@
...
+        secret_arn = secret["ARN"]
+
+        config_path = "C:\\ProgramData\\Amazon\\Deadline\\Config\\worker.toml"
+
```
seant-aws, line 379: I think per previous flakey test, I asserted `is_worker_started` before calling `omain_controller.stop_worker_service()`, to make sure worker is in a started/idle state

not fully proven, but from my test runs this was what fixed my previous test. Possible that it isn't an issue here, but just wanted to note

epmog: So for this one, is-worker-started fails if it's stopped, which it already was. My tests don't keep it running (via finally). we can call stop as basically a no-op on it as well.

seant-aws: oh that makes sense actually

seant-aws (dismissed):

test/e2e/test_domain_user.py
epmog: will remove

test/e2e/test_domain_user.py
godobyte: Please double check automated test infra contains this secret and runner has DescribeSecret API access.

godobyte (approved):

src/deadline_worker_agent/installer/win_installer.py
epmog: updated existing issue to be resilient to this!

seant-aws (approved):

PR #966, opened 2026-06-11
chore(deps): bump ruff, deadline, job-attachments, test-fixtures

epmog:
Grouping dependabot updates

----

### DIFF
--- a/requirements-e2e.txt
+++ b/requirements-e2e.txt
@@ -1,12 +1,12 @@
 backoff == 2.2.*
 boto3 >= 1.34.75
-deadline >= 0.56, < 0.57
-deadline-job-attachments >= 0.0, < 0.1
-deadline-cloud-test-fixtures >= 0.18.11, < 0.19
+deadline >= 0.56, < 0.59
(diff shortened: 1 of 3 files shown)


requirements-e2e.txt
```diff
@@ -1,12 +1,12 @@
 backoff == 2.2.*
 boto3 >= 1.34.75
-deadline >= 0.56, < 0.57
-deadline-job-attachments >= 0.0, < 0.1
-deadline-cloud-test-fixtures >= 0.18.11, < 0.19
+deadline >= 0.56, < 0.59
+deadline-job-attachments == 0.1.0
```
godobyte, line 4: Curious, why do we ping specific version on this?

godobyte (approved):

requirements-e2e.txt
epmog: we currently use some private functions that would not be considered breaking if they change. Should also try to fix that :P

AlexTranAmz (approved):

PR #972, opened 2026-06-18
feat: add SessionRuntime ABC and factory

seant-aws:

The worker agent's `Session` class is tightly coupled to `openjd.sessions.Session`. To support multiple runtime backends (Python and Rust), we need an abstraction seam and a factory to select between them.

Added a `SessionRuntime` ABC, a `RuntimeKind` enum (`python`, `rust`, `service-selected`), a frozen `SessionRuntimeConfig` dataclass, and a `create_session_runtime()` factory with lazy adapter imports.

Purely additive — no existing code paths modified.

None. New modules and tests only. Existing behavior is untouched until follow-up PRs wire `Session` through the adapter.

- 12 unit tests (enum values, config immutability, ABC enforcement, factory error paths, adapter injection)
- `hatch run all:test` — 2901 passed (Python 3.9/3.10/3.11)
- `hatch run integ-test` — 73 passed
- Module coverage: 97%

Docstrings on all public classes and functions. No user-facing docs needed.

No.

----

### DIFF
--- a/src/deadline_worker_agent/sessions/runtime/__init__.py
+++ b/src/deadline_worker_agent/sessions/runtime/__init__.py
@@ -1,0 +1,14 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from ._abc import SessionRuntime
+from ._config import ActionCallback, SessionRuntimeConfig
+from ._factory import create_session_runtime
+from ._kind import RuntimeKind
+
+__all__ = [
+    "ActionCallback",
+    "RuntimeKind",
+    "SessionRuntime",
+    "SessionRuntimeConfig",
+    "create_session_runtime",
+]
--- a/src/deadline_worker_agent/sessions/runtime/_abc.py
+++ b/src/deadline_worker_agent/sessions/runtime/_abc.py
@@ -1,0 +1,96 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from datetime import timedelta
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Optional
+
+if TYPE_CHECKING:
+    from openjd.sessions import (
+        ActionStatus,
+        EnvironmentIdentifier,
+        EnvironmentModel,
+        StepScriptModel,
+    )
+
+
+class SessionRuntime(ABC):
+    """Abstract surface every session backend implements.
+
+    Mirrors ``openjd.sessions.Session``'s public surface 1:1.
+    """
+
+    @abstractmethod
+    def enter_environment(
+        self,
+        *,
+        environment: EnvironmentModel,
+        identifier: Optional[EnvironmentIdentifier] = None,
+        os_env_vars: Optional[dict[str, str]] = None,
+    ) -> EnvironmentIdentifier:
+        """Enter an environment; returns its identifier."""
+        ...
+
+    @abstractmethod
+    def exit_environment(
+        self,
+        *,
+        identifier: EnvironmentIdentifier,
+        os_env_vars: Optional[dict[str, str]] = None,
+        keep_session_running: bool = False,
+    ) -> None:
+        """Exit a previously entered environment."""
+        ...
+
+    @abstractmethod
+    def run_task(
+        self,
+        *,
+        step_script: StepScriptModel,
+        task_parameter_values: dict[str, Any],
+        os_env_vars: Optional[dict[str, str]] = None,
+        log_task_banner: bool = True,
+    ) -> None:
+        """Run a task within the session's active environment(s)."""
+        ...
+
+    @abstractmethod
+    def _run_task_without_session_env(
+        self,
+        *,
+        step_script: StepScriptModel,
+        task_parameter_values: dict[str, Any],
+        os_env_vars: Optional[dict[str, str]] = None,
+        log_task_banner: bool = True,
+    ) -> None:
+        """Run a task without entering a session environment (attachment-sync path)."""
+        ...
+
+    @abstractmethod
+    def cancel_action(
+        self,
+        *,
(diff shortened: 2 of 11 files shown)


src/deadline_worker_agent/sessions/runtime/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+    @abstractmethod
+    def enter_environment(
+        self,
+        *,
+        environment: EnvironmentModel,
+        identifier: Optional[EnvironmentIdentifier] = None,
+        os_env_vars: Optional[dict[str, str]] = None,
+    ) -> EnvironmentIdentifier:
+        """Enter an environment; returns its identifier."""
+        ...
```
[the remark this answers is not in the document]
seant-aws: these are all abstract method stubs

src/deadline_worker_agent/sessions/runtime/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+
+    session_id: str
+    job_parameter_values: dict[str, Any]
+    path_mapping_rules: Optional[list[PathMappingRule]]
+    retain_working_dir: bool
+    user: Optional[SessionUser]
+    action_callback: ActionCallback
+    os_env_vars: Optional[dict[str, str]]
+    session_root_directory: Path
+    spec_revision: str = "2023-09"
```
leongdl: This is an interesting one to have? Curious why?

src/deadline_worker_agent/sessions/runtime/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        kind: Which runtime backend to construct.
+        config: Construction arguments forwarded to the adapter.
+
+    Raises:
+        ValueError: ``kind`` is not a recognised RuntimeKind.
+        NotImplementedError: the adapter module for ``kind`` cannot be imported
+            (e.g. the Rust binding is unavailable on this platform). Callers
+            that want graceful degradation should catch this and fall back.
+    """
+    if not isinstance(kind, RuntimeKind):
```
leongdl: That's odd, we already have typed input, wy is this needed?

src/deadline_worker_agent/sessions/runtime/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+    config layers without manual conversion.
+    """
+
+    PYTHON = "python"
+    RUST = "rust"
+    SERVICE_SELECTED = "service-selected"
+
+
+@dataclass(frozen=True)
+class SessionRuntimeConfig:
```
leongdl: How come the classes are in __init__.py?

Can we move them to dedicated python files? Now the whole class is imported  on package import

src/deadline_worker_agent/sessions/runtime/__init__.py
seant-aws: ah just being safe extra safe I guess 😅 I can remove it

src/deadline_worker_agent/sessions/runtime/__init__.py
seant-aws: as in the dataclass ? or the specific spec revisions ?
Just styling on my end, thought It'd be cleaner to have it as dataclass
If you mean the specific fields, I just went by what was passed through initial OPENJDSession init

src/deadline_worker_agent/sessions/runtime/_abc.py
```diff
@@ -0,0 +1,99 @@
...
+        self,
+        *,
+        step_script: StepScriptModel,
+        task_parameter_values: dict[str, Any],
+        os_env_vars: Optional[dict[str, str]] = None,
+        log_task_banner: bool = True,
+    ) -> None:
+        """Run a task without entering a session environment (attachment-sync path).
+
+        Maps to upstream's private ``_run_task_without_session_env``.
```
godobyte: I wonder why did we decide to make it public here. From the doc I noticed 

> If/when openjd-sessions promotes it to public API, drop the underscore here

Are we planning to do it now or when then the upstream becomes public?

test/unit/sessions/runtime/test_abc_factory.py
```diff
@@ -0,0 +1,217 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
godobyte: Nit: The repo convention is one test file per source module (test_kind.py, test_factory.py, etc.). Consider splitting this into separate files to match the _abc.py / _factory.py / _kind.py / _config.py structure, making it easier to find tests for a specific module later.

seant-aws: ah yea fair point I'll note this down

src/deadline_worker_agent/sessions/runtime/_abc.py
seant-aws: I think from my understanding, with the changes to openjd-sessions (in review) coming in this would change to public*
 
 actually I am completely mistaken. It is not being changed. However the original idea was that since all the other methods are public, and also would it make sense to keep private if its simply upstream, but we call it in our own way in the new interface, so it makes sense to be public ?

godobyte: We could call it in our own way in the new interface. However, making a private method public is backward compatible, but not vice versa. That's why we start with private until we're clear it should be public.

seant-aws: noted, i'll switch to private for now

test/unit/sessions/runtime/test_runtime_config.py
```diff
@@ -0,0 +1,17 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
godobyte, line 1: nit - already in `/runtime/` subdir, we can rename to `test_config.py`

godobyte (approved):

leongdl (approved):

PR #982, opened 2026-06-30
feat: add runtime selection config flags

seant-aws:

The worker agent needs the ability to select between different session runtime
implementations (Python, Rust, service-selected) as part of the SessionRuntime
project. This commit adds the config plumbing so operators can declare their
preferred runtime via CLI or config file, laying the groundwork for future
runtime selection logic.

Added `session_runtime: RuntimeKind` as a new config field plumbed through CLI
and TOML config file layers:

- **CLI:** `--session-runtime python|rust|service-selected`
- **TOML:** `session_runtime = "rust"` in `[worker]` section
- **Default:** `python` (zero behavioral change on upgrade)

Moved `RuntimeKind` from `sessions/runtime/_kind.py` to a package-root leaf
module (`_runtime_kind.py`) to break a circular import cycle between the config
and sessions packages. `sessions.runtime` re-exports it so the public API is
unchanged.

This commit does NOT wire the field into session construction — the value is
stored on `Configuration` but not consumed yet. That wiring comes in a
follow-up.

None. The new config field defaults to `python` which is the existing behavior.
Setting `rust` or `service-selected` has no runtime effect until a future commit
wires the value into `Session.__init__`.

- Unit tests covering all config layers: CLI parsing (valid + invalid), TOML
  parsing (valid + invalid), pydantic settings field metadata, Configuration
  wiring (CLI → settings → config object). All three enum values parametrized.
- `hatch run lint` (ruff check + ruff format + mypy): clean
- `hatch run test` (2927 unit tests): all pass
- Integration tests not applicable — no integ test touches this field and no
  runtime behavior changed.

Field docstring added to `WorkerSettings`. No user-facing documentation changes
needed until the field has runtime effect.

No. Purely additive config field with a default that preserves existing behavior.

### DIFF
--- a/src/deadline_worker_agent/config/cli_args.py
+++ b/src/deadline_worker_agent/config/cli_args.py
@@ -5,6 +5,8 @@ from argparse import ArgumentParser, Namespace
 from pathlib import Path
 import os
 
+from .._session_runtime_kind import SessionRuntimeKind
+
 
 class ParsedCommandLineArguments(Namespace):
     """Represents the parsed AWS Deadline Cloud Worker Agent command-line arguments"""
@@ -28,6 +30,7 @@ class ParsedCommandLineArguments(Namespace):
     host_metrics_logging_interval_seconds: float | None = None
     structured_logs: bool | None = None
     session_root_dir: Path | None = None
+    session_runtime: SessionRuntimeKind | None = None
 
 
 def get_argument_parser() -> ArgumentParser:
@@ -168,5 +171,13 @@ def get_argument_parser() -> ArgumentParser:
         help="Path to the directory where session directories are created under.",
         default=None,
         type=Path,
+    )
+    parser.add_argument(
+        "--session-runtime",
+        help="The session runtime implementation to use.",
+        dest="session_runtime",
+        choices=[k.value for k in SessionRuntimeKind],
+        type=SessionRuntimeKind,
+        default=None,
     )
     return parser
--- a/src/deadline_worker_agent/config/config.py
+++ b/src/deadline_worker_agent/config/config.py
@@ -10,6 +10,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Optional, Sequence, Tuple, cast, TYPE_CHECKING
 
+from .._session_runtime_kind import SessionRuntimeKind
+
 from pydantic.v1 import ValidationError
 
 from openjd.sessions import PosixSessionUser, SessionUser
@@ -105,6 +107,8 @@ class Configuration:
     """Whether or not the Worker Agent logs are structured logs."""
     session_root_dir: Path
     """Path to the root directory where worker session directories are created under"""
+    session_runtime: SessionRuntimeKind
+    """The session runtime implementation to use"""
     region: Optional[str]
     """The AWS region to use. If None, boto3's default region resolution is used."""
 
@@ -130,6 +134,7 @@ class Configuration:
         "retain_session_dir",
         "structured_logs",
         "session_root_dir",
+        "session_runtime",
         "region",
     )
 
@@ -184,6 +189,8 @@ class Configuration:
             settings_kwargs["structured_logs"] = parsed_cli_args.structured_logs
         if parsed_cli_args.session_root_dir is not None:
             settings_kwargs["session_root_dir"] = parsed_cli_args.session_root_dir.absolute()
+        if parsed_cli_args.session_runtime is not None:
+            settings_kwargs["session_runtime"] = parsed_cli_args.session_runtime
 
         settings = WorkerSettings(**settings_kwargs)
 
@@ -263,6 +270,7 @@ class Configuration:
         self.retain_session_dir = settings.retain_session_dir
         self.structured_logs = settings.structured_logs
         self.session_root_dir = settings.session_root_dir
+        self.session_runtime = settings.session_runtime
 
         # Region is loaded directly from the config file as a fallback.
         # It is only used if boto3 cannot resolve a region on its own
--- a/src/deadline_worker_agent/config/config_file.py
+++ b/src/deadline_worker_agent/config/config_file.py
@@ -20,6 +20,7 @@ except ModuleNotFoundError:
     from tomli import load as load_toml, TOMLDecodeError  # type: ignore[no-redef]
 
 from ..capabilities import Capabilities
+from .._session_runtime_kind import SessionRuntimeKind
 from .errors import ConfigurationError
 from . import toml_comment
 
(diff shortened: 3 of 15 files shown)


seant-aws:
## Reviewer Guidance

### Pattern Reference

This PR adds `session_runtime` by following the exact pattern of `session_root_dir` across all four config layers. Each new block mirrors its neighbor:

| Layer | Reference (session_root_dir) | New (session_runtime) |
|-------|-------|-------|
| CLI args | [cli_args.py#L30](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/dd0a43e/src/deadline_worker_agent/config/cli_args.py#L30) | Same file, next field |
| Settings | [settings.py#L125](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/dd0a43e/src/deadline_worker_agent/config/settings.py#L125) | Same file, next field |
| Config file | [config_file.py#L199](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/dd0a43e/src/deadline_worker_agent/config/config_file.py#L199) | Same file, next field |
| Configuration | [config.py#L185-L186](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/dd0a43e/src/deadline_worker_agent/config/config.py#L185-L186) | Same pattern, next block |

### Where the new field touches (diagram)

```
Operator sets: --session-runtime rust  OR  worker.toml: session_runtime = "rust"
                     │                              │
                     ▼                              ▼
          ┌─────────────────┐            ┌──────────────────┐
          │  cli_args.py     │            │  config_file.py   │
          │  (argparse arg)  │            │  (TOML model)     │
          └────────┬─────────┘            └────────┬──────────┘
                   │ if not None                    │ if not None
                   ▼                               ▼
          ┌────────────────────────────────────────────────────┐
          │  config.py  →  settings_kwargs["session_runtime"]   │
          │                        │                            │
          │                        ▼                            │
          │              WorkerSettings (pydantic)              │
          │              merges: CLI > TOML > default           │
          │                        │                            │
          │                        ▼                            │
          │        self.session_runtime = settings.session_runtime │
          └────────────────────────────────────────────────────┘
                                   │
                                   ▼
                    Configuration.session_runtime = RuntimeKind.RUST
                    (stored — not consumed until follow-up PR)
```

### Intentional: no env var

`session_runtime` deliberately has no entry in `WorkerSettings.Config.fields`. The service-side `runtimeHint` serves as the dynamic override when in `service-selected` mode, making a per-worker env var redundant.

src/deadline_worker_agent/config/settings.py
```diff
@@ -121,6 +125,8 @@ class WorkerSettings(BaseSettings):
     host_metrics_logging_interval_seconds: float = 60
     retain_session_dir: bool = False
     structured_logs: bool = False
+    # No env var mapping — runtime selection is via config file or CLI only.
```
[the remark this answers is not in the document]
seant-aws: we discussed this during design review, the idea is that this doesn't provide much value, especially since eventually we are going to migrate everything to rust anyway

src/deadline_worker_agent/config/config_file.py
```diff
@@ -457,6 +459,8 @@ def as_settings(
             output_settings["worker_persistence_dir"] = self.worker.worker_persistence_dir
         if self.worker.session_root_dir is not None:
             output_settings["session_root_dir"] = self.worker.session_root_dir
+        if self.worker.session_runtime is not None:
+            output_settings["session_runtime"] = self.worker.session_runtime
```
epmog, line 463: do we also need to update the worker toml example?

epmog (changes requested):

src/deadline_worker_agent/config/config_file.py
seant-aws: I'm only creating the scaffolding, not wiring up the config yet for the user. I don't think this requires an example yet? I'm planning to wire it up once the rust stuff is in. Is this valid ?

epmog: This currently exposes a new worker agent CLI though? The worker toml example is more like. "here's the default settings with all of the options documented for you for what you can put into the config".

seant-aws: mm I see, I can note something for this. This is what you mean right ?

src/deadline_worker_agent/config/cli_args.py
```diff
@@ -28,6 +30,7 @@ class ParsedCommandLineArguments(Namespace):
     host_metrics_logging_interval_seconds: float | None = None
     structured_logs: bool | None = None
     session_root_dir: Path | None = None
+    session_runtime: RuntimeKind | None = None
```
seant-aws: references [cli_args.py#L30](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/dd0a43e/src/deadline_worker_agent/config/cli_args.py#L30)

src/deadline_worker_agent/config/settings.py
seant-aws, line 128: references [settings.py#L125](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/dd0a43e/src/deadline_worker_agent/config/settings.py#L125)

src/deadline_worker_agent/config/config_file.py
```diff
@@ -197,6 +198,7 @@ class WorkerConfigSection(BaseModel):
     cleanup_session_user_processes: bool = True
     worker_persistence_dir: Optional[Path] = None
     session_root_dir: Optional[Path] = None
+    session_runtime: Optional[RuntimeKind] = None
```
seant-aws: references 
[config_file.py#L199](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/dd0a43e/src/deadline_worker_agent/config/config_file.py#L199)

src/deadline_worker_agent/config/config.py
```diff
@@ -184,6 +189,8 @@ def __init__(
             settings_kwargs["structured_logs"] = parsed_cli_args.structured_logs
         if parsed_cli_args.session_root_dir is not None:
             settings_kwargs["session_root_dir"] = parsed_cli_args.session_root_dir.absolute()
+        if parsed_cli_args.session_runtime is not None:
```
seant-aws, line 192: references [config.py#L185-L186](https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/dd0a43e/src/deadline_worker_agent/config/config.py#L185-L186)

src/deadline_worker_agent/_runtime_kind.py
```diff
@@ -4,6 +4,8 @@
 
 from enum import Enum
 
+__all__ = ["RuntimeKind"]
+
 
 class RuntimeKind(str, Enum):
```
mwiebe: Could we call this SessionRuntimeKind instead of just RuntimeKind? That would be less ambiguous

mwiebe (changes requested):

mwiebe (approved):

epmog (approved):

PR #997, opened 2026-07-05
feat(mrf): Use region from log config options for cross-region CWL ro…

MarinMaksutaj:

Workers may need to write CloudWatch logs to a region different from the one they are running in. Currently the agent infers the CWL target region from local EC2 metadata, which doesn't support this.

Read the optional `region` field from `logConfiguration.options` and pass it as `region_name` when creating the CWL boto3 client. Covers both the agent-level log path and session-level log path. Falls back to local region when absent.

No change for existing fleets. When the service provides a `region` in the log configuration options, the agent will route logs to that region.

- Unit tests for parsing with/without region, and verifying `region_name` is passed to the CWL client
- Integration: custom wheel deployed via fleet host config, worker reached IDLE, job completed, logs written with 0 errors

No. Internal contract between the data plane and worker agent.

No. Backwards compatible. If `region` is absent, existing behavior is preserved.

### DIFF
--- a/src/deadline_worker_agent/aws/deadline/__init__.py
+++ b/src/deadline_worker_agent/aws/deadline/__init__.py
@@ -34,8 +34,9 @@ from ...api_models import (
     UpdateWorkerResponse,
     WorkerStatus,
 )
-from ...log_sync.cloudwatch import (
+from ...log_sync.log_constants import (
     LOG_CONFIG_OPTION_GROUP_NAME_KEY,
+    LOG_CONFIG_OPTION_REGION_KEY,
     LOG_CONFIG_OPTION_STREAM_NAME_KEY,
 )
 
@@ -131,6 +132,11 @@ class WorkerLogConfig:
 
     cloudwatch_log_stream: str
     """The name of the CloudWatch Log Stream that the Agent log should be streamed to"""
+
+    cloudwatch_region: Optional[str] = None
+    """The AWS region where the CloudWatch Log Group resides. If None, the agent's local region
+    is used. This enables satellite workers in multi-region fleets to write logs to the home
+    region's log groups."""
 
 
 def _get_error_code_from_header(response: dict[str, Any]) -> Optional[str]:
@@ -696,6 +702,7 @@ def construct_worker_log_config(log_config: LogConfiguration) -> Optional[Worker
         return WorkerLogConfig(
             cloudwatch_log_group=log_group_name,
             cloudwatch_log_stream=log_stream_name,
+            cloudwatch_region=log_config_options.get(LOG_CONFIG_OPTION_REGION_KEY),
         )
     else:
         _logger.warning(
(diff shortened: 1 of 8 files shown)


src/deadline_worker_agent/aws/deadline/__init__.py
```diff
@@ -39,6 +39,8 @@
     LOG_CONFIG_OPTION_STREAM_NAME_KEY,
 )
 
+LOG_CONFIG_OPTION_REGION_KEY = "region"
```
godobyte: Could we define `LOG_CONFIG_OPTION_REGION_KEY` in the same place as other log options keys?

src/deadline_worker_agent/sessions/log_config.py
```diff
@@ -314,7 +314,11 @@ def create_remote_handler(
         return CloudWatchHandler(
             log_group_name=log_group,
             log_stream_name=log_stream,
-            logs_client=boto_session.client("logs", config=OTHER_BOTOCORE_CONFIG),
+            logs_client=boto_session.client(
+                "logs",
+                config=OTHER_BOTOCORE_CONFIG,
+                **({"region_name": region} if (region := self.options.get("region")) else {}),
```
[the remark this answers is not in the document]
godobyte: +1 on this. The constant LOG_CONFIG_OPTION_REGION_KEY already exists, should use it here for the same reason we use LOG_CONFIG_OPTION_GROUP_NAME_KEY and LOG_CONFIG_OPTION_STREAM_NAME_KEY a few lines above.

godobyte (changes requested):

src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -130,7 +130,10 @@ def entrypoint(cli_args: Optional[list[str]] = None, *, stop: Optional[Event] =
             config=DEADLINE_BOTOCORE_CONFIG,
         )
         s3_client = session.client("s3", config=OTHER_BOTOCORE_CONFIG)
-        logs_client = session.client("logs", config=OTHER_BOTOCORE_CONFIG)
+        logs_client_kwargs: dict[str, Any] = {"config": OTHER_BOTOCORE_CONFIG}
+        if worker_bootstrap.log_config.cloudwatch_region:
+            logs_client_kwargs["region_name"] = worker_bootstrap.log_config.cloudwatch_region
+        logs_client = session.client("logs", **logs_client_kwargs)
```
godobyte, line 136: Is this part covered by unit tests?

   Worth adding two assertions in TestCloudWatchLogStreaming:

   1. When cloudwatch_region is None: session.client("logs", ...) called without region_name
   2. When cloudwatch_region is set: session.client("logs", ..., region_name="us-east-1") called with it

godobyte (dismissed):

godobyte (approved):

jericht (approved):

PR #1002, opened 2026-07-08
feat: add RustSessionRuntime adapter for OpenJD v1 sessions

seant-aws:

This change adds the Rust-backed implementation of our SessionRuntime ABC. It must be a behaviorally-faithful drop-in for the existing Python runtime, and it must keep the rest of the worker on the pydantic `v2023_09` models it already uses.

Add `RustSessionRuntime`, an adapter implementing the existing `SessionRuntime` ABC and backed by `openjd.sessions._v1`. Highlights:

- `run_task` and `enter_environment` serialize the pydantic model to the OpenJD wire shape and rebuild it natively
- `_run_task_without_session_env` (the attachment-sync path) materializes the step's embedded files to the session's files directory, resolves `Task.File.*` argument references, and runs the command with `use_session_env_vars=False`.
- The enabled openjd model extensions are wired from the session config rather than a hardcoded list.

Additive only. Nothing selects `RustSessionRuntime` at runtime yet. The worker still constructs the Python runtime, so there is no behavioral change to existing sessions. 

- New unit tests (`test/unit/sessions/runtime/test_rust.py`), passing
```
Required test coverage of 78.0% reached. Total coverage: 84.57%
=============================================================== slowest 5 durations ===============================================================
1.00s call     test/unit/test_worker.py::TestMonitorEc2Shutdown::test_spot_interruption[spot-shutdown-1-min]
1.00s call     test/unit/test_worker.py::TestMonitorEc2Shutdown::test_spot_interruption[spot-shutdown-15-sec]
1.00s call     test/unit/test_worker.py::TestMonitorEc2Shutdown::test_asg_termination
0.60s call     test/unit/startup/test_host_config_timer.py::TestHostConfigTimer::test_no_further_logs_after_timeout_reached
0.50s call     test/unit/startup/test_host_config_timer.py::TestHostConfigTimer::test_accelerates_interval_when_remaining_low
================================================== 2950 passed, 38 skipped, 8 warnings in 23.37s ==================================================
```
- Integ tests passing
```
================================================================ warnings summary =================================================================
test/integ/startup/test_host_configuration.py:116
  /local/home/stangch/github/deadline-cloud-worker-agent/test/integ/startup/test_host_configuration.py:116: PytestCollectionWarning: cannot collect test class 'TestOS' because it has a __new__ constructor (from: test/integ/startup/test_host_configuration.py)
    class TestOS(Enum):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================================================== slowest 5 durations ===============================================================
3.01s call     test/integ/startup/test_host_configuration.py::TestHostConfigurationScriptRunner::test_timer_timeout_expiration_log_appears
2.01s call     test/integ/startup/test_host_configuration.py::TestHostConfigurationScriptRunner::test_host_configuration_script_runner[Timed out test case]
1.01s call     test/integ/startup/test_host_configuration.py::TestHostConfigurationScriptRunner::test_timer_ticker_logs_appear_during_script_execution
0.69s call     test/integ/config/test_config_cli.py::TestMissingExistingCommented::test[ModifiableSetting.FARM_ID-missing]
0.69s call     test/integ/config/test_config_cli.py::TestMissingExistingCommented::test[ModifiableSetting.FLEET_ID-missing]
============================================== 84 passed, 4 skipped, 1 xfailed, 1 warning in 24.53s ===============================================

```
- Ran E2E tests, hardcoding testing with `SessionRuntimeKind.RUST`
```
617.10s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_worker_shuts_down_host_machine_if_configured
340.63s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_worker_local_session_logs_can_be_turned_off
298.11s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_worker_requires_no_instance_profile
293.99s setup    test/e2e/test_override_job_user.py::TestLinuxJobUserOverride::test_no_user_override
254.97s call     test/e2e/test_worker_config.py::TestWorkerConfiguration::test_session_root_dir
=========================== short test summary info ============================
FAILED test/e2e/test_cap_kill.py::test_worker_only_has_cap_kill - AssertionEr...
============ 1 failed, 48 passed, 23 skipped in 4614.16s (1:16:54) =============
```
One remaining failure to be fixed

No customer facing changes yet.

No. The change is purely additive and no code path constructs `RustSessionRuntime` yet, so runtime behavior is unchanged.

### DIFF
--- a/src/deadline_worker_agent/sessions/runtime/rust.py
+++ b/src/deadline_worker_agent/sessions/runtime/rust.py
@@ -1,0 +1,420 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from __future__ import annotations
+
+import os
+import re
+import stat
+import tempfile
+from datetime import timedelta
+from logging import getLogger
+from pathlib import Path
+from shutil import chown
+from typing import TYPE_CHECKING, Any, Optional
+
+from openjd._openjd_rs import create_environment, deserialize_step
+from openjd.expr import PathFormat, PathMappingRule as RustPathMappingRule
+from openjd.model._v1 import decode_environment_template
+from openjd.model._v1.types import (
+    JobParameterType,
+    JobParameterValue,
+    ModelExtension,
+    ModelProfile,
+    SpecificationRevision,
+    TaskParameterType,
+    TaskParameterValue,
+)
+from openjd.sessions import (
+    ActionState,
+    ActionStatus,
+    PathMappingRule,
+    PosixSessionUser,
+    WindowsSessionUser,
+)
+from openjd.sessions._v1 import (
+    ActionState as RustActionState,
+    ActionStatus as RustActionStatus,
+    PosixSessionUser as RustPosixSessionUser,
+    Session as OpenJDRustSession,
+    WindowsSessionUser as RustWindowsSessionUser,
+)
+
+from . import SessionRuntime, SessionRuntimeConfig
+
+if TYPE_CHECKING:
+    from openjd.sessions import (
+        EnvironmentIdentifier,
+        EnvironmentModel,
+        SessionUser,
+        StepScriptModel,
+    )
+
+__all__ = ["RustSessionRuntime"]
+
+logger = getLogger(__name__)
+
+
+# Deliberate allowlist of the OpenJD specification revisions this adapter
+# supports. Adding a revision is not just adding a map entry: the adapter
+# hardcodes 2023-09 wire strings elsewhere (e.g. enter_environment's
+# "environment-2023-09" envelope), which must be audited for a new revision.
+_SPEC_REVISIONS: dict[str, SpecificationRevision] = {
+    SpecificationRevision.v2023_09.value: SpecificationRevision.v2023_09,
+}
+
+
+# openjd.expr.PathFormat is a non-constructable Rust enum, so map the v0
+# rule's format string to the member by its value.
+_PATH_FORMATS: dict[str, PathFormat] = {
+    "POSIX": PathFormat.POSIX,
+    "WINDOWS": PathFormat.WINDOWS,
+}
+
+
+# The _v1 ActionState is a non-iterable pyo3 enum, so map each member to the
+# v0 enum explicitly (same pattern as _SPEC_REVISIONS/_PATH_FORMATS). A state
+# added on the Rust side must be mapped here deliberately — failing loud beats
+# silently mis-reporting an action's state to the service.
+_ACTION_STATES: dict[str, ActionState] = {
+    str(RustActionState.RUNNING): ActionState.RUNNING,
+    str(RustActionState.CANCELED): ActionState.CANCELED,
+    str(RustActionState.TIMEOUT): ActionState.TIMEOUT,
+    str(RustActionState.FAILED): ActionState.FAILED,
+    str(RustActionState.SUCCESS): ActionState.SUCCESS,
+}
+
+
+def _to_v0_action_state(state: RustActionState) -> ActionState:
+    """Convert a _v1 ActionState into the v0 enum member, failing loud on
+    states this adapter does not recognize."""
+    try:
+        return _ACTION_STATES[str(state)]
+    except KeyError:
+        raise ValueError(f"Unrecognized _v1 ActionState: {state!r}") from None
+
+
+def _to_rust_session_user(
+    user: Optional[SessionUser],
+) -> Optional[RustPosixSessionUser | RustWindowsSessionUser]:
+    """Convert a worker (v0) SessionUser into the equivalent _v1 type.
+
+    The worker builds v0 ``openjd.sessions`` user objects, but the Rust session
+    only accepts its own ``openjd.sessions._v1`` user classes — structurally
+    identical but distinct at the type level. None passes through unchanged
+    (the session runs as the agent's own user in that case).
+    """
+    if user is None:
+        return None
+    if isinstance(user, PosixSessionUser):
+        return RustPosixSessionUser(user.user, group=user.group)
+    if isinstance(user, WindowsSessionUser):
+        return RustWindowsSessionUser(
+            user.user, password=user.password, logon_token=user.logon_token
+        )
+    raise TypeError(f"Unsupported SessionUser type: {type(user).__name__}")
+
+
+def _to_rust_parameter_values(
+    values: dict[str, Any],
+    *,
+    value_cls: type,
+    type_cls: type,
+) -> dict[str, Any]:
+    """Convert worker (v0) parameter values into native _v1 parameter values.
+
+    The worker supplies ``openjd.model`` ParameterValue objects; the Rust
+    session accepts the native pyo3 value types exported from
+    ``openjd.model._v1.types``, which are constructable from Python. Values
+    already in the ``{"type", "value"}`` dict form pass through unchanged.
+    Parameter types the binding does not define fail loud rather than being
+    silently re-encoded.
+    """
+    converted: dict[str, Any] = {}
+    for name, value in values.items():
+        if isinstance(value, dict):
+            converted[name] = value
+            continue
+        native_type = getattr(type_cls, value.type.value, None)
+        if native_type is None:
+            raise ValueError(
+                f"Parameter {name!r} has type {value.type.value!r}, which the "
+                f"_v1 binding's {type_cls.__name__} does not define."
+            )
+        converted[name] = value_cls(type=native_type, value=value.value)
+    return converted
+
+
+def _to_rust_job_parameter_values(values: dict[str, Any]) -> dict[str, Any]:
+    """Convert worker (v0) job parameter values into native _v1 JobParameterValues."""
+    return _to_rust_parameter_values(values, value_cls=JobParameterValue, type_cls=JobParameterType)
+
+
+def _to_rust_task_parameter_values(values: dict[str, Any]) -> dict[str, Any]:
+    """Convert worker (v0) task parameter values into native _v1 TaskParameterValues."""
+    return _to_rust_parameter_values(
+        values, value_cls=TaskParameterValue, type_cls=TaskParameterType
+    )
+
+
+def _to_rust_path_mapping_rule(rule: PathMappingRule) -> RustPathMappingRule:
+    """Convert one worker (v0) PathMappingRule into the _v1 (openjd.expr) type.
+
+    The worker builds ``openjd.sessions`` path-mapping rules, but the Rust
+    session only accepts ``openjd.expr.PathMappingRule`` — a distinct class with
+    the same shape. Paths are coerced to strings, which the _v1 rule stores.
+    """
+    return RustPathMappingRule(
+        source_path_format=_PATH_FORMATS[rule.source_path_format.value],
+        source_path=str(rule.source_path),
+        destination_path=str(rule.destination_path),
+    )
+
+
+def _to_rust_path_mapping_rules(
+    rules: Optional[list[PathMappingRule]],
+) -> Optional[list[RustPathMappingRule]]:
+    """Convert worker (v0) PathMappingRule lists into the _v1 type, passing None through."""
+    if rules is None:
+        return None
+    return [_to_rust_path_mapping_rule(rule) for rule in rules]
+
+
+def _to_v0_action_status(status: RustActionStatus) -> ActionStatus:
+    """Convert a _v1 ActionStatus back into the v0 type the worker consumes.
+
+    The Rust session reports status with its own ``openjd.sessions._v1``
+    ActionStatus/ActionState, but the worker's scheduler compares against the v0
+    ActionState enum. The state is remapped through the explicit _ACTION_STATES
+    table and the remaining fields are copied across.
+    """
+    return ActionStatus(
+        state=_to_v0_action_state(status.state),
+        progress=status.progress,
+        status_message=status.status_message,
+        fail_message=status.fail_message,
+        exit_code=status.exit_code,
+    )
+
+
+class RustSessionRuntime(SessionRuntime):
+    """SessionRuntime backed by openjd.sessions._v1 (Rust implementation)."""
+
+    _session: OpenJDRustSession
+    _user: Optional[SessionUser]
+
+    def __init__(self, config: SessionRuntimeConfig) -> None:
+        try:
+            revision = _SPEC_REVISIONS[config.spec_revision]
+        except KeyError:
+            raise ValueError(
+                f"Unsupported OpenJD specification revision: {config.spec_revision!r}"
+            ) from None
+
+        # The extensions to enable are supplied by the config as strings
+        # (sourced from the OpenJD model library). The v1 API needs native
+        # ModelExtension enums, so convert each name to its enum member.
+        #
+        # The Python openjd.model library may know extensions the Rust crate
+        # doesn't yet (e.g. WRAP_ACTIONS). Skip those rather than fail: a job
+        # template that actually requires a skipped extension is rejected at
+        # decode time with a clear error, matching how the v0 session
+        # tolerates extension names it doesn't recognize.
+        extensions: list[ModelExtension] = []
+        for name in config.supported_extensions:
+            extension = ModelExtension.from_str(name)
+            if extension is None:
+                logger.warning(
+                    "OpenJD model extension %r is not supported by the Rust session runtime; ignoring it.",
+                    name,
+                )
+                continue
+            extensions.append(extension)
+
+        # Kept for the attachment-sync path: on POSIX it grants the files it
+        # writes group-read access for the session user's group, so the job
+        # (which runs as that user) can read them.
+        self._user = config.user
+
+        # The _v1 session reports status with _v1 ActionStatus/ActionState, but
+        # the worker's callback expects the v0 types, so translate on the way out.
+        v0_callback = config.action_callback
+
+        def _rust_action_callback(session_id: str, status: RustActionStatus) -> None:
+            v0_callback(session_id, _to_v0_action_status(status))
+
+        self._session = OpenJDRustSession(
+            session_id=config.session_id,
+            job_parameter_values=_to_rust_job_parameter_values(config.job_parameter_values),
+            path_mapping_rules=_to_rust_path_mapping_rules(config.path_mapping_rules),
+            retain_working_dir=config.retain_working_dir,
+            user=_to_rust_session_user(config.user),
+            callback=_rust_action_callback,
+            os_env_vars=config.os_env_vars,
+            session_root_directory=config.session_root_directory,
+            profile=ModelProfile(revision=revision, extensions=extensions),
+        )
+
+    def enter_environment(
+        self,
+        *,
+        environment: EnvironmentModel,
+        identifier: Optional[EnvironmentIdentifier] = None,
+        os_env_vars: Optional[dict[str, str]] = None,
+    ) -> EnvironmentIdentifier:
+        # The shared action layer hands a pydantic v2023_09 environment, but the
+        # Rust session needs a native _v1 environment. Serialize to the OpenJD
+        # wire shape and rebuild it natively. exclude_none=True is required, not
+        # cosmetic: the Rust decoder rejects explicit nulls (OpenJD treats
+        # absent and null as equivalent).
+        native_environment = create_environment(
+            decode_environment_template(
+                {
+                    "specificationVersion": "environment-2023-09",
+                    "environment": environment.model_dump(
+                        mode="json", by_alias=True, exclude_none=True
+                    ),
+                }
+            )
+        )
+        return self._session.enter_environment(
+            environment=native_environment,
+            identifier=identifier,
+            os_env_vars=os_env_vars,
+        )
+
+    def exit_environment(
+        self,
+        *,
+        identifier: EnvironmentIdentifier,
+        os_env_vars: Optional[dict[str, str]] = None,
+        keep_session_running: bool = False,
+    ) -> None:
+        self._session.exit_environment(
+            identifier=identifier,
+            os_env_vars=os_env_vars,
+            keep_session_running=keep_session_running,
+        )
+
+    def run_task(
+        self,
+        *,
+        step_script: StepScriptModel,
+        task_parameter_values: dict[str, Any],
+        os_env_vars: Optional[dict[str, str]] = None,
+        log_task_banner: bool = True,
+    ) -> None:
+        # The shared action layer hands a pydantic v2023_09 StepScript, but the
+        # Rust session needs a native _v1 step. Serialize to the OpenJD wire
+        # shape and rebuild it natively. deserialize_step requires a named step
+        # and the binding has no bare step-script deserializer (tracked in
+        # OpenJobDescription/openjd-sessions-for-python#334), so wrap the script
+        # as ``{"name": ..., "script": ...}`` and unwrap it after decoding.
+        # exclude_none=True is required, not cosmetic: the Rust decoder rejects
+        # explicit nulls (OpenJD treats absent and null as equivalent).
+        native_step_script = deserialize_step(
+            {
+                "name": "Placeholder",
+                "script": step_script.model_dump(mode="json", by_alias=True, exclude_none=True),
+            }
+        ).script
+        self._session.run_task(
+            step_script=native_step_script,
+            task_parameter_values=_to_rust_task_parameter_values(task_parameter_values),
+            os_env_vars=os_env_vars,
+            log_task_banner=log_task_banner,
+        )
+
+    def _run_task_without_session_env(
+        self,
+        *,
+        step_script: StepScriptModel,
+        task_parameter_values: dict[str, Any],
+        os_env_vars: Optional[dict[str, str]] = None,
+        log_task_banner: bool = True,
+    ) -> None:
+        # Attachment-sync path: the native run_task's embedded-file handling is
+        # unavailable here, so materialize the script's embedded files to the
+        # session's files directory ourselves, resolve ``Task.File.*`` argument
+        # references to those paths, and run the command as a bare subprocess.
+        #
+        # STOPGAP — tracked in OpenJobDescription/openjd-sessions-for-python#332.
+        # Remove this block and delegate once the _v1 wrapper exposes a
+        # ``run_task(..., use_session_env_vars=False)``-style API (which would
+        # also restore the library's cross-platform file permission handling).
+        file_paths: dict[str, str] = {}
+        if step_script.embeddedFiles:
+            for embedded_file in step_script.embeddedFiles:
+                if embedded_file.data is None:
+                    continue
+                fd, path = tempfile.mkstemp(
+                    dir=str(self._session.files_directory),
+                    prefix=f"{embedded_file.name}_",
+                )
+                # UTF-8 explicitly: the platform default (e.g. cp1252 on
+                # Windows) can corrupt or reject non-ASCII paths in the
+                # manifest JSON, which the reader script decodes as UTF-8.
+                with os.fdopen(fd, "w", encoding="utf-8") as f:
+                    f.write(str(embedded_file.data))
+                # Owner read/write. On POSIX, also grant the session user's group
+                # read access so the job user can read the materialized file.
+                mode = stat.S_IRUSR | stat.S_IWUSR
+                if self._user is not None and os.name == "posix":
+                    group = getattr(self._user, "group", None)
+                    if group is not None:
+                        chown(path, group=group)
+                        mode |= stat.S_IRGRP
+                os.chmod(path, mode)
+                file_paths[embedded_file.name] = path
+
+        command = str(step_script.actions.onRun.command)
+        args: list[str] = []
+        if step_script.actions.onRun.args:
+            for arg in step_script.actions.onRun.args:
+                # Resolve {{ Task.File.<name> }} references with any interior
+                # whitespace, matching the format-string parser's tolerance.
+                # References to files that were not materialized (unknown name
+                # or no data) are left untouched.
+                resolved = re.sub(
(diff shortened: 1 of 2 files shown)


src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,223 @@
...
+
+import os
+import stat
+import tempfile
+from datetime import timedelta
+from pathlib import Path
+from shutil import chown
+from typing import TYPE_CHECKING, Any, Optional
+
+from openjd._openjd_rs import create_environment, deserialize_step
```
leongdl, line 15: One comment about imports. Now that we have the ABC and importing both the V0 and V1,  and given Python's interpreter based imports we may want to delay imports here and also the python session so we can get clean imports into the namespace.

EG: Import ONLY if we create a rust session.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,223 @@
...
+        task_parameter_values: dict[str, Any],
+        os_env_vars: Optional[dict[str, str]] = None,
+        log_task_banner: bool = True,
+    ) -> None:
+        # Attachment-sync path: the native run_task's embedded-file handling is
+        # unavailable here, so materialize the script's embedded files to the
+        # session's files directory ourselves, resolve ``Task.File.*`` argument
+        # references to those paths, and run the command as a bare subprocess.
+        file_paths: dict[str, str] = {}
+        if step_script.embeddedFiles:
```
leongdl, line 345: This seems fishy - are we by passing OpenJD Sessions modules / crates? I remember the file materialization is part of it and should be the responsibility of the crate instead of worker agent.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,223 @@
...
+                file_paths[embedded_file.name] = path
+
+        command = str(step_script.actions.onRun.command)
+        args: list[str] = []
+        if step_script.actions.onRun.args:
+            for arg in step_script.actions.onRun.args:
+                resolved = str(arg)
+                for name, path in file_paths.items():
+                    # Resolve both the spaced and unspaced reference spellings.
+                    resolved = resolved.replace("{{ Task.File." + name + " }}", path)
```
leongdl: This logic should be in OpenJD core crates too, seems awkward to have it in worker agent.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl, line 15: Following up on my own import comment — I dug into the import graph and this is already handled. `_factory.create_session_runtime` imports both adapters lazily (`from .rust import RustSessionRuntime` only executes when a RUST runtime is requested, with `ImportError` mapped to `NotImplementedError`), and `runtime/__init__.py` doesn't import either adapter module. The v0 `openjd.sessions` imports in this file are all under `TYPE_CHECKING`, so creating a PYTHON session never loads the Rust binding, and vice versa.

One residual nit that can't be fixed in this repo: importing `openjd.sessions._v1` still executes the parent `openjd.sessions` package `__init__`, which pulls in the v0 implementation. A fully clean namespace would need a restructure on the openjd side. Consider my original ask resolved.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl, line 345: Confirming my earlier concern after reading both openjd implementations — yes, this bypasses the library. In v0, `_run_task_without_session_env` lives *inside* openjd-sessions: it hands the full StepScript to `StepScriptRunner`, which materializes embedded files via `write_file_for_user` (handling both POSIX perms and Windows ACLs) and resolves format strings through the symbol table. `PythonSessionRuntime` just delegates.

The `_v1` wrapper only exposes `run_task` and `run_subprocess`, and `run_subprocess` explicitly builds a StepScript with `embeddedFiles=None` and no format-string substitution — so the capability genuinely doesn't exist there yet.

Suggestion: add a supported "run task without session env" API to the `_v1` wrapper (or the Rust crate) — e.g. `run_task(..., use_session_env_vars=False)` — and have this adapter delegate, mirroring the v0 shape. Worth noting the two bot findings on this PR (missing Windows ACLs, platform-default text encoding) are both bugs that openjd's `write_file_for_user` already solves, which is strong evidence this logic belongs down in the library. If we keep this as a stopgap to unblock the PR, let's file an issue on openjd-sessions-for-python and link it in a comment here.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl: Refining my earlier comment: this belongs in OpenJD **Sessions** rather than Model — `{{ Task.File.* }}` resolution is runtime format-string evaluation against the session's symbol table, which v0 does inside `StepScriptRunner`. The model library only defines the template types.

One concrete correctness gap with the naive `str.replace`: it only handles two spacing spellings (`{{ Task.File.X }}` and `{{Task.File.X}}`), while the real format-string parser accepts other whitespace variants like `{{ Task.File.X}}`. So this isn't just awkward placement — it can silently leave references unresolved.

Same suggestion as my other thread: push this down into the `_v1` wrapper / Rust crate so the format-string engine does the resolution, and treat this code as a documented stopgap with a linked issue.

leongdl (commented):
Follow-ups on my three earlier comments after digging through the import graph, the v0 `openjd.sessions.Session`, and the `_v1` wrapper. Details inline.

seant-aws:
## E2E status — Rust session runtime (Linux, full suite)

All runs executed with the worker agent forced onto the Rust (OpenJD v1) session runtime,
against locally built, checksum-pinned wheels from the fix branches.

| Run | Result | Notes |
|---|---|---|
| Baseline (before fixes) | **16 failed** / 33 passed / 23 skipped | type-conversion crashes at the runtime boundary, job-attachment failures, all cancel scenarios stuck `CANCELING`, plus cascade failures on wedged workers |
| After boundary + task-parameter conversion fixes | attachments bucket **14/14** | `Cannot convert ParameterValue to ExprValue` eliminated |
| After cross-user cancel routing fix (openjd-rs #258 + openjd-model #312 + openjd-sessions #331) | cancel bucket **6/6** | cancels now land in ~40 ms; canceled actions report `CANCELED` (previously ran to completion and reported `SUCCEEDED`) |
| **LATEST RESULTS** | **1 failed / 48 passed / 23 skipped** (1:16:54) | sole failure: `test_worker_only_has_cap_kill` — pre-existing host-capabilities issue, unrelated to the session runtime |

Fixes validated end-to-end:
- OpenJD v1 type conversions at the runtime boundary (session user, action status, job/task parameters, path mapping)
- Mid-action cancellation delivery to cross-user helper subprocesses, including `time_limit` grace handling and correct `Canceled` classification (paired changes: OpenJobDescription/openjd-rs#258, OpenJobDescription/openjd-model-for-python#312, OpenJobDescription/openjd-sessions-for-python#331)
- Unsupported model extensions are warned and skipped instead of failing session start

src/deadline_worker_agent/sessions/runtime/rust.py
seant-aws: Yea totally fair, I brougth this up a bit last week, in which Mark had some guidance on this part, which was to follo whis `bindings-rs` prototype (https://github.com/aws-deadline/deadline-cloud-worker-agent/compare/mainline...mwiebe:deadline-cloud-worker-agent:bindings-rs), which we are following as the reference for driving the `_v1` session from the worker agent. 

Originally I did approach it at this angle. I'll file this as an issue since this seems a bit cleaner

Filed OpenJobDescription/openjd-sessions-for-python#332 proposing a supported `run_task(..., use_session_env_vars=False)`-style API on the `_v1` wrapper so this adapter can delegate — treating this code as a documented stopgap until then. The encoding and whitespace-resolution findings on this block are fixed in c8e390a; the Windows ACL gap is called out in the issue as a further reason to push down rather than reimplement.

src/deadline_worker_agent/sessions/runtime/rust.py
seant-aws: Good catch on the whitespace variants — fixed in c8e390a with a whitespace-tolerant pattern (references to unmaterialized files still pass through untouched, preserving the existing behavior for data-less embedded files).

Agreed the resolution itself belongs in the sessions library's format-string engine; that is covered by the same tracking issue: OpenJobDescription/openjd-sessions-for-python#332.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,223 @@
...
+        #
+        # Divergence from PythonSessionRuntime: the v0 session accepts unknown
+        # extension strings silently. The v1 API can't represent an unknown
+        # one, so an unrecognized name fails loud here instead of being dropped.
+        extensions: list[ModelExtension] = []
+        for name in config.supported_extensions:
+            extension = ModelExtension.from_str(name)
+            if extension is None:
+                raise ValueError(f"Unknown OpenJD model extension: {name!r}")
+            extensions.append(extension)
```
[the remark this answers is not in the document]
seant-aws: Addressed in the current revision: unsupported extensions are now logged and skipped rather than failing construction, matching the v0 session's tolerance of unrecognized extension names. A job template that actually requires a skipped extension is still rejected at decode time with a clear error.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,223 @@
...
+        if step_script.embeddedFiles:
+            for embedded_file in step_script.embeddedFiles:
+                if embedded_file.data is None:
+                    continue
+                fd, path = tempfile.mkstemp(
+                    dir=str(self._session.files_directory),
+                    prefix=f"{embedded_file.name}_",
+                )
+                with os.fdopen(fd, "w") as f:
+                    f.write(str(embedded_file.data))
```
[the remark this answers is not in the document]
seant-aws: Fixed in c8e390a — the materialized file is now written with explicit `encoding="utf-8"`.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,223 @@
...
+                # Owner read/write. On POSIX, also grant the session user's group
+                # read access so the job user can read the materialized file.
+                mode = stat.S_IRUSR | stat.S_IWUSR
+                if self._user is not None and os.name == "posix":
+                    group = getattr(self._user, "group", None)
+                    if group is not None:
+                        chown(path, group=group)
+                        mode |= stat.S_IRGRP
+                os.chmod(path, mode)
```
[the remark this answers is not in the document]
seant-aws: Known gap in this stopgap — tracked in OpenJobDescription/openjd-sessions-for-python#332, which proposes pushing embedded-file materialization down into the `_v1` wrapper where `write_file_for_user` already handles both POSIX permissions and Windows ACLs. The Rust session runtime is not yet enabled on Windows, so this does not affect any supported configuration in this PR.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,373 @@
...
+    ``openjd.model`` ParameterValue objects, but the Rust session accepts either
+    native TaskParameterValue objects or ``{"type", "value"}`` dicts. Each value
+    is rebuilt as the equivalent dict; values already in dict form pass through.
+    """
+    converted: dict[str, Any] = {}
+    for name, value in values.items():
+        if isinstance(value, dict):
+            converted[name] = value
+        else:
+            converted[name] = {"type": value.type.value, "value": value.value}
```
leongdl: Can we use a py03 type somehow? This type and value here maybe hard to maintain long term.

leongdl: Similar to all the bridge functions below.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,373 @@
...
+    )
+
+__all__ = ["RustSessionRuntime"]
+
+logger = getLogger(__name__)
+
+
+# SpecificationRevision can't be constructed from a string, so map the
+# config's revision string to the enum member by its value.
+_SPEC_REVISIONS: dict[str, SpecificationRevision] = {
```
leongdl, line 61: I wonder if we can iterate on SpecificationRevision and then not have to update in the future?

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,373 @@
...
+        log_task_banner: bool = True,
+    ) -> None:
+        # The shared action layer hands a pydantic v2023_09 StepScript, but the
+        # Rust session needs a native _v1 step. Serialize to the OpenJD wire
+        # shape and rebuild it natively. deserialize_step requires a named step,
+        # so wrap the bare script as ``{"name": ..., "script": ...}`` (the same
+        # wrapping the job-entity step decode uses).
+        native_step_script = deserialize_step(
+            {
+                "name": "Placeholder",
```
leongdl, line 316: This seems odd?

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,373 @@
...
+        # Attachment-sync path: the native run_task's embedded-file handling is
+        # unavailable here, so materialize the script's embedded files to the
+        # session's files directory ourselves, resolve ``Task.File.*`` argument
+        # references to those paths, and run the command as a bare subprocess.
+        file_paths: dict[str, str] = {}
+        if step_script.embeddedFiles:
+            for embedded_file in step_script.embeddedFiles:
+                if embedded_file.data is None:
+                    continue
+                fd, path = tempfile.mkstemp(
```
leongdl, line 349: This seems odd to have worker agent write the embedded file. It should be in OpenJD sessions crate?

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl: Adding evidence to my earlier question — this is more actionable than I first thought. The docstring says "JobParameterValue is not exported from the _v1 package", but `JobParameterValue`, `TaskParameterValue`, `JobParameterType`, and `TaskParameterType` are all exported pyo3 types from `openjd.model._v1.types` — the exact module this file already imports `ModelExtension`/`ModelProfile`/`SpecificationRevision` from and constructs `ModelProfile` from directly.

So if `JobParameterValue`/`TaskParameterValue` are constructable from Python (worth checking — some pyo3 classes in the binding aren't, per the `_PATH_FORMATS` comment about `PathFormat`), we should build the native types here instead of hand-encoding the `{"type", "value"}` wire shape. If they're *not* constructable, the docstring should say that instead — the current justification is inaccurate.

Smaller nit: `_to_rust_job_parameter_values` and `_to_rust_task_parameter_values` have identical bodies — one shared helper would halve the surface my maintainability concern applies to.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl, line 61: Answering my own question after checking the binding: probably not worth it, and the explicit map may actually be the right call. Two reasons:

1. The pyo3 enums aren't Python `Enum` subclasses, so `{m.value: m for m in SpecificationRevision}` only works if the binding implements iteration.
2. More importantly, this adapter has revision-specific logic hardcoded elsewhere — `enter_environment` serializes with a literal `"specificationVersion": "environment-2023-09"`. Auto-accepting future revisions from the enum would claim support the adapter doesn't actually have.

Suggestion: keep the map, but reframe the comment to say it's a deliberate allowlist of revisions this adapter supports (and that adding one requires auditing the hardcoded 2023-09 wire strings), rather than a workaround for enum construction.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl, line 316: Expanding on my "seems odd": the root cause is a binding API gap — `_openjd_rs` exposes `deserialize_step` (which requires a full named Step) but no direct step-*script* deserializer, so we fabricate a synthetic step and pull `.script` off it. Functional, but the fake `"Placeholder"` name could leak into Rust-side validation error messages and confuse whoever reads the session log.

Suggestion: request a `deserialize_step_script`-style entry point in the binding — could ride along on OpenJobDescription/openjd-sessions-for-python#332 or a sibling issue — and note the issue link in this comment block so the wrapping is clearly a tracked workaround.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl, line 349: Retracting the question half of this — I see @seant-aws already agreed on the earlier-revision thread and filed OpenJobDescription/openjd-sessions-for-python#332 proposing a `run_task(..., use_session_env_vars=False)`-style API on the `_v1` wrapper, with this block documented as a stopgap (and the Windows ACL gap noted; Rust runtime isn't enabled on Windows yet). That resolves my concern for this PR — consider adding the issue link as a code comment on this block so the stopgap is discoverable without the PR history.

leongdl (commented):
Follow-ups on my comments against the current revision (c8e390a), after verifying against the `openjd.model._v1` / `openjd.sessions._v1` sources. Inline details below. Two additional non-inline notes:

1. `extend_path_mapping_rules` uses `assert converted is not None` for control flow — asserts are stripped under `python -O`, which would silently pass `None` through. An explicit check (or a non-Optional conversion helper for this call site) would be safer.
2. The `exclude_none=True` added to both `model_dump` calls looks correct (the Rust decoder presumably rejects explicit nulls, and OpenJD templates don't distinguish null-vs-absent), but it's an easy future "cleanup" regression — worth a one-line code comment explaining why it's required.

Overall the direction is good: with #332 tracking the layering items, the remaining asks on this PR are the type-conversion hardening below.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,373 @@
...
+    """Convert a _v1 ActionStatus back into the v0 type the worker consumes.
+
+    The Rust session reports status with its own ``openjd.sessions._v1``
+    ActionStatus/ActionState, but the worker's scheduler compares against the v0
+    ActionState enum. The v0 ActionState is a string enum whose values match the
+    _v1 state names, so the state is remapped by value and the remaining fields
+    are copied across.
+    """
+    return ActionStatus(
+        state=ActionState(str(status.state)),
```
leongdl: +1 to the bot's finding here — this is the one remaining conversion I'd want hardened before merge. It's also internally inconsistent with this file's own conventions: `_SPEC_REVISIONS` and `_PATH_FORMATS` use explicit value→member lookup tables precisely because the Rust enums are awkward, yet the hottest-path conversion (runs on every status poll and every callback) relies on `str()` output matching a v0 enum value verbatim. An explicit state map that fails loudly on an unrecognized state would match the file's pattern and turn a silent cross-binding drift into a clear error. The unit test mocking `__str__` to return `"running"` validates the assumption against itself, so a test against the real binding enum would also help.

src/deadline_worker_agent/sessions/runtime/rust.py
mwiebe: We can also update the bindings to support the iterator protocol for that style of code to work.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl: Sounds great!

src/deadline_worker_agent/sessions/runtime/rust.py
godobyte: +1

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,373 @@
...
+    ) -> EnvironmentIdentifier:
+        # The shared action layer hands a pydantic v2023_09 environment, but the
+        # Rust session needs a native _v1 environment. Serialize to the OpenJD
+        # wire shape and rebuild it natively.
+        native_environment = create_environment(
+            decode_environment_template(
+                {
+                    "specificationVersion": "environment-2023-09",
+                    "environment": environment.model_dump(
+                        mode="json", by_alias=True, exclude_none=True
```
godobyte, line 274: Nit: `model_dump(..., exclude_none=True)` in enter_environment and run_task. I checked the OpenJD spec models and every Optional field uses None exclusively to mean "absent" (never as a distinct semantic value), so this is safe. Worth a one-line comment noting why `exclude_none=True` is needed (Rust serde rejects explicit null fields) so a future reader doesn't remove it thinking it's cosmetic.

src/deadline_worker_agent/sessions/runtime/rust.py
seant-aws: yea this is a good one, ill make sure to add to my steering

seant-aws: noted,  added

```
native_type = getattr(type_cls, value.type.value, None)
        if native_type is None:
            raise ValueError(
                f"Parameter {name!r} has type {value.type.value!r}, which the "
                f"_v1 binding's {type_cls.__name__} does not define."
            )
        converted[name] = value_cls(type=native_type, value=value.value)
        ```

Comment (seant-aws) [reply src/deadline_worker_agent/sessions/runtime/rust.py]: replied to first comment related to this, doing:

```
        native_type = getattr(type_cls, value.type.value, None)
        if native_type is None:
            raise ValueError(
                f"Parameter {name!r} has type {value.type.value!r}, which the "
                f"_v1 binding's {type_cls.__name__} does not define."
            )
        converted[name] = value_cls(type=native_type, value=value.value)
```

Comment (seant-aws) [reply src/deadline_worker_agent/sessions/runtime/rust.py]: also refactored `_to_rust_job_parameter_values` and `_to_rust_task_parameter_values` to use the same `_to_rust_parameter_values` helper

Comment (seant-aws) [reply src/deadline_worker_agent/sessions/runtime/rust.py]: also addressed your point 1, cleaned up the assert and made it mor3e straightforward

File: src/deadline_worker_agent/sessions/runtime/rust.py (hunk 8 above)
Comment (seant-aws) [reply src/deadline_worker_agent/sessions/runtime/rust.py:61]: done, readded comment

File: src/deadline_worker_agent/sessions/runtime/rust.py (hunk 9 above)
Comment (seant-aws) [reply src/deadline_worker_agent/sessions/runtime/rust.py:316]: https://github.com/OpenJobDescription/openjd-sessions-for-python/issues/334 I have it here, I'll track these

File: src/deadline_worker_agent/sessions/runtime/rust.py (hunk 10 above)
Comment (seant-aws) [reply src/deadline_worker_agent/sessions/runtime/rust.py:349]: yup, will address this followip

File: src/deadline_worker_agent/sessions/runtime/rust.py (hunk 12 above)
Comment (seant-aws) [reply src/deadline_worker_agent/sessions/runtime/rust.py:274]: included with David's comment above as well

```
# ... exclude_none=True is required, not
# cosmetic: the Rust decoder rejects explicit nulls (OpenJD treats
# absent and null as equivalent).
```

File: src/deadline_worker_agent/sessions/runtime/rust.py (hunk 11 above)
Comment (seant-aws) [reply src/deadline_worker_agent/sessions/runtime/rust.py]: done:

```
# The _v1 ActionState is a non-iterable pyo3 enum, so map each member to the
# v0 enum explicitly (same pattern as _SPEC_REVISIONS/_PATH_FORMATS). A state
# added on the Rust side must be mapped here deliberately — failing loud beats
# silently mis-reporting an action's state to the service.
_ACTION_STATES: dict[str, ActionState] = {
    str(RustActionState.RUNNING): ActionState.RUNNING,
    str(RustActionState.CANCELED): ActionState.CANCELED,
    str(RustActionState.TIMEOUT): ActionState.TIMEOUT,
    str(RustActionState.FAILED): ActionState.FAILED,
    str(RustActionState.SUCCESS): ActionState.SUCCESS,
}


def _to_v0_action_state(state: RustActionState) -> ActionState:
    """Convert a _v1 ActionState into the v0 enum member, failing loud on
    states this adapter does not recognize."""
    try:
        return _ACTION_STATES[str(state)]
    except KeyError:
        raise ValueError(f"Unrecognized _v1 ActionState: {state!r}") from None


```

**Comment by seant-aws** [comment]
https://github.com/aws-deadline/deadline-cloud-worker-agent/compare/c8e390ab66d5156e9a7c0cce283f2b286dfc8c77..33e16b99034fe91ebc81524337ca4dd029b4cc53

this is the link for the diff

File: src/deadline_worker_agent/sessions/runtime/rust.py [hunk 13]
```diff
@@ -0,0 +1,420 @@
...
+                mode = stat.S_IRUSR | stat.S_IWUSR
+                if self._user is not None and os.name == "posix":
+                    group = getattr(self._user, "group", None)
+                    if group is not None:
+                        chown(path, group=group)
+                        mode |= stat.S_IRGRP
+                os.chmod(path, mode)
+                file_paths[embedded_file.name] = path
+
+        command = str(step_script.actions.onRun.command)
```
wyongzhi, line 369: **Non-blocking** (current attachment-sync callers pass a literal command, and the STOPGAP note above already tracks moving this into the library via OpenJobDescription/openjd-sessions-for-python#332), but flagging the divergence while this block is live:

Want to confirm this first, `{{ Task.File.* }}` resolution is applied to `args` (L377-381) but not to `command`, which is passed through unresolved. > Both the v0 path and the Rust engine's native runner resolve `command` and `args` identically ([openjd-rs `runner/mod.rs#L273-L282`](https://github.com/OpenJobDescription/openjd-rs/blob/28797d0b56d568c5c8ca065cef322a9b099a8268/crates/openjd-sessions/src/runner/mod.rs#L273-L282) resolves `command` through the same `resolve_string_with` as `args`). A script using a file reference in the command position, e.g. `command: "{{ Task.File.syncScript }}"`, would pass the raw template string to the OS here while working fine on the Python runtime. If we want to keep the consistency btw Openjd-rs and this layer, shall we add command resolutions too?

Cheap fix — apply the same substitution to `command` after `file_paths` is built:

```python
command = re.sub(
    r"\{\{\s*Task\.File\.(\w+)\s*\}\}",
    lambda m: file_paths.get(m.group(1), m.group(0)),
    str(step_script.actions.onRun.command),
)

File: src/deadline_worker_agent/sessions/runtime/rust.py [hunk 14]
```diff
@@ -0,0 +1,420 @@
...
+def _to_rust_path_mapping_rules(
+    rules: Optional[list[PathMappingRule]],
+) -> Optional[list[RustPathMappingRule]]:
+    """Convert worker (v0) PathMappingRule lists into the _v1 type, passing None through."""
+    if rules is None:
+        return None
+    return [_to_rust_path_mapping_rule(rule) for rule in rules]
+
+
+def _to_v0_action_status(status: RustActionStatus) -> ActionStatus:
```
leongdl, line 182: As a follow up, I think we should reverse the relationship.

The code should default to V1 types and then bridge V1 types to V0 types. Then after our migration, we delete code, and don't need to add and update code.

Think of the testing required afterwards if we did a V0->V1 bridge. It means V0 can't be easily deleted.

leongdl (approved):
Please do the type migration reversal in a follow up.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -0,0 +1,420 @@
...
+    ) -> None:
+        # The shared action layer hands a pydantic v2023_09 StepScript, but the
+        # Rust session needs a native _v1 step. Serialize to the OpenJD wire
+        # shape and rebuild it natively. deserialize_step requires a named step
+        # and the binding has no bare step-script deserializer (tracked in
+        # OpenJobDescription/openjd-sessions-for-python#334), so wrap the script
+        # as ``{"name": ..., "script": ...}`` and unwrap it after decoding.
+        # exclude_none=True is required, not cosmetic: the Rust decoder rejects
+        # explicit nulls (OpenJD treats absent and null as equivalent).
+        native_step_script = deserialize_step(
```
wyongzhi, line 314: **Non-blocking, Suggestion** (applies here and also to the `create_environment(decode_environment_template(...))` call in `enter_environment`, L269-278): consider wrapping the decode round-trip in a try/except that re-raises with boundary context.

This hop is a failure surface the Python runtime doesn't have: the Rust decoder is an independent implementation of the same spec, and the hand-built wire shape (hardcoded spec version, "Placeholder" wrapping, `exclude_none`) is only exercised at decode time. When it fails, the worker's action handler reports FAILED with good *action* context (session/action/queue IDs) — no traceback is logged, and the raw message flows straight to the customer-visible `progressMessage` — so nothing points at the conversion boundary unless the message itself does.

Same pattern this revision already applies at the smaller boundaries (`_ACTION_STATES`, `_SPEC_REVISIONS`, the `getattr` guard), extended to the largest one:

```python
try:
    native_environment = create_environment(
        decode_environment_template({...})
    )
except (ValueError, RuntimeError) as exc:
    raise RuntimeError(
        "Failed to convert the environment for the Rust session runtime; "
        f"this indicates a v0/_v1 model divergence or wire-shape mismatch: {exc}"
    ) from exc
```

src/deadline_worker_agent/sessions/runtime/rust.py
seant-aws: I'll log this and put this in the followup PR where I add the tests I wrote, thnx

src/deadline_worker_agent/sessions/runtime/rust.py
seant-aws: i can address these in the next PR as a followup as well, thnx

godobyte (approved):

PR #1005, opened 2026-07-13
fix: properly capture session action structure errors

epmog:

```
'StepDetailsError' object has no attribute 'step_id'
```

This has been a latent bug in the worker agent for a while now, in that was not filling in error information (step/task id) when failing to parse details from the service. This can happen if there's a mismatch in the version of openjd model that validated the submission vs. the version of openjd model that is used by the worker agent.

We need to raise the error properly and have it report as a job failure, instead of a re-attempting to parse that information.

session actions that fail on parsing should properly be marked as failed.

added tests to the base class and all children. Also ran the worker agent in a container to validate the old and new behaviour.

N/A

Fixing

----

### DIFF
--- a/src/deadline_worker_agent/scheduler/session_queue.py
+++ b/src/deadline_worker_agent/scheduler/session_queue.py
@@ -406,6 +406,16 @@ class SessionActionQueue:
             action_type = action_queue_entry.definition["actionType"]
             action_definition = action_queue_entry.definition
             action_id = action_definition["sessionActionId"]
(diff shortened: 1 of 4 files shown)


jericht (dismissed):

seant-aws (dismissed):
nice catch

godobyte (dismissed):

seant-aws (approved):

jericht (approved):

PR #1010, opened 2026-07-15
fix: Add timeout to IMDS requests to prevent shutdown hang

folouiseAWS:
## What was the problem/requirement?

The IMDSv2 token request in `Worker._get_ec2_metadata_imdsv2_token()` (and the spot instance-action and ASG lifecycle-state requests) had no timeout and only caught `requests.ConnectionError`.

On hosts where the IMDS address (169.254.169.254) is silently dropped rather than actively refused — possible on any non-EC2 host depending on network configuration — the token request in `Worker.run()` blocks the **main thread indefinitely**. The scheduler keeps running in the executor, so the agent appears healthy and processes work, but SIGTERM shutdown can never complete: the worker is never transitioned to `STOPPED` and must be SIGKILLed, leaving the service-side worker resource in `STARTED` until it ages out to `NOT_RESPONDING`.

## What was the solution?

- Add `timeout=0.5` to all three IMDS requests in `worker.py`, matching the timeout and rationale already used by `startup.bootstrap._get_metadata`.
- Treat `requests.Timeout` the same as `requests.ConnectionError`: IMDS is unavailable.
- Extracted as a class constant `Worker._IMDS_REQUEST_TIMEOUT_SECONDS` with documentation.

## What was the testing done?

- Unit tests added for timeout behavior of all three IMDS queries; existing call assertions updated. Full `test/unit/test_worker.py` module passes (36 tests).
- `hatch run lint` (ruff + mypy) passes.
- Reproduced live on an arm64 macOS host on a network where the IMDS address is silently dropped, running the agent against a real Deadline Cloud farm: **without** this change, the agent hung indefinitely after SIGTERM (thread dump showed the main thread blocked in `requests.put` at `worker.py:391`) and the worker stayed `STARTED`; **with** this change, the agent transitioned the worker to `STOPPED` and exited in under one second.


andychoquette:
Independent confirmation of this from a second direction, plus one mechanism detail and one review concern.

**Reproduced in CI.** I hit this while adding a macOS e2e suite on a GitHub-hosted runner ([#1080](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1080)), where `169.254.169.254` has a route via `en0` but nothing answering ARP — the exact blackhole condition in your description. `test_worker_lifecycle_status_is_expected` and `test_worker_enters_stopping_state_while_draining` both fail there, and a SIGUSR1 thread dump from the stalled agent parks in the same frame you found:

```
entrypoint.py:205   worker_sessions.run()
worker.py:225       if self._get_ec2_metadata_imdsv2_token():
worker.py:394       response = requests.put(
urllib3/util/connection.py:85   sock.connect(sa)
```

It is intermittent rather than constant, which took a while to pin down: the kernel's ARP negative cache makes some attempts fail in milliseconds, so the hang only appears when SIGTERM lands inside a connect window.

**The causal chain to `STARTED`, for the record.** The probe at `worker.py:225` is on the *startup* path, not the shutdown path — it runs after `self._executor.submit(self._scheduler.run)` but before `wait(futures)`. So the scheduler thread registers the worker, polls and drains normally while the main thread is still parked, which is why the agent looks healthy right up until it cannot exit. `_agent_shutdown()` — the thing that reports `STOPPED` — is at `entrypoint.py:213`, *after* `worker_sessions.run()` returns. A parked main thread therefore means `run()` never returns and the status update never runs, exactly as you describe.

**One review concern on the fix as written.** The probe result is latched once at startup and gates the EC2 monitor for the agent's whole lifetime. With a bare `0.5s` timeout and no retry, a genuine EC2 host that is momentarily slow to answer silently loses spot-interruption and ASG-lifecycle monitoring until it restarts — trading a hang for a quiet loss of function. Worth considering a couple of retries on the `worker.py:225` probe specifically (the two in-loop calls are re-issued each poll, so they self-heal and need nothing). I had a branch doing this at `1.0s` x 3 attempts, but your change is the earlier and smaller one, so I would rather see this land than open a competing PR — happy for it to be a follow-up, or I can send you the loop if you want it here.

Also worth noting `installer/__init__.py:41` and `:50` have the same unbounded pattern, if you want them in scope.

Two practical notes: the branch is `mergeable_state: behind`, so it needs a mainline merge before checks will run, and mainline is currently red on an unrelated `openjd-model` resolution drift that [#1082](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1082) fixes. Once this merges, the macOS e2e suite in #1080 covers this path on every run.

andychoquette:
closing in favor of https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1083, which addresses this issue more comprehensively.

PR #1013, opened 2026-07-17
fix: retry transient network errors in UpdateWorkerSchedule

ZainAallii:
## fix: retry transient network errors in UpdateWorkerSchedule

Add retry handling for transient network errors in UpdateWorkerSchedule. These errors were previously caught by the generic `except Exception` handler and raised as `DeadlineRequestUnrecoverableError`, causing the agent to terminate on brief network blips. Now they are retried with exponential backoff, matching the existing retry behavior for throttling and internal server errors.

Any transient network error (connection reset, timeout, endpoint unreachable) during an `UpdateWorkerSchedule` heartbeat call caused the worker agent to immediately terminate. The generic `except Exception` handler wrapped these as `DeadlineRequestUnrecoverableError` without any retry attempt. This affected customers with momentary network instability (VPN reconnection, load balancer rotation, NIC flap) — their agents would die and lose in-progress sessions. It was reported from the field and corroborated by simultaneous CloudWatch Logs upload failures observed during brief network interruptions.

Added an `except _TRANSIENT_NETWORK_EXCEPTIONS` block before the generic `except Exception` catch-all in `update_worker_schedule`. Rather than enumerating individual leaf exceptions, it catches botocore's two transport base classes, `ConnectionError` and `HTTPClientError`. This covers `ConnectionClosedError`, `ConnectTimeoutError`, `EndpointConnectionError`, and `ReadTimeoutError`, and also `SSLError`, `ProxyConnectionError`, and `ResponseStreamingError`, mirroring how botocore itself classifies retryable connection errors. The shared handling was extracted into a `_handle_transient_network_error` helper that logs, performs an interruptible backoff wait, and raises `DeadlineRequestUnrecoverableError` once the retry limit is reached. Retries use the existing exponential backoff (capped at 30s); the limit is 5 retries, which is roughly 15s of worst-case tolerance before giving up.

Workers now survive brief network blips during `UpdateWorkerSchedule` calls instead of terminating. There is no behavior change for actual unrecoverable errors (4xx responses, validation errors, etc.) — those still terminate as before. On a sustained outage the agent still gives up after the retry limit and exits.

Note: the full `hatch run test` suite was not run in this environment (the local hatch env could not be provisioned), so full-suite pass counts and coverage should be taken from CI. The following were performed and verified directly:

- Unit tests: ran `test/unit/aws/deadline/test_update_worker_schedule.py` against the branch source — 25 passed. This includes parametrized cases for all seven covered transient exception types (verifying a retry occurs and the call succeeds on the second attempt), a test that the call raises `DeadlineRequestUnrecoverableError` after the max retry count, and a test that the backoff wait is interruptible and a set interrupt breaks the loop with `DeadlineRequestInterrupted`.
- Lint: `ruff check` passed on the changed source and test files. `ruff format` and `mypy` were not run locally and should be confirmed by CI.
- Live worker test: installed the branch as an editable install and ran a real worker against a customer-managed fleet. Confirmed the running code was the branch source (module resolved to the working tree, `_handle_transient_network_error` present, transient tuple = `ConnectionError, HTTPClientError`). With the worker heartbeating `UpdateWorkerSchedule` successfully, a real network fault was injected by adding a host firewall outbound rule scoped only to the endpoint IPs on port 443 (remote access and all other traffic unaffected). The agent logged `Transient network error during UpdateWorkerSchedule (EndpointConnectionError: ...). Retrying in N seconds...` across five retries with increasing exponential backoff, then logged `Transient network error persisted after 5 retries (...). Giving up.` and exited cleanly with `DeadlineRequestUnrecoverableError`. The firewall rule was removed afterward and connectivity confirmed restored.

No external documentation changes required. The fix is internal retry behavior, with a warning log per retry and an error log on give-up for observability.

No. This only adds retry behavior where the agent would previously crash. No API, config, or interface changes.

### DIFF
--- a/src/deadline_worker_agent/aws/deadline/__init__.py
+++ b/src/deadline_worker_agent/aws/deadline/__init__.py
@@ -9,7 +9,12 @@ from functools import wraps
 import random
 
 from botocore.retries.standard import RetryContext
-from botocore.exceptions import ClientError
+from botocore.exceptions import (
+    ClientError,
+    # Aliased to avoid shadowing the Python builtin ConnectionError.
+    ConnectionError as BotocoreConnectionError,
+    HTTPClientError,
+)
 
 from deadline.job_attachments import version as deadline_job_attachments_version
 from deadline.job_attachments.progress_tracker import SummaryStatistics
@@ -41,6 +46,21 @@ from ...log_sync.log_constants import (
 )
 
 _logger = logging.getLogger(__name__)
+
+# Transport-level errors that indicate transient connectivity issues and should be retried.
+# We catch botocore's two transport base classes rather than enumerating leaf types, which
+# mirrors how botocore itself classifies retryable connection errors and covers cases like
+# ConnectionClosedError, ReadTimeoutError, ResponseStreamingError (HTTPClientError) and
+# EndpointConnectionError, ConnectTimeoutError, ProxyConnectionError, SSLError (ConnectionError).
+# These are distinct from ClientError (service responses), so they don't overlap with the
+# error-code handling above.
+_TRANSIENT_NETWORK_EXCEPTIONS = (BotocoreConnectionError, HTTPClientError)
+
+# Maximum retries for transient network errors before treating as unrecoverable.
+# Uses the same exponential backoff as throttle retries (base 2, per-retry delay
+# randomized up to ~0.5s, 1s, 2s, 4s, 8s and capped at 30s), so 5 retries gives
+# roughly 15s of worst-case tolerance before giving up.
+_MAX_TRANSIENT_NETWORK_RETRIES = 5
 
 # Generic function return type.
 F = TypeVar("F", bound=Callable[..., Any])
@@ -165,6 +185,44 @@ def _get_resource_id_and_status_from_conflictexception_header(
     context = response.get("context", {})
     resourceId = response.get("resourceId")
     return resourceId, context.get("status")
+
+
+def _handle_transient_network_error(
+    e: Exception,
+    transient_retries: int,
+    backoff: Backoff,
+    interrupt_event: Optional[Event],
+    api_name: str,
+) -> None:
+    """Handles a transient network error raised during a Deadline Cloud API call by logging
+    and performing an interruptible backoff wait before the caller retries.
+
+    Transient network errors (connection reset, timeout, etc.) commonly occur due to VPN
+    reconnection, load balancer rotation, or brief network blips, and should be retried
+    rather than treated as fatal.
+
+    Note: The caller owns the ``transient_retries`` counter and is responsible for
+    incrementing it after this function returns.
+
+    Raises:
+        DeadlineRequestUnrecoverableError -- When ``transient_retries`` has reached
+            ``_MAX_TRANSIENT_NETWORK_RETRIES``, i.e. the error has persisted past the retry limit.
+    """
+    if transient_retries >= _MAX_TRANSIENT_NETWORK_RETRIES:
+        _logger.error(
+            f"Transient network error persisted after {_MAX_TRANSIENT_NETWORK_RETRIES} "
+            f"retries ({type(e).__name__}: {e}). Giving up."
+        )
+        raise DeadlineRequestUnrecoverableError(e)
+    delay = backoff.delay_amount(RetryContext(transient_retries))
+    _logger.warning(
+        f"Transient network error during {api_name} ({type(e).__name__}: {e}). "
+        f"Retrying in {delay} seconds..."
+    )
+    if interrupt_event:
+        interrupt_event.wait(delay)
+    else:
+        sleep(delay)
 
 
 def assume_fleet_role_for_worker(
@@ -747,6 +805,7 @@ def update_worker_schedule(
     # Retry API call when being throttled
     backoff = Backoff(max_backoff=30)
     retry = 0
+    transient_retries = 0
     while True:
         if interrupt_event is not None and interrupt_event.is_set():
             raise DeadlineRequestInterrupted("UpdateWorkerSchedule interrupted")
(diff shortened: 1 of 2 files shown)


leongdl:
**Call stack and why the change was needed**

```
WorkerScheduler.run() (scheduler/scheduler.py — the heartbeat loop)
  → update_worker_schedule() (aws/deadline/__init__.py — hand-rolled retry loop)
    → boto3 deadline client.update_worker_schedule()
      → botocore HTTP layer (raises ConnectionClosedError / ConnectTimeoutError /
        EndpointConnectionError / ReadTimeoutError on network blips)
```

Before this PR, `update_worker_schedule`'s loop only caught `ClientError` (handling ThrottlingException, InternalServerErrorException, ConflictException, etc. by error code). Network-level exceptions are not `ClientError` — they're raised before any HTTP response exists — so they fell through to the generic `except Exception`, got wrapped as `DeadlineRequestUnrecoverableError`, and the agent terminated, killing in-progress sessions on a momentary blip. The PR inserts an `except _TRANSIENT_NETWORK_EXCEPTIONS` block ahead of the catch-all, retrying up to 5 times with the existing `Backoff(max_backoff=30)`.

**Why the boto client retry strategy didn't absorb this**

The retry strategy in the boto client is deliberately disabled. From `src/deadline_worker_agent/boto/config.py`:

```python
DEADLINE_BOTOCORE_CONFIG = Config(
    retries={"max_attempts": 1},   # one attempt, zero retries
    user_agent_extra=construct_user_agent(),
)
```

with a docstring explicitly saying: *"the worker agent has its own retry logic for AWS Deadline Cloud API requests. See the `aws/deadline` sub-package."*

So the answer has two parts:

1. Botocore's standard retry mode would normally absorb all four of these exceptions — connection errors and timeouts are classified as transient-retryable by botocore. But the agent opts out with `max_attempts: 1`.
2. It opts out for good reasons: the agent is long-running and needs (a) interruptible waits (`interrupt_event.wait(delay)` so shutdown/drain isn't blocked mid-sleep — botocore's internal retry sleeps are uninterruptible), (b) unbounded throttle retries with `NoOverflowExponentialBackoff` (in `boto/retries.py`, built so a fleet-wide service issue doesn't require restarting every worker host), and (c) per-error-code semantics like honoring `retryAfterSeconds` from the service response and treating `ConflictException/STATUS_CONFLICT` as "worker offline."

The bug was simply that when we took ownership of retries away from botocore, the hand-rolled loop covered service errors but forgot the transport-level exceptions botocore would have handled. This PR closes that gap. Worth noting: the same gap almost certainly exists in the sibling functions (`create_worker`, `update_worker`, `batch_get_job_entity`, `assume_*_role_for_worker`) which use the same pattern with the same catch-all — this PR only fixes `update_worker_schedule`.

**Suggestion — extract a reusable helper**

Since the sibling functions need the same fix, rather than duplicating this `except` block into each, could we extract the shared pieces into a helper? Something like:

```python
def _handle_transient_network_error(
    e: Exception,
    transient_retries: int,
    backoff: Backoff,
    interrupt_event: Event | None,
    api_name: str,
) -> None:
    """Logs and waits before retrying a transient network error.
    Raises DeadlineRequestUnrecoverableError when retries are exhausted."""
    if transient_retries >= _MAX_TRANSIENT_NETWORK_RETRIES:
        _logger.error(
            f"Transient network error persisted after {_MAX_TRANSIENT_NETWORK_RETRIES} "
            f"retries ({type(e).__name__}: {e}). Giving up."
        )
        raise DeadlineRequestUnrecoverableError(e)
    delay = backoff.delay_amount(RetryContext(transient_retries))
    _logger.warning(
        f"Transient network error during {api_name} ({type(e).__name__}: {e}). "
        f"Retrying in {delay} seconds..."
    )
    if interrupt_event:
        interrupt_event.wait(delay)
    else:
        sleep(delay)
```

Then the `except` block in each API wrapper reduces to:

```python
except _TRANSIENT_NETWORK_EXCEPTIONS as e:
    _handle_transient_network_error(
        e, transient_retries, backoff, interrupt_event, "UpdateWorkerSchedule"
    )
    transient_retries += 1
```

(We considered a decorator, but the retry state — throttle counter, backoff, transient counter — lives inside each function's loop, and a function-level decorator would re-invoke from scratch and reset the throttle backoff state. A helper keeps the state machines intact.)

Happy to have the sibling functions handled in a follow-up PR if you'd prefer to keep this one scoped to `UpdateWorkerSchedule`.

seant-aws:
wanted to double check we only had one layer of retry to prevent too many nested ones, looks like we do:

```
Retry Stack (scheduler path)
Copy
scheduler.run()                          ← polling loop, NOT a retry loop
  └→ _sync()
       └→ update_worker_schedule()       ← THE ONLY retry layer
            └→ deadline_client.update_worker_schedule()
                 └→ botocore (max_attempts=1 — retries DISABLED)
```

src/deadline_worker_agent/aws/deadline/__init__.py
```diff
@@ -42,6 +48,18 @@
 
 _logger = logging.getLogger(__name__)
 
+# Network errors that indicate transient connectivity issues and should be retried
```
seant-aws: How did we land on these specific errors, just curious. I did a bit of looking, there might be some missing we can consider adding ?


Exception | Base class | Scenario
-- | -- | --
ResponseStreamingError | HTTPClientError | Connection dropped mid-response-stream
ProxyConnectionError | ConnectionError | Proxy restart/rotation (farms behind corporate proxies)
SSLError | ConnectionError | Transient TLS handshake failures

<p style="box-sizing: border-box; margin: 12px 0px 4px; padding: 0px; color: rgb(237, 238, 240); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(17, 17, 19); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Alternatively, catch the two base classes directly, i think that might cover your exisitng exceptions, and follows botocore's defnition a bit more closely ?

leongdl (changes requested):
Please refactor as suggested.

src/deadline_worker_agent/aws/deadline/__init__.py
ZainAallii: Added the missing ones.

seant-aws (approved):
thnx for the changes ! good from my end

leongdl (approved):

PR #1014, opened 2026-07-17
fix(credentials): Retry with bootstrap creds on hibernate/sleep expiry

ZainAallii:
## Problem

Fleet-role credentials are held in a non-auto-refreshing session (a `PassThroughCredentialProvider` wrapping `SettableCredentials`), so they're only rotated by our explicit `refresh_credentials()` call.

`refresh_credentials()` picks which session to call `AssumeFleetRoleForWorker` with based on `are_expired()`: if the current fleet-role creds look valid, it uses them; otherwise it falls back to the bootstrap session.

There's a race: if the machine hibernates/sleeps between the `are_expired()` check and the API call completing, the fleet-role creds expire mid-flight. The resulting 403 was mapped to `DeadlineRequestUnrecoverableError` and caused the worker to exit.

## Fix

After catching `DeadlineRequestUnrecoverableError`, we re-check `are_expired()`. If the creds expired during the call and we weren't already using bootstrap, we retry once with the bootstrap session. Non-hibernate failures are unaffected (re-raised as before), and we never retry bootstrap→bootstrap, so there's no loop.

## Notes for reviewers

- **Why bootstrap is the fallback:** you can't refresh expired fleet-role creds using those same expired creds. The bootstrap session is a plain boto3 `Session` whose credentials come from the standard chain (instance role/IMDS, profile, env).

- **"What if bootstrap is also expired?"** For refreshable sources (IMDS, etc.) botocore auto-refreshes the bootstrap creds on demand, so the retry self-heals after a long hibernate. The only unrecoverable case is revoked static creds, which correctly propagates. No extra code needed — verified and covered by a test.

- **`are_expired()` is a heuristic:** we can't tell from the error alone whether expiry caused it, so we infer it from the clock. A genuine AccessDenied that coincides with expiry will also be retried. The blast radius is small (bootstrap is the more-privileged path, timing overlap is narrow) and there's an inline comment calling this out.

## Testing

Unit tests cover all branches:
- expiry mid-call → retry with bootstrap succeeds
- non-expiry failure → re-raised, no retry
- already on bootstrap → re-raised, no retry (no loop)
- expiry mid-call → bootstrap retry also fails → error propagates, no creds set

### DIFF
--- a/src/deadline_worker_agent/aws_credentials/worker_boto3_session.py
+++ b/src/deadline_worker_agent/aws_credentials/worker_boto3_session.py
@@ -89,17 +89,37 @@ class WorkerBoto3Session(BaseBoto3Session):
             )
         )
 
-        deadline_client = session.client("deadline", config=DEADLINE_BOTOCORE_CONFIG)
-        try:
-            response = assume_fleet_role_for_worker(
+        def _assume(boto_session: Session) -> Any:
+            deadline_client = boto_session.client("deadline", config=DEADLINE_BOTOCORE_CONFIG)
+            return assume_fleet_role_for_worker(
                 deadline_client=deadline_client,
                 farm_id=self._farm_id,
                 fleet_id=self._fleet_id,
                 worker_id=self._worker_id,
             )
+
+        try:
+            response = _assume(session)
         except DeadlineRequestUnrecoverableError as e:
-            # Re-raise to let the caller know, and handle the exception.
-            raise e from None
+            # If credentials expired during the call (e.g. machine hibernated/slept between
+            # the are_expired() check and the API call), retry with bootstrap credentials.
+            #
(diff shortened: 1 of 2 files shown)


leongdl (approved):

src/deadline_worker_agent/aws_credentials/worker_boto3_session.py
```diff
@@ -89,17 +89,37 @@ def refresh_credentials(self) -> None:
             )
         )
 
-        deadline_client = session.client("deadline", config=DEADLINE_BOTOCORE_CONFIG)
-        try:
-            response = assume_fleet_role_for_worker(
+        def _assume(boto_session: Session) -> Any:
```
godobyte, line 92: nit - could return `AssumeFleetRoleForWorkerResponse` instead of `Any`.

godobyte (approved):

PR #1029, opened 2026-07-27
feat: emit HostConfigDuration metric to customer CloudWatch

RenaL27:

Customers have no visibility into how long their host configuration scripts take to execute on worker instances. This makes it difficult to identify slow startup scripts that delay workers from accepting jobs.

Added a `HostConfigDuration` CloudWatch metric that the worker agent emits to the customer's `AWS/DeadlineCloud` namespace after the host configuration script completes. The metric is published using the fleet role credentials via `PutMetricData` with dimensions FarmId, FleetId, and Region. Includes retry logic (1 retry with 0.5s backoff) and graceful failure handling: metric emission failures are logged as warnings but do not block the worker from starting.

The feature is gated behind the `DEADLINE_WORKER_EMIT_HOST_CONFIG_METRIC` environment variable. The metric is only emitted when this variable is set, allowing controlled rollout.

- No impact in production. gated by env var (not currently enabled)
- No API changes required
- No impact on worker agent startup if CloudWatch call fails (fire-and-forget with warning log)
- Adds `backoff` as a new dependency

- Unit tests: 4 new tests covering successful publish, CloudWatch failure handling, retry behavior, and region resolution
- Live testing: Ran agent in Docker against a dev CMF fleet with a host configuration script (`sleep 60`) attached. Verified `PutMetricData` returns 200 and metric appears in CloudWatch under `AWS/DeadlineCloud` namespace with a value of ~60 seconds.

Screenshot 2026-07-27 at 4 21 35 PM

No external documentation changes required. The metric will appear automatically in the customer's CloudWatch console once enabled.

No.

----

### DIFF
--- a/src/deadline_worker_agent/startup/entrypoint.py
+++ b/src/deadline_worker_agent/startup/entrypoint.py
@@ -8,7 +8,9 @@ import logging
 import os
 import subprocess
 import sys
-from time import sleep
+from time import monotonic, sleep
+
+import backoff
 from botocore.exceptions import NoRegionError
 from logging.handlers import TimedRotatingFileHandler
 from threading import Event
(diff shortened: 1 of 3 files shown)


src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -443,6 +445,71 @@ def _configure_base_logging(
...
+    if not os.environ.get("DEADLINE_WORKER_EMIT_HOST_CONFIG_METRIC"):
+        return
+
+    region = session.region_name
+
+    try:
+        cw_client: Any = session.client("cloudwatch", config=OTHER_BOTOCORE_CONFIG)
+        _put_metric_data_with_retry(
+            cw_client,
+            namespace="AWS/DeadlineCloud",
```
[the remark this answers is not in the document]
RenaL27: This uses the same namespace (AWS/DeadlineCloud) as the existing WorkerStartDuration and WorkerCreateDuration metrics. We verified in dev testing that PutMetricData succeeds (200 response) using the fleet role credentials.

src/deadline_worker_agent/startup/entrypoint.py
```diff
@@ -443,6 +445,71 @@ def _configure_base_logging(
...
+            "Failed to publish HostConfigDuration to customer",
+            extra={
+                "farm_id": config.farm_id,
+                "fleet_id": config.fleet_id,
+                "error": str(e),
+            },
+        )
+
+
+@backoff.on_exception(backoff.constant, Exception, max_tries=2, interval=0.5)
```
jericht: Do we need to add a dependency on `backoff` here? Boto clients already have configurable retries built in, can we just configure that in the way we want?

For reference, I also noticed we have our own Backoff class as well we might want to reuse: https://github.com/aws-deadline/deadline-cloud-worker-agent/blob/2457b71463a1bd39aa0e57e4e11244486b80a90d/src/deadline_worker_agent/boto/retries.py#L11

RenaL27: Good call! I removed the backoff dependency entirely. Now just calling `put_metric_data` directly and relying on the built-in retries.

PR #1036, opened 2026-07-31
fix: sample host CPU utilization on a long-lived metrics thread

waninggibbon:

Host CPU utilization was frequently logged as `0.0` even while a worker was active.
`HostMetricsLogger` scheduled each collection with a new `threading.Timer`, but
non-blocking `psutil.cpu_percent()` compares against the previous call made from
the *same thread*. Every timer tick ran on a fresh thread, so every sample was
that thread's first call, and every first call returns psutil's uninitialized
zero.

- Replace the timer chain with one long-lived metrics thread and an interruptible
  stop event.
- Prime `psutil.cpu_percent()` on that thread, wait one collection interval, and
  collect every subsequent sample on the same thread. Prime the network and disk
  counters there too, so the first logged rates cover a real interval.
- Signal and join the metrics thread during context-manager shutdown, bounded by
  `JOIN_TIMEOUT_S` (1s) so a stuck collection cannot hold up worker shutdown.
- Degrade to logging no host metrics if the metrics thread cannot be started,
  rather than propagating `RuntimeError` out of the context manager and taking
  down the worker over an optional observability feature.
- Bound the `nvidia-smi` query with `GPU_QUERY_TIMEOUT_S` (5s), so an unresponsive
  driver cannot keep the metrics thread from observing the stop event.
- Give each metrics thread its own stop `Event`, passed as an argument rather than
  read back off `self`, so a thread abandoned by a timed-out join cannot be revived
  by its successor resetting a shared event.
- Stop the running thread if the context manager is re-entered, instead of
  overwriting the only references to it and orphaning a daemon thread that nothing
  can signal or join.

CPU utilization now represents activity over a complete collection interval
instead of repeatedly reporting psutil's uninitialized value.

Other behaviour changes worth confirming, all deliberate:

1. **The first host metrics event is now emitted after one complete collection
   interval** (default 60s) instead of immediately at startup. An immediate sample
   is precisely the invalid `0.0` reading this PR removes, and shortening only the
   first wait would make the first line's rate metrics wrong, because they divide
   by the constant `self.interval_s` rather than measured elapsed time. The
   consequence is that a worker living less than one interval logs no host metrics
   — previously it logged one line, but that line's CPU value was always garbage
   and its four rate values were always `0`.
2. **`cpu-usage-percent` changes meaning.** Anything built on the previously
   constant `0.0` sees real values after upgrade.
3. **Metrics reliably stop after context exit.** Previously, if a `Timer` had
   already fired, `__exit__`'s `cancel()` was a no-op and the in-flight
   `log_metrics` scheduled a fresh timer in its `finally`, so host metrics could
   keep logging for the life of the process — and each non-daemon `Timer` could
   delay interpreter exit.
4. **The metrics thread is a daemon**, so a collection in flight at shutdown can be
   lost, where before a wedged `nvidia-smi` could hold the process open.
5. **Up to 1s of new serial shutdown latency** from the bounded join, paid only
   when a collection is in flight.
6. **A slow `nvidia-smi` now omits `gpu-*` keys** from that line rather than
   blocking, logged at debug. Making that observable, and making a transient
   `nvidia-smi` failure stop disabling GPU metrics for the life of the process, is
   left to the follow-up described below.
7. Four new `WARNING`-level messages on the `deadline_worker_agent.metrics` logger.

- `hatch run test`: 3,136 passed, 47 skipped; 85% coverage. Run repeatedly to check
  the real-thread test for flakiness (60 consecutive isolated runs, no failures).
- `hatch run lint`: Ruff lint and formatting passed; Mypy passed for 199 source
  files.
- Mutation-checked the tests that pin this fix. Deleting `_prime_metrics()` from
  `_run` fails `test_run_primes_baselines_before_logging` and nothing else;
  shortening the first `stop_event.wait` fails it too; deleting the re-entry
  teardown fails `test_enter_twice_stops_the_first_thread`; deleting
  `self._stop_event.set()` from `__exit__` fails five tests.

No documentation change is required. Existing metric names and the log schema are
unchanged.

No. Metric names, units, and the log schema are all unchanged. The only behavioural
changes are the timing and content described above.

### Scope

Two things that previously rode along in this PR have been split out, so that each
lands with an honest changelog entry and can be reverted independently:

- The `total-disk-used-percent` scale correction, which rescaled a published metric
  by 100x and needed its own `BREAKING CHANGE` note. Already shipped separately as
  #1067.
- The `nvidia-smi` retry policy. Mainline disables GPU metrics for the lifetime of
  the process on a single transient `nvidia-smi` failure; fixing that grew to about
  60% of this diff across three review rounds, none of it needed to fix CPU
  sampling, so it will follow as its own change. What remains here is the thread
  rewrite plus the three things it directly necessitates: the thread is a daemon,
  the shutdown join is bounded, and the `nvidia-smi` query has a timeout — before
  this PR an unbounded `nvidia-smi` inside a non-daemon `Timer` thread could block
  interpreter exit.

----


src/deadline_worker_agent/metrics.py
```diff
@@ -20,27 +20,39 @@ class HostMetricsLogger:
...
-            self._timer = None
+        self._stop_event.set()
+        if self._thread:
+            self._thread.join()
+            self._thread = None
+
+    def _run(self) -> None:
+        # psutil tracks non-blocking CPU samples per thread. Prime the baseline on this
+        # long-lived thread so every logged value covers one complete metrics interval.
+        psutil.cpu_percent()
```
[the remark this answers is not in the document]
waninggibbon: Addressed in [58fcf78](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1036/commits/58fcf78802993e928b17ad94e074eca1f744ad08)

src/deadline_worker_agent/metrics.py
```diff
@@ -18,29 +18,76 @@
...
+        self._thread.start()
         return self
 
     def __exit__(self, type, value, traceback) -> None:
-        if self._timer:
-            self._timer.cancel()
-            self._timer = None
+        self._stop_event.set()
+        if self._thread:
+            # Bounded join so that a wedged metrics collection (e.g. an unresponsive
```
leongdl: wedged seems to be a common Opus 5 favorite word :)

src/deadline_worker_agent/metrics.py
```diff
@@ -18,29 +18,76 @@
...
+            # nvidia-smi) cannot block Worker shutdown. The thread is a daemon, so it
+            # will not keep the process alive if it outlives this join.
+            self._thread.join(timeout=self.JOIN_TIMEOUT_S)
```
leongdl, line 70: I would be careful here, 5s is a long time to join, since this is a context enter / exit, does it cost the call stack higher using this context any issues?

leongdl:
Thanks for this — the CPU-priming diagnosis is a good catch, and the shift to one long-lived thread is the right shape for it. I dug into the `__enter__`/`__exit__` change specifically to see whether the bounded join can propagate anything upstream. Summary: **the join itself won't crash anything, and in normal operation it costs nothing** — but 5s is the wrong number, and the `__enter__` half introduces a new failure path worth closing before merge.

### The join is safe, and usually free

`Thread.join(timeout=...)` never raises on timeout; the code correctly checks `is_alive()` afterwards and logs instead of assuming success. `__exit__` returns `None`, so it won't accidentally suppress the in-flight exception on the `raise` paths in `Worker.run()`. And because `_run` blocks on `self._stop_event.wait(self.interval_s)` and `__exit__` calls `set()` first, the wait is woken immediately and the join returns in microseconds. Good.

The full 5s is only ever paid when the thread is mid-`log_metrics()` and stuck, and there's exactly one thing in there that can block unboundedly: the `nvidia-smi` `subprocess.check_output` has no `timeout=`. That's the scenario the new comment cites — and it's also fixable at the root.

### Where 5s does cost something

**Windows service stop is the tight one.** `SvcStop()` in `win_service.py` reports `SERVICE_STOP_PENDING` once, with no explicit wait hint and no checkpoint loop. pywin32's default wait hint is on the order of 5s, so a 5s metrics join can consume that whole budget before the rest of the unwind (executor join, then `_agent_shutdown`'s `update_worker` network call) even starts. To be fair this path was already over budget — a single `STOP_PENDING` with no checkpoints was never going to cover a multi-second shutdown — so the join worsens an existing problem rather than creating one. Caveat: I couldn't verify pywin32's exact default locally, so that number is worth a second pair of eyes.

**Spot / ASG shutdown is additive.** `_monitor_ec2_shutdown` returns a `grace_time` equal to the entire remaining window (2 min for ASG, whatever IMDS reports for Spot), and `self._scheduler.shutdown(grace_time=...)` runs *inside* the `with`. So: scheduler potentially consumes the whole window → up to 5s on the metrics join → `_agent_shutdown` then tries `update_worker` to record STOPPED/STOPPING. The 5s eats into a budget already sized to be fully consumed, which can cost the final status update.

**systemd is fine.** The unit generated by `install.sh` sets no `TimeoutStopSec`, so it inherits `DefaultTimeoutStopSec` (typically 90s). 5s against 90s is comfortable.

One ordering note: in `Worker.run()` the managers are entered as `(self._executor, AwsCredentialsRefresher(...), host_metrics_logger)`, so they exit in reverse and the metrics join runs **first** — ahead of `self._executor.__exit__()`. The 5s is strictly serial added latency on every shutdown path, not something overlapping work that was going to happen anyway.

### The unremarked risk is in `__enter__`, not `__exit__`

Previously `__enter__` called `log_metrics()`, which swallows everything internally — it could not fail. Now it calls `Thread.start()`, which raises `RuntimeError` if the OS can't create a thread (resource exhaustion / hitting a thread limit — plausible on a loaded render host running many session subprocesses). That propagates out of `__enter__`, out of the `with` in `Worker.run()`, up to `entrypoint`, where the generic handler calls `sys.exit(1)`. So an optional best-effort *observability* feature can now take down the whole agent, where before it would at worst have logged nothing.

This is the inverse of the bot's earlier finding: that one caught the unprotected psutil prime *inside* `_run` (nicely fixed in 58fcf78 via `_prime_metrics`), but the unprotected `start()` in `__enter__` is still there and has a much larger blast radius.

### Suggested changes, in priority order

1. **Wrap `__enter__`'s thread creation in try/except** and degrade to "no metrics" rather than killing the agent. This is the only real crash path.
2. **Add `timeout=` to the `nvidia-smi` `check_output`.** This bounds the one unbounded blocker and fixes the root cause the join is compensating for.
3. **Drop `JOIN_TIMEOUT_S` to ~1s, or skip the join entirely.** The thread is a daemon and `_stop_event.set()` already wakes it immediately, so there's no correctness reason to wait 5s; a bounded join is really just tidiness, and tidiness shouldn't cost 5s on the Spot path.

I've left these as inline comments at the three spots too. Nothing here is a blocker on the core fix — the metric corrections themselves look right.

src/deadline_worker_agent/metrics.py
```diff
@@ -18,29 +18,76 @@
...
         self._prev_network = None
         self._prev_disk_counters = None
         self.logger = logger
         self.interval_s = interval_s
 
     def __enter__(self) -> HostMetricsLogger:
-        self.log_metrics()
+        self._stop_event.clear()
+        self._thread = Thread(target=self._run, name="HostMetricsLogger", daemon=True)
+        self._thread.start()
```
leongdl: **Suggestion 1 of 3 (highest priority): this is a new crash path for the whole agent.**

Before this PR, `__enter__` called `log_metrics()`, which try/excepts everything internally — it could not fail. Now it calls `Thread.start()`, which raises `RuntimeError` when the OS can't create a thread (resource exhaustion, hitting a thread/process limit — plausible on a loaded render host running many session subprocesses).

That `RuntimeError` propagates out of `__enter__` → out of the `with (...)` in `Worker.run()` → up to `entrypoint`, where the generic `except Exception` handler calls `sys.exit(1)`. So a failure in an optional, best-effort *observability* feature now takes down the Worker Agent, where previously it would at worst have logged nothing.

This is the mirror image of the earlier bot finding: that one caught the unprotected psutil prime *inside* `_run` (fixed nicely in 58fcf78 via `_prime_metrics`), but the unprotected `start()` here is still open and has a much bigger blast radius.

Suggest degrading to "no metrics" instead of failing:

```python
def __enter__(self) -> HostMetricsLogger:
    self._stop_event.clear()
    thread = Thread(target=self._run, name="HostMetricsLogger", daemon=True)
    try:
        thread.start()
    except RuntimeError as e:
        module_logger.warning(
            f"Failed to start the host metrics thread. Host metrics will not be logged. Error: {e}"
        )
    else:
        self._thread = thread
    return self
```

Assigning `self._thread` only on success also keeps `__exit__` correct for free — it already guards on `if self._thread`, so it becomes a no-op rather than joining a thread that never started.

leongdl (dismissed):
Please add the 3 changes, especially the nvidia-smi timeout one.

src/deadline_worker_agent/metrics.py
leongdl, line 1: **Suggestion 2 of 3: this is the unbounded blocker that the new 5s join is compensating for.**

This `check_output` has no `timeout=`, so it can block indefinitely. A wedged `nvidia-smi` is a real failure mode on unhealthy GPU drivers, where the process can sit in uninterruptible sleep — and it's exactly the "unresponsive nvidia-smi" case cited in the new `__exit__` comment.

This line is outside the diff, so it's pre-existing rather than something this PR introduced. But it's worth fixing here, because it's the *root cause*: it's the only thing inside `log_metrics()` that can keep the metrics thread from noticing `_stop_event` promptly, and therefore the only reason the bounded join can ever run out its clock. Bounding the collection makes the join's timeout nearly unreachable in practice.

```python
output = subprocess.check_output(
    ["nvidia-smi", f"--query-gpu={query_str}", "--format=csv,noheader,nounits"],
    stderr=subprocess.PIPE,
    universal_newlines=True,
    timeout=...,  # a few seconds, comfortably under interval_s
)
```

One thing to check if you take this: `subprocess.TimeoutExpired` is not a subclass of `CalledProcessError`, so it would fall through to the existing `except Exception` catch-all below. That does the right thing (debug log, then `_host_has_no_gpu = True`), but note the `_host_has_no_gpu` latch means a single transient timeout permanently disables GPU metrics for the process lifetime. A dedicated `except subprocess.TimeoutExpired` that logs and returns *without* setting the latch is probably what you want, so a one-off hang doesn't cost you GPU metrics forever.

src/deadline_worker_agent/metrics.py
```diff
@@ -18,29 +18,76 @@
 class HostMetricsLogger:
     """Context manager that regularly logs host metrics"""
 
+    # How long to wait for the metrics thread to exit during shutdown before
+    # abandoning it.
+    JOIN_TIMEOUT_S = 5.0
```
leongdl: **Suggestion 3 of 3: 5s is a long time to spend on a metrics thread during shutdown.**

To be clear about what's *not* wrong here: the join is correctly written. `join(timeout=...)` never raises, the `is_alive()` check afterwards logs rather than assuming success, and `__exit__` returning `None` means it won't suppress the in-flight exception on the `raise` paths in `Worker.run()`. And because `_run` blocks on `_stop_event.wait(self.interval_s)` and `__exit__` sets the event first, the normal case returns in microseconds. This value is only ever paid when the thread is stuck mid-collection.

The concern is what it costs when that does happen, given where `__exit__` sits in the shutdown sequence:

- **Windows service stop.** `SvcStop()` reports `SERVICE_STOP_PENDING` once, with no explicit wait hint and no checkpoint loop. pywin32's default hint is on the order of 5s, so this join alone can consume the whole SCM budget before the executor join and `_agent_shutdown`'s `update_worker` call even start. That path was already over budget pre-PR, so this worsens rather than creates the problem — and I couldn't verify pywin32's exact default, so worth a second look.
- **Spot / ASG.** `_monitor_ec2_shutdown` returns a `grace_time` covering the entire remaining window, and `_scheduler.shutdown(grace_time=...)` runs *inside* the `with`. The scheduler can consume that whole window, then this join adds up to 5s, and only then does `_agent_shutdown` try to record STOPPED/STOPPING. That can cost the final status update.
- **systemd is fine** — the generated unit sets no `TimeoutStopSec`, so it inherits the ~90s default.

Worth noting the ordering too: in `Worker.run()` the managers are entered as `(self._executor, AwsCredentialsRefresher(...), host_metrics_logger)`, so they exit in reverse and this join runs **first**, ahead of `self._executor.__exit__()`. It's strictly serial added latency on every shutdown path.

Given the thread is a daemon and `_stop_event.set()` already wakes it immediately, there's no correctness reason to wait 5s — a bounded join here is really just tidiness, and tidiness shouldn't cost 5s on the Spot path. I'd suggest `JOIN_TIMEOUT_S = 1.0` (or dropping the join entirely and relying on the daemon flag). If suggestion 2 lands and `nvidia-smi` gets a `timeout=`, the collection itself becomes bounded and this timeout gets even harder to reach.

Minor, while you're in here: the comment just below uses "wedged" and then "unresponsive" two words later — worth settling on "unresponsive" or "hung" for both.

src/deadline_worker_agent/metrics.py
waninggibbon: Thanks for the suggested fix, applied in [latest commit](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1036/commits/07192af34ac23407200a4fadd04f840d38d05140)

src/deadline_worker_agent/metrics.py
waninggibbon: I went ahead with a 5 second time out, skipped the `_host_has_no_gpu` latch, and added a dedicated timeout expired handler. 

Agree that a transient hang causing us to drop a sample is preferable here. Thanks for the suggestion!

src/deadline_worker_agent/metrics.py
waninggibbon: Yeah good callout, I went ahead and dropped the timeout to just 1 second. The fix you suggested in (2) should make the timeout harder to reach at all. I think this leaves us a place where the existing pywin32 issue isn't addressed but also isn't exacerbated. Let me know what you think here. 

I also dropped "wedged" in favor of unresponsive.

crowecawcaw:
Picking this up from @waninggibbon — thanks for the diagnosis, the per-thread `cpu_percent` cache is a subtle one and the shift to a single long-lived thread is the right fix rather than a workaround.

Merged current mainline (was 19 commits behind, no conflicts, nothing had touched `metrics.py` or `worker.py`). @leongdl's three asks all landed in 07192af, so this is unblocked on your side — re-requesting review.

### On the previously red CI

`gh pr checks` showed every unit job red, but only **macOS 3.12** actually failed; fail-fast cancelled the other 14. The log blob has since expired so the cause isn't recoverable. I ran the full suite 6× locally plus the metrics tests 8× more and couldn't reproduce it, and mainline CI is green, so this reads as a flake. The merge above will give us a clean signal either way.

### Two changes on top

**1. Per-thread stop event.** `__enter__` was calling `self._stop_event.clear()` on a shared event. If a previous `__exit__` timed out its join and abandoned a still-running thread, re-entering would reset the event that thread was watching and revive it, double-logging alongside its successor. Each thread now gets its own `Event`, passed as an argument rather than read back off `self`, so an abandoned thread keeps observing its own already-set event and exits. Not reachable through `Worker` today, which enters once — closing it as a latent footgun.

**2. An end-to-end test for the thread.** Every test in the file patched out `Thread` or `_stop_event`, so nothing would have caught the thread target being mis-wired, the loop never reaching `log_metrics`, or `__exit__` failing to stop the thread. `test_thread_lifecycle_end_to_end` starts the genuine thread at `interval_s=0.01`, waits for a real metrics event, and asserts the thread is stopped and drained on exit. Mutation-checked: deleting `self._stop_event.set()` from `__exit__` fails that test and nothing else.

### One correction to the comments, no behavior change

The rationale for dropping `JOIN_TIMEOUT_S` to 1s was that bounding `nvidia-smi` makes the join timeout "harder to reach." Two things soften that, both now recorded in the comments:

- `GPU_QUERY_TIMEOUT_S` (5s) is **larger** than `JOIN_TIMEOUT_S` (1s), so a merely-slow query still outlasts the join. That's the right trade — `nvidia-smi` can legitimately take over a second on a busy host and a tighter bound would discard good samples — it just means the abandon-warning path stays reachable.
- More importantly, `timeout=` doesn't cover the case it was added for. @leongdl's comment cited `nvidia-smi` sitting in **uninterruptible sleep**; on `TimeoutExpired`, CPython does `process.kill()` and then an *unbounded* `process.wait()`, and a D-state process won't honour SIGKILL until it leaves D-state. So the timeout bounds a *slow* `nvidia-smi`, not a wedged one.

Net effect: the daemon thread plus the bounded join are load-bearing rather than tidiness. Both stay, and the comments now say so, so nobody deletes the join later thinking the subprocess timeout covers it.

### Answering the two open bot threads

**[No metrics line at t=0](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1036#discussion_r3706957104)** — intended, declining the suggested fix. An immediate sample is precisely the invalid `0.0` this PR exists to remove. Emitting a priming sample or shortening only the first wait would also make that line's *rate* metrics wrong, because they divide by the constant `self.interval_s` rather than measured elapsed time. So a worker living under one interval now logs nothing instead of one line of garbage. Called out explicitly in the PR description.

**[`total-disk-used-percent` scale change](https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1036#discussion_r3716615614)** — good catch, and split out to #1067. It's unrelated to thread scheduling and it's the only customer-visible change here, so it gets its own `fix!:` with a `BREAKING CHANGE` footer telling operators to re-tune alarms. Bundling it would have branded this non-breaking sampling fix as breaking in the generated changelog. This PR is now purely the CPU sampling fix and its hardening.

jericht (dismissed):

src/deadline_worker_agent/metrics.py
```diff
@@ -18,29 +18,115 @@
...
         self.interval_s = interval_s
 
     def __enter__(self) -> HostMetricsLogger:
-        self.log_metrics()
+        # A fresh event, handed to the thread explicitly rather than read back off self:
+        # if a previous __exit__ gave up waiting and abandoned a thread that was still
+        # running, that thread keeps observing its own already-set event and exits, rather
+        # than being revived by a reset event and double-logging alongside its successor.
+        stop_event = Event()
+        self._stop_event = stop_event
```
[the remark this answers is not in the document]
crowecawcaw: Not taking this one. The read is correct, but it is unreachable: `Worker.run()` (`worker.py:219`) enters the single `_host_metrics_logger` inside one `with`, and `entrypoint.py` calls `run()` once per process. It is also not a regression — mainline has the same shape in a worse form, since a second `__enter__` there orphans an entire self-perpetuating `Timer` chain, and `threading.Timer` inherits daemon status (non-daemon from the main thread), so the orphan would keep the process alive rather than being reaped.

Happy to add the guard if `HostMetricsLogger` ever gets a second caller, but adding it now would be new code for a state no caller can reach.

src/deadline_worker_agent/metrics.py
```diff
@@ -18,29 +18,115 @@
...
+            if self._thread.is_alive():
+                module_logger.warning(
+                    "Host metrics thread did not exit within "
+                    f"{self.JOIN_TIMEOUT_S} seconds. Abandoning it."
+                )
+            self._thread = None
+
+    def _run(self, stop_event: Event) -> None:
+        self._prime_metrics()
+        while not stop_event.wait(self.interval_s):
```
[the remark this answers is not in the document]
crowecawcaw: Intentional, and covered in the PR description. An immediate first sample *is* the invalid `0.0` CPU reading this change exists to remove, and — as you note — a shortened first wait is only correct if `log_metrics` takes the elapsed window instead of reading `self.interval_s` (the divisors at 244/248/262/265). That refactor is larger than this sampling fix warrants and would land untested against real short-lived workers.

The fair part of the critique is the 8 gauge fields (`memory-*`, `swap-used-bytes`, `total-disk-*`, `user-disk-available-bytes`) that were valid at t=0 and are now withheld because they share a line with CPU. Note the other 4 non-CPU fields are rates, which were emitted as `0` at t=0 anyway, so the loss is 8 of 13 rather than ~13. If that turns out to matter operationally, the follow-up is the `elapsed_s` refactor plus a test asserting an early sample — not a half-correct early line bolted onto this PR.

src/deadline_worker_agent/metrics.py
```diff
@@ -105,6 +192,14 @@ def _get_gpu_metrics(self) -> Dict[str, str]:
...
                 avg_mem_util = round(mem_util_sum / valid_gpu_count, 1)
                 gpu_metrics["gpu-memory-utilization-percent"] = str(avg_mem_util)
+        except subprocess.TimeoutExpired:
+            # Returned without latching _host_has_no_gpu so that a single unresponsive
+            # nvidia-smi does not disable GPU metrics for the lifetime of the process.
+            module_logger.debug(
+                f"nvidia-smi did not respond within {self.GPU_QUERY_TIMEOUT_S} seconds, "
+                "skipping GPU metrics collection"
+            )
+            return {}
```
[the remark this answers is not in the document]
crowecawcaw: Good catch — fixed in 1491243.

`_host_has_no_gpu` now latches only on the two signals that establish it: `FileNotFoundError` (no nvidia-smi) and a clean query reporting no GPUs. `CalledProcessError`, `TimeoutExpired`, unexpected errors, and output we cannot parse each cost one sample instead. The rationale, including the cost of *not* latching (one nvidia-smi call per interval on a host that never recovers), is now in the `_get_gpu_metrics` docstring rather than split across two comments that contradicted each other.

One correction to the finding: "a single malformed line latches too" only held on single-GPU hosts — on multi-GPU, one bad line leaves `valid_gpu_count > 0` and just drops that GPU from the average. Added a test pinning that, alongside tests for the latch/no-latch split and recovery after a `CalledProcessError`.

Developed red/green: the new tests fail on 94dbda8 (6 failures, exactly the changed cases) and pass on 1491243. Full suite 3,142 passed / 47 skipped; ruff and mypy clean.

src/deadline_worker_agent/metrics.py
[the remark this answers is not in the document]
crowecawcaw: Reversing this: the guard is in, as of f4a8eb8.

I declined it above on the grounds that it is unreachable, which is still true — `Worker.run()` enters this once and `entrypoint.py` calls `run()` once. What changed my mind is that two independent reviews of the surrounding code both read the comment at the top of `__enter__` as already covering this case. If the comment misleads readers about which orphaning case is handled, that is worth two lines to make true.

It stops the running thread rather than asserting, reusing the tested `__exit__` path. An assert would fail a Worker over misuse of an optional observability feature, which is inconsistent with the `RuntimeError` degradation a few lines below. Pinned by `test_enter_twice_stops_the_first_thread`; deleting the two lines fails that test and nothing else.

src/deadline_worker_agent/metrics.py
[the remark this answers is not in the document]
crowecawcaw: Update: this work has been split out of this PR rather than landed here.

The latch is a real bug — mainline disables GPU metrics for the lifetime of the process on a single transient `nvidia-smi` failure — but fixing it properly grew to roughly 60% of this diff across three review rounds, none of it needed to fix CPU sampling. It also went through two designs before settling: latching only on definitive signals, then bounding the retry after review showed that `nvidia-smi` reports a permanently GPU-less host the same way it reports a temporarily unavailable one (exit 9 for a driver that cannot be reached, exit 6 for no visible devices), so the "clean query reporting no GPUs" signal was near-unreachable and those hosts would have queried every collection forever.

What it settles on, for the follow-up: geometric backoff between attempts, capped, reset on success — so a host that never reports GPUs stops paying for the query while one that recovers is picked up again, with no permanent give-up. Plus the exit code and stderr recorded rather than discarded, and one above-debug line naming the reason, since that message is the only thing distinguishing a host with no GPU from one whose driver is unhealthy.

What stays in this PR is the `nvidia-smi` timeout, because the thread rewrite requires it: before this change an unbounded `nvidia-smi` inside a non-daemon `Timer` thread could block interpreter exit. The consequence, called out in the PR description, is that a slow `nvidia-smi` now omits `gpu-*` keys from that line with only a debug breadcrumb. Making that observable belongs with the retry change.

crowecawcaw:
Force-pushed a narrowed version of this PR, and renamed it.

**The `nvidia-smi` retry work is out.** It had grown to about 60% of the diff and none of it is needed to fix CPU sampling. It also fixes a distinct bug that deserves its own titled change and its own changelog line — mainline disables GPU metrics for the lifetime of the process on a single transient `nvidia-smi` failure. It will follow as a separate PR; details in the thread above.

What remains is the thread rewrite plus the three things it directly necessitates: the metrics thread is a daemon, the shutdown join is bounded, and the `nvidia-smi` query has a timeout. Before this change, an unbounded `nvidia-smi` inside a non-daemon `Timer` thread could block interpreter exit, so those belong with the commit that introduces the long-lived thread. `metrics.py` and `test_metrics.py` are the only files touched.

**Renamed** from "correct host CPU and disk utilization metrics" — this PR makes no disk change. That fix shipped separately as #1067 with its own `BREAKING CHANGE` note. Since the squash subject becomes the release-notes line, leaving "disk" in it would have promised a correction that is both absent from this diff and already published.

**Two additions since the last review pass**, both from re-reading the thread lifecycle rather than the GPU code:

- `__enter__` now stops a running thread instead of overwriting the only references to it. Latent today, but see the thread above.
- The first `stop_event.wait` is now pinned by assertion. It was checked with `assert_called_with`, which inspects only the last call, so nothing caught the first wait being shortened. That is the obvious change to make when someone asks why there is no metrics line at startup, and it silently restores the original bug, because a `cpu_percent()` call taken moments after priming reports psutil's near-zero again.

The PR description now enumerates every behaviour change, including three nobody had written down: metrics reliably stop after context exit (previously an already-fired `Timer` survived `cancel()` and rescheduled itself in its `finally`), the thread being a daemon means an in-flight collection can be lost at shutdown, and the four new `WARNING`-level messages.

3,136 passed / 47 skipped, 85% coverage, Ruff and Mypy clean.

leongdl (approved):
Reviewed the full diff at `c12e22a`. Approving.

## The production change is sound

- The core fix is correct: one long-lived thread, `psutil.cpu_percent()` primed on that thread, every sample measured against the prior call on the same thread. That's exactly what the per-thread psutil semantics require.
- The failure paths flagged across review rounds are closed: `thread.start()` degrades to a warning instead of killing the Worker; `_prime_metrics` and each `log_metrics` call are wrapped so the thread can't die silently; the join is bounded at 1s with a daemon thread behind it; `nvidia-smi` has a 5s timeout whose `TimeoutExpired` path skips the sample without latching `_host_has_no_gpu`.
- The stop-event design is careful: each thread gets its own `Event` passed as an argument, so an abandoned thread can't be revived by a successor resetting a shared event, and the re-entry guard reuses the tested `__exit__` path rather than adding a parallel one.
- The `CalledProcessError` latch (permanent GPU-metric disable on one transient failure) is still there — but that's mainline behavior, explicitly deferred to a follow-up with the backoff design already written down in the review thread. Reasonable scoping call.

## The tests are the strongest part

Eleven new/rewritten tests, including a real-thread end-to-end lifecycle test (not mock-only) that distinguishes the priming sample from the measured one, plus tests pinning the abandoned-thread, re-entry, start-failure, and timeout-recovery paths. Mutation-checking is documented with named failing tests per mutant, and the comment in `test_run_primes_baselines_before_logging` pins the design decision (no shortened first wait), not just the code.

## How it works

```
Worker.run()  (worker.py)
│
└─ with host_metrics_logger:          # HostMetricsLogger context manager
   │
   ├─ __enter__()                                    [main thread]
   │   ├─ if self._thread: __exit__()                # re-entry guard
   │   ├─ stop_event = Event()                       # fresh, per-thread
   │   ├─ Thread(target=_run, args=(stop_event,),
   │   │         daemon=True).start()
   │   │     └─ RuntimeError? → warn, no thread      # degrade, don't crash
   │   └─ return self
   │
   │   ┌─────────────────────────────────────────────┐
   │   │ _run(stop_event)          [metrics thread]  │
   │   │   ├─ _prime_metrics()                       │
   │   │   │    ├─ psutil.cpu_percent()   # baseline │
   │   │   │    ├─ net_io_counters()      # baseline │
   │   │   │    ├─ disk_io_counters()     # baseline │
   │   │   │    └─ Exception? → warn, continue       │
   │   │   │                                         │
   │   │   └─ while not stop_event.wait(interval_s): │  ← 60s sleep, wakes
   │   │        ├─ log_metrics()                     │    instantly on set()
   │   │        │    ├─ psutil cpu/mem/disk/net      │
   │   │        │    ├─ _get_gpu_metrics()           │
   │   │        │    │    └─ nvidia-smi              │
   │   │        │    │        timeout=5s             │
   │   │        │    │        TimeoutExpired → {}    │  (no latch)
   │   │        │    └─ logger.info(SYSTEM line)     │
   │   │        └─ Exception? → warn, loop again     │  thread never dies
   │   └─────────────────────────────────────────────┘
   │
   └─ __exit__()                                     [main thread]
       ├─ stop_event.set()            # wakes the wait() immediately
       ├─ thread.join(timeout=1.0)    # bounded — can't block shutdown
       │    └─ still alive? → warn "Abandoning it"
       │        (daemon flag reaps it at process exit)
       └─ self._thread = None
```

Key timing behaviors:

- Normal shutdown: `set()` wakes the idle `wait()` in microseconds; join returns instantly.
- Hung collection (e.g. stuck nvidia-smi): join gives up after 1s, thread abandoned; daemon flag means it can't keep the process alive.
- First metrics line appears at t = interval_s (60s default), not t = 0 — the interval between the prime and the first `cpu_percent()` call is what makes the CPU value real.

AlexTranAmz (approved):

PR #1042, opened 2026-08-07
fix(scheduler): handle start-time mismatch ValidationException gracefully

ZainAallii:
## Problem

When multiple environment-enter actions complete faster than the heartbeat interval (e.g. fast environment-enter scripts in Perforce/UE workflows), the agent re-reports a `startedAt` timestamp that the service already recorded differently from an earlier heartbeat. The service rejects with `ValidationException: Cannot update inactive sessionaction-... because the provided start time X is different from the original start time Y`.

Previously this was treated as unrecoverable, causing a death spiral: the stale entry stays in `_action_updates_map` (commit never called on failure), gets re-sent every heartbeat, and the agent terminates the session.

## Fix

In `_sync()`, catch `DeadlineRequestUnrecoverableError` when the inner exception is a `ValidationException` containing the specific service message about start-time mismatch. On match: log WARNING, call `commit_completed_actions()` to clear the stale entries, return interval=1 for immediate re-sync. Other ValidationExceptions and unrecoverable errors still propagate as before.

## Tradeoff: batch-level commit

`commit_completed_actions()` drops all pending updates in the rejected batch, not just the offending action. This is correct assuming the service rejects the entire batch atomically (the error is a single ValidationException, no partial-success). The service already considers the affected actions inactive/done, and fresh updates will be re-reported on the next heartbeat (~1s later).

## Testing

- Recovery test: scheduler survives, stale entry cleared, returns interval=1.
- Non-matching ValidationException still fatal (not accidentally swallowed).
- Non-ClientError inner_exc still fatal (isinstance guard works).

## Impact

Any customer with multiple fast-completing environment-enter actions (common in Perforce/UE workflows) can hit this. The session is terminated and the job fails despite all actions executing successfully.

### DIFF
--- a/src/deadline_worker_agent/scheduler/scheduler.py
+++ b/src/deadline_worker_agent/scheduler/scheduler.py
@@ -61,6 +61,7 @@ from ..aws.deadline import (
     DeadlineRequestUnrecoverableError,
     update_worker_schedule,
 )
+from botocore.exceptions import ClientError as BotocoreClientError
 from ..config import JobsRunAsUserOverride
 from ..utils import MappingWithCallbacks
 from ..file_system_operations import FileSystemPermissionEnum, make_directory, touch_file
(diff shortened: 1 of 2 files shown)


src/deadline_worker_agent/scheduler/scheduler.py
```diff
@@ -465,8 +466,34 @@ def _sync(self, *, interruptable: bool = True) -> int:
...
+                # their start time after the service already inferred/recorded it from a
+                # previous heartbeat. The actions already completed successfully on the worker;
+                # the service considers them inactive (done). Commit to clear the stale entries
+                # from the map and prevent a death spiral (re-sending the same rejected update).
+                logger.warning(
+                    f"UpdateWorkerSchedule rejected with start-time mismatch "
+                    f"(actions likely completed faster than heartbeat interval): {inner}. "
+                    f"Dropping stale action updates to recover."
+                )
+                commit_completed_actions()
```
epmog, line 494: I'm not convinced by this. Why would fast action times cause this? Nothing here actually explains that, just presume that it's correct. I think doing this is a mistake and is potentially hiding an instance config issue (two workers on one machine?) or some other issue that isn't being fixed.

epmog (changes requested):

src/deadline_worker_agent/scheduler/scheduler.py
ZainAallii: I was able to reproduce the issue using two Workers. You are right. I am going to take this down.

ZainAallii:
This fix is not required for now. The actual cause of the issue is unknown.

PR #1049, opened 2026-08-12
feat: forward OpenJD extensions to template parsing 

seant-aws:

Two problems, both about which OpenJD extensions a job is allowed to use.

The agent never told its template parsers which extensions to accept. `parse_model` gates extension-dependent syntax — wrap-action hooks (`onWrapEnvEnter`/`onWrapTaskRun`/`onWrapEnvExit`) and EXPR format strings — so any template using them was rejected at parse time, before any runtime code ran, on both session runtimes.

Separately, the extension set was hardcoded to every name the model library knows, rather than the set the job actually requested.

Resolve the extension list per job entity from the service and forward it to every parse site.

`BatchGetJobEntity` carries an optional `extensions` list on `JobDetailsEntity`, `StepDetailsEntity`, and `EnvironmentDetailsEntity` (a sibling of `template`, not nested inside it). The agent now reads it and passes it to `parse_model` in both `EnvironmentDetails.from_boto` and `StepDetails.from_boto`, and to the Rust adapter's `decode_environment_template`.

The field is additive-optional and absent until the service deploys it. **Absent resolves to empty** — no extension is enabled for template validation unless the service says the job uses it. Absent is distinguished from an explicitly empty list, so a job that requests nothing is not silently upgraded.

The list is passed through verbatim with no filtering against `ExtensionName`: the service's contract is a pattern-validated string, not a closed enum, so it may legitimately send a name this agent's openjd version does not know. Unrecognized names in `supported_extensions` are inert.

One deliberate asymmetry: the **session-scoped** list (`RUNTIME_CAPABILITY_EXTENSIONS`, feeding the Python `RevisionExtensions` and the Rust `ModelProfile`) stays at the full known set. That value is a runtime *capability* declaration, not a validation gate — `openjd/sessions/_action_filter.py` gates `openjd_redacted_env` environment-variable propagation on whether `REDACTED_ENV_VARS` appears in the session's `RevisionExtensions`, so emptying it would stop that variable reaching subsequent actions. Template validation is gated by the per-entity value; runtime behavior by the session value.

**Dependency floor.** `openjd-model` moves from `>= 0.11.1` to `>= 0.11.2`. The Rust adapter resolves extension names through `ModelExtension.from_str`, and in 0.11.1 the PyO3 binding was missing `WRAP_ACTIONS` — a catch-all arm over the `non_exhaustive` Rust enum silently reported it as `EXPR`, so a profile built with `WRAP_ACTIONS` enabled `EXPR` instead and wrap-hook templates were rejected without the cause being named. This is a floor bump only: the pinned `openjd-sessions` already requires `openjd-model >= 0.11.2`, so the installed set does not change. Declaring it here makes the requirement local to the package that depends on it rather than inherited from an exact pin on another dependency.

Extension-gated templates parse successfully when the service reports the job uses those extensions. No behavior change today: the service currently rejects at `CreateJob` any template declaring these extensions unless the account is enabled for the feature, so no such template can reach a worker yet.

Because the field is additive-optional and absent means empty, no coordination between agent and service rollout is required, and nothing needs cleaning up when the service ships.

Unit tests cover the resolver (absent, present, explicitly empty, and an unrecognized name passing through), both `from_boto` parse sites including the legacy `StepScript` step shape, and the Rust adapter's extension forwarding. Wrap-action templates are asserted to parse through the OpenJD model at both the environment and step level.

Build, unit suite, lint (ruff + format + mypy), and the integration suite all pass. Each commit passes standalone.

The resolver's absent-versus-empty distinction is verified by mutation rather than assertion alone: replacing the absent-key return with the session-scoped capability list fails `test_key_absent_returns_empty` and `test_key_none_returns_empty`, and both pass once reverted.

Wrap-action differential and end-to-end tests are deliberately held for a follow-up PR — they depend on this change, so they cannot be verified against `mainline` until it lands.

No user-facing documentation change needed — this enables existing OpenJD extension semantics that an implementation gap was blocking.

No.


src/deadline_worker_agent/sessions/session.py
```diff
@@ -220,10 +220,7 @@ def openjd_session_action_callback(session_id: str, action_status: ActionStatus)
                 os_env_vars=self._env,
                 session_root_directory=session_root_dir,
                 spec_revision="2023-09",
-                # Currently for simplicity request that our session allow all extensions
-                # This does not obey the spec.  It should be changed at a later date to the list of requested
-                # extensions once those are returned by BatchGetJobEntity
-                supported_extensions=tuple(v.value for v in ExtensionName),
+                supported_extensions=RUNTIME_CAPABILITY_EXTENSIONS,
```
seant-aws, line 223: I know the old comment says we can replace it once backend catches up, but I think we should swap this once we think backend is flipped to a good state. for now, the service still gates extensions, but runtime does this

wyongzhi:
I think we should bump the openjd-model to 0.11.2 for this PR to deliver what it intends on the Rust runtime. 0.11.1 has a binding bug in PyModelExtension (the enum shipped only 4 variants, missing WRAP_ACTIONS), so ModelExtension.from_str("WRAP_ACTIONS") returns the EXPR variant instead of a distinct one — verified against both wheels:
```
    0.11.1:  from_str("WRAP_ACTIONS") -> EXPR   (same object as from_str("EXPR"))
    0.11.2:  from_str("WRAP_ACTIONS") -> WRAP_ACTIONS
```

seant-aws:
> I think we should bump the openjd-model to 0.11.2 for this PR to deliver what it intends on the Rust runtime. 0.11.1 has a binding bug in PyModelExtension (the enum shipped only 4 variants, missing WRAP_ACTIONS), so ModelExtension.from_str("WRAP_ACTIONS") returns the EXPR variant instead of a distinct one — verified against both wheels:
> 
> ```
>     0.11.1:  from_str("WRAP_ACTIONS") -> EXPR   (same object as from_str("EXPR"))
>     0.11.2:  from_str("WRAP_ACTIONS") -> WRAP_ACTIONS
> ```

good catch, bumping here; I'll also have a followup for auditing and bumping all the relevant versions across openjd/worker agent

wyongzhi:
resolve_supported_extensions has no direct unit tests — the description says "Unit tests cover the resolver (absent, present, explicitly empty, and an unrecognized name passing through)", but the diff exercises it only indirectly through from_boto, and no unrecognized-name case anywhere. Since absent-vs-explicitly-empty is the load-bearing distinction of this module, a small test file locking the three branches (key absent → (), value None → (), list incl. unknown names → verbatim) seems worth it. Either way the description's test claim should be updated to match what actually ships. (ai assisted comment)

seant-aws:
Both addressed — thanks, and the second one was a real regression on this branch rather than just a description error.

**openjd-model floor** — agreed, and done in `22018e3` (pushed a minute after your comment, so we crossed; we reached the same conclusion independently and my probe matches yours exactly: `0.11.1` gives `from_str("WRAP_ACTIONS") -> EXPR`, `0.11.2` gives `WRAP_ACTIONS`). Root cause for the record: `PyModelExtension` shipped four variants, and the catch-all arm in `From<ModelExtension> for PyModelExtension` over the `non_exhaustive` Rust enum silently reported the missing one as `EXPR`. Note this is a floor bump only — the exact `openjd-sessions == 0.10.14` pin already requires `openjd-model >= 0.11.2`, so the installed set does not change; the point is making the requirement local to the package that depends on it rather than inherited from another dependency's pin.

**`resolve_supported_extensions` had no direct tests** — correct, and worse than a stale description: the test file existed in an earlier revision of this PR and was deleted during a rebase on this branch, which is why the description described tests that were no longer shipping. Restored in `3419397` with all five branches you asked for — key absent, value explicitly `None`, non-empty list verbatim, explicitly empty list, and a name unknown to `ExtensionName` passing through unfiltered.

Since red-before-green does not apply to tests added for pre-existing behavior, I verified sensitivity by mutation instead of asserting it: replacing the absent-key return with the session-scoped capability list fails `test_key_absent_returns_empty` and `test_key_none_returns_empty` (`2 failed, 3 passed`), and all five pass once reverted. So the absent-versus-empty distinction is genuinely pinned, not just covered.

Gates on the new tip: build, 3.9 unit 3059 passed / 39 skipped, lint clean, integ 85 passed / 4 skipped.

wyongzhi (dismissed):

AlexTranAmz (dismissed):

leongdl (approved):

wyongzhi (approved):

PR #1050, opened 2026-08-12
refactor!: Remove cross-region CWL routing from log config options

MarinMaksutaj:

#997 (a3ce25c) added cross-region CloudWatch Logs routing: the worker agent reads a region from the log configuration options and routes session logs to that region. This code is redundant: workers already route session logs to the home region by default today, so the extra region handling is not needed.

Remove the cross-region routing added in #997. This is mostly a revert, but not a pure one, which is why it is typed refactor! rather than revert:

Removed (from #997):
- The region key in the log configuration options and all handling of it
- Region parameter threading through construct_worker_log_config, LogConfiguration, and the CloudWatch Logs client creation in the entrypoint
- The unit tests covering the region option

Kept (from #997):
- The log_constants.py extraction (minus the region key). #997 moved the log config option key constants out of log_sync/cloudwatch.py into a leaf module, which broke a module-level cyclic import chain (sessions/log_config -> log_sync/cloudwatch -> scheduler). A pure revert reintroduces that cycle and CodeQL flags it with 5 errors. cloudwatch.py re-exports the constants for backwards compatibility.

The worker agent no longer honors a region in the log configuration options; CloudWatch Logs clients use the default routing.

Unit tests (3047 passed), ruff check, ruff format, and mypy all pass.

No documentation changes needed; the removed option was not documented.

Yes.

BREAKING CHANGE: the worker agent no longer honors a region in the log configuration options; CloudWatch Logs clients use the default routing.

----


leongdl (approved):

godobyte (changes requested):
Please check prior revert PR and follow the revert commit convention. The title should show up as revert with the prior commit.

e.g. https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/813

godobyte:
Could you update PR description to  follow the PR template.

godobyte (approved):

PR #1051, opened 2026-08-13
fix: resolve environment job parameter references on the Rust runtime

seant-aws:

On the Rust session runtime, any OpenJD environment whose script referenced a job parameter failed to enter:

```
ModelValidationError: 1 validation error for EnvironmentTemplate
environment -> script -> actions -> onEnter -> args[0]:
Failed to parse interpolation expression. Undefined variable: 'Param.Message'
```

`RustSessionRuntime.enter_environment` lifts the environment out of its job and decodes it as a standalone environment template. That template omitted the job's `parameterDefinitions`, so the decoder correctly reported the parameter as undefined. The Python runtime is unaffected — it never decodes the environment itself, and hands the pydantic model to the openjd session which resolves format strings later with the job parameters already in scope.

The template now declares the job's parameters, synthesized from the parameter values the adapter already receives and retained at construction. Declaring a parameter does not oblige the caller to supply a value, and `create_environment` strips `parameterDefinitions`, so nothing is exposed downstream. The key is omitted entirely when there is nothing to declare, because an empty `parameterDefinitions` list is rejected by the schema.

No type filter is applied: `decode_environment_template` was empirically verified to accept every `JobParameterType` member. Only task-parameter-space types such as `CHUNK[INT]` are rejected, and those cannot reach this path.

Environments referencing `{{Param.X}}` now enter successfully on the Rust runtime. No change to the Python runtime.

Known limitation: only parameters that have values are declared, so a job parameter with no value supplied would still fail validation. In practice everything reaching the worker has a default or a submitted value.

A new differential integration test exercises the real decoder through the adapter — it fails on the Rust runtime without this change and passes on the Python runtime either way. Unit tests cover the multi-parameter, non-primitive type, empty, and mixed-value-shape cases. Build, unit suite, lint, and the integration suite all pass.

**Falsifiability (mutation run).** Disabling the `parameterDefinitions` injection in `enter_environment` makes the integration test fail by name on the Rust runtime only:

```
PASSED  test_enter_environment_when_job_has_typed_params_resolves_param_reference[python]
FAILED  test_enter_environment_when_job_has_typed_params_resolves_param_reference[rust]
  E   Failed to parse interpolation expression at [7, 24]. Undefined variable: 'Param.Message'.
```

Restored, both parametrizations pass. This confirms `env_template.environment` strips the definitions, so the test genuinely depends on the fix.

No user-facing documentation change needed — this fixes a runtime gap in existing OpenJD semantics.

No.

### DIFF
--- a/src/deadline_worker_agent/sessions/runtime/rust.py
+++ b/src/deadline_worker_agent/sessions/runtime/rust.py
@@ -157,6 +157,33 @@ def _to_rust_task_parameter_values(values: dict[str, Any]) -> dict[str, Any]:
     )
 
 
+def _to_environment_parameter_definitions(values: dict[str, Any]) -> list[dict[str, str]]:
+    """Synthesize parameterDefinitions entries from job parameter values.
+
+    The environment template is decoded standalone (lifted out of its job
+    context), so the job's parameter declarations must be re-supplied here for
+    ``{{Param.X}}`` references to resolve. Every ``JobParameterType`` member is a
+    valid environment-template parameter type — only task-parameter-space types
+    like ``CHUNK[INT]`` are rejected, and those cannot reach here: ParameterValue
+    objects are constrained by ``JobParameterType`` at session construction, and
+    raw dicts originate from the wire which only carries job parameter types.
+
+    No type filter is applied. All values are declared unconditionally.
+
+    Known limitation: only parameters that have *values* are declared. A job
+    parameter with no value supplied would still fail validation at decode time.
+    In practice everything reaching the worker has a default or submitted value.
+    """
+    definitions: list[dict[str, str]] = []
+    for name, value in values.items():
+        if isinstance(value, dict):
+            type_str = value["type"]
+        else:
+            type_str = value.type.value
+        definitions.append({"name": name, "type": type_str})
+    return definitions
+
+
 def _to_rust_path_mapping_rule(rule: PathMappingRule) -> RustPathMappingRule:
     """Convert one worker (v0) PathMappingRule into the _v1 (openjd.expr) type.
 
@@ -202,6 +229,7 @@ class RustSessionRuntime(SessionRuntime):
 
     _session: OpenJDRustSession
     _user: Optional[SessionUser]
+    _environment_parameter_definitions: list[dict[str, str]]
 
     def __init__(self, config: SessionRuntimeConfig) -> None:
         try:
@@ -236,6 +264,13 @@ class RustSessionRuntime(SessionRuntime):
         # (which runs as that user) can read them.
         self._user = config.user
 
+        # Job parameter values are fixed for the session's lifetime, so the
+        # declarations are synthesized once here. See
+        # _to_environment_parameter_definitions for why they are needed at all.
+        self._environment_parameter_definitions = _to_environment_parameter_definitions(
+            config.job_parameter_values
+        )
+
         # The _v1 session reports status with _v1 ActionStatus/ActionState, but
         # the worker's callback expects the v0 types, so translate on the way out.
         v0_callback = config.action_callback
@@ -268,16 +303,16 @@ class RustSessionRuntime(SessionRuntime):
         # wire shape and rebuild it natively. exclude_none=True is required, not
         # cosmetic: the Rust decoder rejects explicit nulls (OpenJD treats
         # absent and null as equivalent).
-        native_environment = create_environment(
-            decode_environment_template(
-                {
-                    "specificationVersion": "environment-2023-09",
-                    "environment": environment.model_dump(
-                        mode="json", by_alias=True, exclude_none=True
-                    ),
-                }
-            )
-        )
+        template: dict[str, Any] = {
+            "specificationVersion": "environment-2023-09",
+            "environment": environment.model_dump(mode="json", by_alias=True, exclude_none=True),
+        }
+        # The job's parameter declarations must accompany the environment or its
+        # {{Param.X}} references cannot resolve. The key is omitted entirely when
+        # empty, because the schema rejects "parameterDefinitions": [].
+        if self._environment_parameter_definitions:
+            template["parameterDefinitions"] = self._environment_parameter_definitions
+        native_environment = create_environment(decode_environment_template(template))
         return self._session.enter_environment(
             environment=native_environment,
             identifier=identifier,
--- a/src/deadline_worker_agent/sessions/runtime/rust.py
+++ b/src/deadline_worker_agent/sessions/runtime/rust.py
@@ -164,9 +164,10 @@ def _to_environment_parameter_definitions(values: dict[str, Any]) -> list[dict[s
     context), so the job's parameter declarations must be re-supplied here for
     ``{{Param.X}}`` references to resolve. Every ``JobParameterType`` member is a
     valid environment-template parameter type — only task-parameter-space types
-    like ``CHUNK[INT]`` are rejected, and those cannot reach here: ParameterValue
-    objects are constrained by ``JobParameterType`` at session construction, and
-    raw dicts originate from the wire which only carries job parameter types.
+    like ``CHUNK[INT]`` are rejected, and those cannot reach here:
+    ``JobDetails._validate_job_parameters`` restricts jobDetails parameters to
+    the job-parameter-type subset (string/path/int/float), and raw dicts
+    originate from the wire which only carries job parameter types.
 
     No type filter is applied. All values are declared unconditionally.
 
@@ -177,7 +178,14 @@ def _to_environment_parameter_definitions(values: dict[str, Any]) -> list[dict[s
     definitions: list[dict[str, str]] = []
     for name, value in values.items():
         if isinstance(value, dict):
-            type_str = value["type"]
+            type_str = value.get("type")
+            if type_str is None:
+                # Skip malformed entries rather than raising — the same raw dict
+                # is passed to _to_rust_job_parameter_values moments later, so a
+                # genuinely invalid value still fails at Rust session construction
+                # with the session's own clear error. This matches the sibling
+                # helper's convention of deferring dict validation to the Rust session.
+                continue
         else:
             type_str = value.type.value
         definitions.append({"name": name, "type": type_str})
--- a/test/integ/sessions/test_differential_runtime.py
+++ b/test/integ/sessions/test_differential_runtime.py
@@ -381,3 +381,74 @@ class TestDifferentialSessionRuntime:
         _cancel_when_registered(runtime)
 
         assert _wait_for_terminal(runtime) == "CANCELED"
+
+    @pytest.mark.timeout(60)
+    @pytest.mark.parametrize("runtime_kind", _RUNTIME_KINDS)
+    def test_enter_environment_when_job_has_typed_params_resolves_param_reference(
+        self, runtime_kind: SessionRuntimeKind, tmp_path: Path
+    ) -> None:
+        """An environment referencing {{Param.X}} succeeds when the runtime is
+        constructed with non-empty job_parameter_values including non-STRING
+        types. This exercises the REAL decoder through the adapter — unit tests
+        mock it and cannot validate payload shape."""
+        from openjd.model._types import ParameterValue, ParameterValueType
+
+        root = tmp_path / f"session-params-{runtime_kind.name.lower()}"
+        root.mkdir(parents=True, exist_ok=True)
+        recorder = _StatusRecorder()
+        config = SessionRuntimeConfig(
+            session_id=f"session-params-{runtime_kind.name.lower()}",
+            job_parameter_values={
+                "Message": ParameterValue(type=ParameterValueType.STRING, value="hello-world"),
+                "Count": ParameterValue(type=ParameterValueType.INT, value="42"),
+            },
+            path_mapping_rules=None,
+            retain_working_dir=False,
+            user=None,
+            action_callback=recorder,
+            os_env_vars=None,
+            session_root_directory=root,
+        )
+        runtime = create_session_runtime(runtime_kind, config)
+        try:
+            # The environment's onEnter script references {{Param.Message}} — the
+            # decoder must resolve it using the parameterDefinitions we supply.
+            env_template = decode_environment_template(
+                template={
+                    "specificationVersion": "environment-2023-09",
+                    "environment": {
+                        "name": "ParamEnv",
+                        "script": {
+                            "actions": {
+                                "onEnter": {
+                                    "command": sys.executable,
+                                    "args": ["-c", "print('{{Param.Message}}')"],
+                                }
+                            }
+                        },
+                    },
+                    "parameterDefinitions": [
+                        {"name": "Message", "type": "STRING"},
+                        {"name": "Count", "type": "INT"},
+                    ],
+                }
+            )
+            environment = env_template.environment
+
+            prev = len(recorder.statuses)
+            identifier = runtime.enter_environment(environment=environment)
+            assert _wait_for_new_action(recorder, prev)
+            terminal = _wait_for_terminal(runtime)
+            assert terminal == "SUCCESS", (
+                f"Expected SUCCESS but got {terminal}; statuses={recorder.state_names}"
+            )
+
+            prev = len(recorder.statuses)
+            runtime.exit_environment(identifier=identifier)
+            assert _wait_for_new_action(recorder, prev)
+            assert _wait_for_terminal(runtime) == "SUCCESS"
+        finally:
+            try:
+                runtime.cleanup()
+            except Exception:
+                pass
--- a/test/unit/sessions/runtime/test_rust.py
+++ b/test/unit/sessions/runtime/test_rust.py
@@ -20,6 +20,7 @@ from deadline_worker_agent.sessions.runtime._abc import SessionRuntimeCrashError
 from deadline_worker_agent.sessions.runtime import rust as rust_module
 from deadline_worker_agent.sessions.runtime.rust import RustSessionRuntime
 from deadline_worker_agent.sessions.runtime.rust import _to_rust_task_parameter_values
+from deadline_worker_agent.sessions.runtime.rust import _to_environment_parameter_definitions
 
 
 @pytest.fixture()
@@ -238,11 +239,14 @@ class TestRustSessionRuntimeDelegation:
             )
 
         # The pydantic environment is serialized and rebuilt natively before
-        # being handed to the session.
+        # being handed to the session. The fixture's job_parameter_values
+        # contains one STRING param in dict form, which must appear as a
+        # parameterDefinitions entry.
         mock_decode.assert_called_once_with(
             {
                 "specificationVersion": "environment-2023-09",
                 "environment": environment.model_dump.return_value,
+                "parameterDefinitions": [{"name": "Param1", "type": "STRING"}],
             }
         )
         environment.model_dump.assert_called_once_with(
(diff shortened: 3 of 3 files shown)


seant-aws:
Split out of #1049, which now carries only the OpenJD extensions work. These two PRs are independent — this branch was verified to pass standalone without the extensions commits, and vice versa — so they can merge in either order.

seant-aws:
### Why this change is needed

```python
repr_sh(flatten([["-e", e] for e in WrappedAction.Environment]))
```

This is the natural idiom for passing environment variables to `docker run`. It works when the wrapped action has env vars. When it has **none** (empty list), it fails:

```
No matching signature for repr_sh(list[int])
```

Root cause: `flatten_fn` at `list.rs:68` defaulted an empty result's element type to `ExprType::INT`. Since `repr_sh` declares overloads for `list[string]`, `list[path]`, `list[nulltype]` but NOT `list[int]`, overload resolution fails.

### How it changes things

**Before:** Any expression using `flatten()` over an empty comprehension result, followed by a function without a `list[int]` overload, fails with a type error that names a type nothing in the template mentions. Data-dependent: same template, same fleet, outcome decided by whether the wrapped action happened to set env vars.

**After:** Empty flatten result types as `NULLTYPE` (the bottom type for empty lists). `repr_sh` already declares `(list[nulltype]) -> string`, so overload resolution succeeds and returns an empty string.

leongdl (commented):
## Review: understanding and code flow

### The contract

The worker agent can run jobs on two interchangeable engines: the Python session runtime and the Rust session runtime. Customers don't choose the engine, so a job that works on one must work identically on the other — any behavioral gap between the two is a bug.

### Why this PR exists

A job can have job parameters (named values like `Message = "hello"`), referenced in scripts as `{{Param.Message}}`. A job can also have environments — setup/teardown scripts that run around the work. On the Rust runtime, any environment whose script used `{{Param.Message}}` refused to enter with `Undefined variable: 'Param.Message'`. The same job ran fine on the Python runtime.

### Mechanism of the bug

The Rust engine can't consume the pydantic objects directly, so `RustSessionRuntime.enter_environment` serializes the environment into a **standalone** environment-template document and asks the Rust decoder to rebuild it. An environment lifted out of its job loses the job's parameter declarations: the standalone document says "here is a script using `Param.Message`" but never says "a parameter named `Message` exists." The Rust decoder validates references at decode time, saw an undeclared name, and correctly rejected it. The decoder wasn't wrong — the adapter was handing it an incomplete document.

The Python runtime structurally cannot have this bug: `PythonSessionRuntime.enter_environment` is a one-line delegation with no serialize/decode round-trip, and parameters resolve later with the job's parameters already in scope (verified by reading `python.py` at mainline).

### The fix, mechanically

The adapter already receives every job parameter's name and type at construction. The fix re-derives the missing declarations from that: `__init__` builds `[{"name": ..., "type": ...}]` once, and `enter_environment` attaches it to the standalone document under `parameterDefinitions` before decoding. Two edge details handled: the key is omitted entirely when empty (the schema rejects `"parameterDefinitions": []`), and no type filter is applied (all `JobParameterType` members are claimed valid environment-template types; task-only types like `CHUNK[INT]` can't reach this path).

### Code flow

```
Job assigned to worker
  │
  ▼
RustSessionRuntime.__init__(config)                      [rust.py]
  │  config.job_parameter_values = {"Message": ParameterValue(STRING, "hello-world"), ...}
  │
  ├─ NEW: _to_environment_parameter_definitions(values)
  │     for each param: type from dict["type"] or ParameterValue.type.value
  │     → self._environment_parameter_definitions = [{"name": "Message", "type": "STRING"}]
  │
  └─ OpenJDRustSession(...) constructed with the param VALUES (unchanged)

Scheduler starts an environment
  │
  ▼
RustSessionRuntime.enter_environment(environment)        [rust.py]
  │
  ├─ environment.model_dump(json, by_alias, exclude_none)   ← pydantic → wire JSON
  │
  ├─ template = {"specificationVersion": "environment-2023-09",
  │              "environment": {...script with {{Param.Message}}...}}
  │
  ├─ NEW: if self._environment_parameter_definitions:
  │          template["parameterDefinitions"] = [...]       ← the fix
  │
  ├─ decode_environment_template(template)
  │     BEFORE: "Undefined variable: Param.Message" → ModelValidationError
  │     AFTER:  declaration present → reference resolves → decodes fine
  │
  ├─ create_environment(decoded)                            ← strips parameterDefinitions
  │
  └─ self._session.enter_environment(native_environment)
        Rust session resolves {{Param.Message}} → "hello-world" and runs the script
```

### Scope of this review

Verified by reading at tip `0ee0607`: all three changed files in full context, and the Python runtime reference implementation. Not run locally: the test suite, mutation testing, and the CI matrix — findings that depend on those are labeled as unverified in the inline comments. Individual findings are posted inline at the lines they apply to. All are non-blocking.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -157,6 +157,33 @@ def _to_rust_task_parameter_values(values: dict[str, Any]) -> dict[str, Any]:
...
+    No type filter is applied. All values are declared unconditionally.
+
+    Known limitation: only parameters that have *values* are declared. A job
+    parameter with no value supplied would still fail validation at decode time.
+    In practice everything reaching the worker has a default or submitted value.
+    """
+    definitions: list[dict[str, str]] = []
+    for name, value in values.items():
+        if isinstance(value, dict):
+            type_str = value["type"]
```
leongdl: **Finding R1 (non-blocking, low): dict-shaped values missing `"type"` now raise `KeyError` in `__init__`.**

A wire-form dict without a `"type"` key raises `KeyError` (a `LookupError`) here, and this now runs at construction time — before the Rust session's own validation would have caught the malformed value. `__init__` is not wrapped by `@convert_runtime_crashes` (only the methods are), so this is a new exception type on a new path.

Contrast with `_to_rust_parameter_values` just above, which passes dicts through untouched and lets the Rust session validate them.

Trigger requires a malformed wire payload, and as far as I can tell the wire always carries `"type"` (unverified against the BatchGetJobEntity schema) — so low severity. A `value.get("type")` with an explicit error naming the parameter, or simply skipping the malformed entry to preserve the old failure point, would close it.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -236,6 +264,13 @@ def __init__(self, config: SessionRuntimeConfig) -> None:
         # (which runs as that user) can read them.
         self._user = config.user
 
+        # Job parameter values are fixed for the session's lifetime, so the
+        # declarations are synthesized once here. See
+        # _to_environment_parameter_definitions for why they are needed at all.
+        self._environment_parameter_definitions = _to_environment_parameter_definitions(
```
leongdl, line 278: **Finding R2 (non-blocking, low): the snapshot assumes `job_parameter_values` never changes after construction.**

The comment claims job parameter values are fixed for the session's lifetime, which is plausible (Deadline job params are per-job), but I did not audit every writer of the config to verify it. If anything ever mutated `config.job_parameter_values` after construction, `enter_environment` would silently use stale declarations. If the invariant holds — and I expect it does — this is fine as-is; flagging so a maintainer can confirm.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -268,16 +303,16 @@ def enter_environment(
...
+        template: dict[str, Any] = {
+            "specificationVersion": "environment-2023-09",
+            "environment": environment.model_dump(mode="json", by_alias=True, exclude_none=True),
+        }
+        # The job's parameter declarations must accompany the environment or its
+        # {{Param.X}} references cannot resolve. The key is omitted entirely when
+        # empty, because the schema rejects "parameterDefinitions": [].
+        if self._environment_parameter_definitions:
+            template["parameterDefinitions"] = self._environment_parameter_definitions
+        native_environment = create_environment(decode_environment_template(template))
```
leongdl, line 323: **Finding R3 (question, likely not real): does the sibling `run_task` path have the same class of bug?**

`run_task` does the same standalone-decode trick via `deserialize_step`. If `deserialize_step` validates `{{Param.X}}` references at decode time the way `decode_environment_template` does, step scripts referencing job parameters would fail identically — and this PR fixes only the environment half.

Counter-evidence: step scripts referencing job params are ubiquitous, so if that were broken the Rust runtime would be visibly unusable. Most likely `deserialize_step` defers resolution to the session (which holds `job_parameter_values`). Unverified either way — worth one targeted differential test with a step script containing `{{Param.X}}`, or a maintainer's confirmation of why decode-time validation differs between the two decoders.

test/integ/sessions/test_differential_runtime.py
```diff
@@ -381,3 +381,74 @@ def test_cancel_mid_action_ends_canceled(
...
+                            }
+                        },
+                    },
+                    "parameterDefinitions": [
+                        {"name": "Message", "type": "STRING"},
+                        {"name": "Count", "type": "INT"},
+                    ],
+                }
+            )
+            environment = env_template.environment
```
leongdl, line 436: **Test-coverage note: this is the one mutation to run first.**

The test builds its input by passing `parameterDefinitions` directly to `decode_environment_template`, then extracts `.environment` and relies on the adapter to re-supply the definitions. That's correct **only if** `env_template.environment` truly strips the definitions — if any parameter context survives on the extracted environment, this test could pass without the fix.

Suggested verification: revert the `parameterDefinitions` injection in `rust.py:enter_environment` and confirm the RUST parametrization of this test fails by name (and the PYTHON one still passes). If that mutation has already been run, noting it in the PR description would make the falsifiability explicit.

test/unit/sessions/runtime/test_rust.py
```diff
@@ -256,6 +259,44 @@ def test_enter_environment_when_called_converts_env_and_delegates(
         )
         assert result is mock_session_instance.enter_environment.return_value
 
+    def test_enter_environment_includes_all_job_parameter_types(
```
leongdl, line 263: **Coverage gap: no test drives a dict-shaped (wire-form) parameter value through the new path.**

`_to_environment_parameter_definitions` has two branches — `ParameterValue` objects (`value.type.value`) and raw dicts (`value["type"]`) — but the unit tests here only exercise `ParameterValue` objects. The PR description mentions a "mixed-value-shape" case, but I could not find it in the diff. A small addition passing `{"A": {"type": "STRING", "value": "x"}}` alongside a `ParameterValue` would pin the dict branch (and is where the R1 `KeyError` behavior would be asserted, whichever way it's resolved).

leongdl:
## Follow-up: spec and reference-implementation cross-check for R2 and R3

I cross-checked findings R2 and R3 from my review against the OpenJD specification and the `openjd-model-for-python` Rust bindings. Both are resolved. R1 stands.

### R2 — resolved, invariant confirmed by the spec

The [2023-09 Template Schemas, §7.4 Template Processing Stages](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas) defines exactly three stages: template validation, **job creation** ("Parameter values are bound"), and task execution. `Param.*` values move from *unknown* to *known* at job creation and are never rebound — at task execution they are listed among the already-known values, with no mechanism in the spec for changing them afterward. A session belongs to one job, so job parameter values are constant for the session's lifetime by specification, not just by convention. The comment in `__init__` ("Job parameter values are fixed for the session's lifetime") is spec-accurate, and computing the declarations once at construction is correct. **Withdrawing the concern.**

### R3 — withdrawn, the asymmetry is by design

I read [`rust-bindings/src/model/create_job_fns.rs` in openjd-model-for-python](https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/rust-bindings/src/model/create_job_fns.rs). The two decoders are not siblings — they operate on different sides of the template/job divide:

- `py_deserialize_step` deserializes a **job-side** `openjd_model::job::Step` via plain `serde_json::from_str`, with no template validation of any kind. Its doc comment is explicit: the `GetStepDetails`/`BatchGetJobEntity` payload "is a serialized `openjd_model::job::Step`, **not a template-side StepTemplate**." A job-side Step is a post-instantiation entity — its format strings carry `{{Param.X}}` unresolved by design, resolved later by the session, which holds `job_parameter_values`.
- `decode_environment_template` decodes a **template-side** document, and template decoding is where reference validation lives — which is exactly why the environment path failed and the task path never did.

So `run_task` cannot have this class of bug, and this PR's scope (environment path only) is complete. This also explains why the fix is shaped the way it is: `enter_environment` must go through a template-side decode (there is no job-side environment deserializer exposed), so it must satisfy template-side validation, and supplying `parameterDefinitions` is the spec-sanctioned way to do that — [§1.2 Environment Template](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas) defines `parameterDefinitions` as exactly this: job parameter declarations carried by a standalone environment template. The spec also confirms the empty-list omission is required ("If provided, then this list must contain at least one element").

### Updated verdict

With R2 and R3 off the table, the only remaining finding is **R1** (the `KeyError` in `_to_environment_parameter_definitions` for a dict-shaped value missing `"type"`, raised at `__init__` outside `@convert_runtime_crashes`) — real but low severity, and cheap to close with a `.get()` plus an explicit error, ideally alongside a unit test covering the dict-shaped value branch.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl: **Withdrawn — invariant confirmed against the OpenJD specification.** (Thread resolved.)

What we checked: the [2023-09 Template Schemas, §7.4 Template Processing Stages](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas). The spec defines exactly three processing stages — template validation, job creation ("Parameter values are bound"), and task execution — and `Param.*` values move from *unknown* to *known* at job creation with no mechanism for rebinding afterward. A session belongs to one job, so job parameter values are constant for the session's lifetime **by specification**, not just by convention.

Conclusion: the comment above ("Job parameter values are fixed for the session's lifetime") is spec-accurate, and computing the declarations once in `__init__` is correct. No change needed.

src/deadline_worker_agent/sessions/runtime/rust.py
leongdl: **Withdrawn — the asymmetry is by design, `run_task` cannot have this bug class.** (Thread resolved.)

What we checked: [`rust-bindings/src/model/create_job_fns.rs` in openjd-model-for-python](https://github.com/OpenJobDescription/openjd-model-for-python/blob/mainline/rust-bindings/src/model/create_job_fns.rs). The two decoders are not siblings — they operate on opposite sides of the template/job divide:

- `py_deserialize_step` deserializes a **job-side** `openjd_model::job::Step` via plain `serde_json::from_str`, with **no template validation of any kind**. Its doc comment is explicit: the `GetStepDetails`/`BatchGetJobEntity` payload "is a serialized `openjd_model::job::Step`, **not a template-side StepTemplate**." A job-side Step is post-instantiation — its format strings carry `{{Param.X}}` unresolved by design, and the session (which holds `job_parameter_values`) resolves them at run time.
- `decode_environment_template` decodes a **template-side** document, and template decoding is where reference validation lives — which is exactly why the environment path failed and the step path never did.

Conclusion: this PR's scope (environment path only) is complete. It also confirms `parameterDefinitions` is the right fix rather than a workaround — [§1.2 Environment Template](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas) defines it as exactly this: job parameter declarations carried by a standalone environment template, with the empty-list omission required by the schema ("If provided, then this list must contain at least one element").

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -268,16 +303,16 @@ def enter_environment(
...
-            decode_environment_template(
-                {
-                    "specificationVersion": "environment-2023-09",
-                    "environment": environment.model_dump(
-                        mode="json", by_alias=True, exclude_none=True
-                    ),
-                }
-            )
-        )
+        template: dict[str, Any] = {
```
leongdl, line 314: Nit for the future, I hope we can stick with all rust types and none of these ad-hoc declarations

src/deadline_worker_agent/sessions/runtime/rust.py
seant-aws: Fixed in `5d1d52f`.

Went with **skipping** rather than an explicit error, for the reason you identified: the same raw dict reaches `_to_rust_job_parameter_values` moments later, so a genuinely malformed value still fails at Rust session construction with the session's own clear error. Skipping preserves that pre-existing failure point and matches the sibling helper's convention of deferring dict validation to the Rust session.

Added `test_enter_environment_when_dict_param_missing_type_key_is_skipped`, confirmed red-before-green — `KeyError: 'type'` at `rust.py:180` before the fix, passes after.

Also corrected the docstring while in there: you are right that nothing constrains `ParameterValue` construction. The actual guard is `JobDetails._validate_job_parameters` restricting jobDetails parameters to `string`/`path`/`int`/`float`, and the docstring now names that instead of the invariant it previously claimed.

test/unit/sessions/runtime/test_rust.py
seant-aws: The dict branch is covered in two places — `TestToEnvironmentParameterDefinitions::test_extracts_type_from_dict_form` for the helper in isolation, and `test_enter_environment_when_mixed_dict_and_object_params_both_appear` for a dict alongside a `ParameterValue` through the adapter. That second one is the "mixed-value-shape" case the description refers to.

Worth flagging why it may not have been visible: the branch was briefly force-pushed to a rewritten base earlier today, which dropped several of these tests. That has been reverted — `26817df` is the base again and all of them are present, with the missing-`type` case added on top per R1. Apologies for the churn.

test/integ/sessions/test_differential_runtime.py
seant-aws: Ran it. Disabling the `parameterDefinitions` injection in `enter_environment` (replacing the assignment with `pass`):

```
PASSED  test_enter_environment_when_job_has_typed_params_resolves_param_reference[python]
FAILED  test_enter_environment_when_job_has_typed_params_resolves_param_reference[rust]
  E   Failed to parse interpolation expression at [7, 24]. Undefined variable: 'Param.Message'.
1 failed, 1 passed, 89 deselected
```

Restored: `2 passed, 89 deselected`.

So `env_template.environment` does strip the definitions — the test cannot pass without the fix, and it fails by name on `[rust]` only. Noting it in the PR description as you suggested.

src/deadline_worker_agent/sessions/runtime/rust.py
seant-aws: Agreed — and this is a live thread rather than just a wish.

The direction you are describing was raised on #1002 as a follow-up: have the adapter default to `_v1` types and bridge `_v1` -> v0 in the *Python* runtime, so that retiring the Python path is a deletion rather than a growing v0 -> `_v1` compat layer. It was implemented in #1028 (new `_v0_compat.py`, scalars crossing the interface as `_v1`-native, `rust.py` reduced to near pass-through), but that PR was closed while still in draft, so the work is currently unowned.

These ad-hoc dict declarations are exactly the kind of thing that disappears under it. Happy to revive it as a standalone PR if you think it is worth picking back up.

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -157,6 +157,33 @@ def _to_rust_task_parameter_values(values: dict[str, Any]) -> dict[str, Any]:
...
+def _to_environment_parameter_definitions(values: dict[str, Any]) -> list[dict[str, str]]:
+    """Synthesize parameterDefinitions entries from job parameter values.
+
+    The environment template is decoded standalone (lifted out of its job
+    context), so the job's parameter declarations must be re-supplied here for
+    ``{{Param.X}}`` references to resolve. Every ``JobParameterType`` member is a
+    valid environment-template parameter type — only task-parameter-space types
+    like ``CHUNK[INT]`` are rejected, and those cannot reach here: ParameterValue
+    objects are constrained by ``JobParameterType`` at session construction, and
+    raw dicts originate from the wire which only carries job parameter types.
```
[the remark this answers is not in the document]
seant-aws: Correct, and fixed in `5d1d52f`. There is no constraint on `ParameterValue` construction — `config.job_parameter_values` comes from `parameters_from_api_response`, the same parser used for task parameters, so a `CHUNK_INT` value is constructible. The docstring now names the guard that actually applies, `JobDetails._validate_job_parameters` restricting jobDetails parameters to `string`/`path`/`int`/`float`, rather than the invariant it previously claimed.

src/deadline_worker_agent/sessions/runtime/rust.py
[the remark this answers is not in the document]
seant-aws: Checked, and the second of your two possibilities is the correct one — `deserialize_step` does not validate `Param` references at all, so `run_task` cannot have this bug class.

The two decoders sit on opposite sides of the template/job divide. `py_deserialize_step` deserializes a **job-side** `openjd_model::job::Step` via plain `serde_json::from_str` with no template validation; its doc comment in `rust-bindings/src/model/create_job_fns.rs` states the `BatchGetJobEntity` payload "is a serialized `openjd_model::job::Step`, not a template-side StepTemplate". A job-side Step is post-instantiation, so its format strings carry `{{Param.X}}` unresolved by design and the session resolves them at run time using `job_parameter_values`. `decode_environment_template` decodes a **template-side** document, and template decoding is where reference validation lives — which is exactly why the environment path failed and the task path never did.

This also answers your second concern: `parameterDefinitions` is the spec-sanctioned mechanism rather than a workaround. OpenJD 2023-09 §1.2 defines it as job parameter declarations carried by a standalone environment template, and requires the empty-list omission this PR implements ("If provided, then this list must contain at least one element").

Credit for the cross-check to @leongdl, who verified this independently against the same source and withdrew the equivalent finding (R3) in this review.

AlexTranAmz (approved):

leongdl (approved):

PR #1061, opened 2026-08-18
fix: give step-scoped environments their step's name and let bindings

seant-aws:

A step-scoped environment's scripts can reference `{{ Step.Name }}` and the step's `let` values, but the worker entered every environment without that context, so those references failed to resolve. `openjd.sessions.Session.enter_environment` already accepts `step_name` and `extra_let_bindings` — the worker simply never passed them.

Resolve the step from the environment id. Environment ids are scope-prefixed and carry their scope's id (`STEP:step-<uuid>:<name>` vs `JOB:job-<uuid>:<name>`), so `_step_id_from_environment_id()` parses the prefix and the step entity is fetched from that. The context then flows `EnterEnvironmentAction` → `Session` → `SessionRuntime` → both adapters.

Two properties fall out of using the id rather than the action's position in the queue, and a position-based discriminator gets both wrong:

- A step may declare **several** step environments, so they arrive as consecutive `ENV_ENTER`s before the `TASK_RUN`. Each resolves its own step context independently of ordering.
- A **job**-scoped environment is never handed a step's context. It is entered once and shared across steps, so binding it to any single step would be a silent wrong answer rather than a failure.

`exit_environment` needs no change — openjd-sessions replays the values recorded at entry.

Step-scoped environment scripts can resolve `Step.Name` and step-level `let` bindings. Job-scoped environments are unaffected. Both runtime adapters gained the two keyword arguments; they default to `None`, so callers that do not supply them behave as before.

If the step entity cannot be resolved, the environment still enters without step context (previous behaviour) and the step's task run reports the underlying entity error.

The Rust adapter no longer forwards `step_name` or `extra_let_bindings` to `_v1.Session.enter_environment`, which does not accept those keywords. The Python adapter continues to forward both. Equivalent Rust context flows through `resolved_symtab` in a separate change. This prevents the unsupported-keyword crash but does not provide end-to-end Rust step-environment support (upstream `decode_environment_template` rejects step expressions before any symbol table is read).

`hatch run lint` clean; full unit suite 3079 passed / 39 skipped.

New `test/unit/scheduler/test_step_env_context.py` covers the parser (step-scoped, job-scoped, colons in the environment name, missing separator, empty step id, unprefixed) and the dequeue path: a `STEP:` environment receives the template's real name and `let` list, a `JOB:` environment receives neither and does not trigger a step-entity lookup, two consecutive `STEP:` enters for one step **both** receive context, and an entity-resolution failure degrades to no context.

Sensitivity checked by mutation: forcing `_step_id_from_environment_id` to always return `None` fails 4 of the new tests, including the two-consecutive-environments case; reverting restores green.

Not verified: no end-to-end run. Submitting a job whose step environment references `Step.Name` currently depends on service-side support for requesting the relevant OpenJD extensions, so this is covered at the unit tier only.

No — no public API or configuration surface changes. The non-obvious part (why the discriminator is the environment id rather than queue position) is documented in the helper's docstring and at the call site.

No. The new parameters are keyword-only with `None` defaults, and behaviour for job-scoped environments is unchanged.

----

### DIFF
--- a/src/deadline_worker_agent/scheduler/session_queue.py
+++ b/src/deadline_worker_agent/scheduler/session_queue.py
@@ -61,6 +61,25 @@ else:
 
 logger = getLogger(__name__)
 
+_STEP_ENVIRONMENT_ID_PREFIX = "STEP:"
+
+
+def _step_id_from_environment_id(environment_id: str) -> str | None:
+    """Extract the step id from a step-scoped environment id.
+
+    Environment ids are scope-prefixed and carry their scope's id, as in
+    ``STEP:step-<uuid>:<environment name>`` or ``JOB:job-<uuid>:<name>``.
+    Returns None for anything that is not step-scoped, so job-scoped
+    environments are never given a step's context.
+    """
+    if not environment_id.startswith(_STEP_ENVIRONMENT_ID_PREFIX):
+        return None
+    remainder = environment_id[len(_STEP_ENVIRONMENT_ID_PREFIX) :]
+    step_id, separator, _ = remainder.partition(":")
+    if not separator or not step_id:
+        return None
+    return step_id
+
 
 @dataclass(frozen=True)
 class SessionActionQueueEntry(Generic[D]):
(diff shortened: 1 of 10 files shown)


leongdl (dismissed):

seant-aws:
**Amendment (force-push `57957c7`):**

Fixed the Rust adapter's `enter_environment` to not forward `step_name`/`extra_let_bindings` to the `_v1` session, which does not accept those keywords. The previous implementation used a conditional-kwargs dict that would pass them through when non-None, causing a `TypeError` on every step-scoped `ENV_ENTER` action on the Rust runtime.

Changes in this push:
- **`rust.py`**: Replaced the dynamic `enter_kwargs` dict + conditionals with a direct 3-argument call. Added a comment explaining the deliberate omission (Rust receives equivalent context via `resolved_symtab`).
- **`test_python.py`**: Added `test_enter_environment_forwards_step_context_to_wrapped_session` — asserts non-None `step_name` and `extra_let_bindings` reach the Python session.
- **`test_rust.py`**: Added `test_enter_environment_omits_step_context_from_rust_session` — asserts non-None `step_name` and `extra_let_bindings` do NOT appear in the `_v1` call.

Lint clean, 3081 tests passed.

leongdl (dismissed):
Github comments seems to be nits - can be small followups.

AlexTranAmz (dismissed):

godobyte (dismissed):

AlexTranAmz (approved):

godobyte (approved):

PR #1064, opened 2026-08-18
feat: add support for EXPR parameter types in API response parsing

seant-aws:

The service now supports 8 new EXPR parameter types (`bool`, `rangeExpr`, `stringList`, `pathList`, `intList`, `floatList`, `boolList`, `intListList`) via CreateJob. When these parameters reach the worker agent through GetJob/UpdateWorkerSchedule, `parameters_from_api_response()` raises `ValueError` on the unrecognized wire keys, crashing the session.

Added TypedDict definitions, union type updates, and elif parsing branches for all 8 new parameter types. The wire keys map to their corresponding `ParameterValueType` enum values already present in openjd-model 0.11.2+.

Workers can now parse jobs that use EXPR parameter types without crashing. No behavioral change for existing parameter types.

- 8 new parametrized unit test cases (all pass)
- `hatch run lint` clean (ruff + format + mypy)
- Full unit suite: 3078 passed, 39 skipped

No documentation changes needed — internal parsing logic only.

No.

### DIFF
--- a/src/deadline_worker_agent/sessions/job_entities/job_details.py
+++ b/src/deadline_worker_agent/sessions/job_entities/job_details.py
@@ -22,13 +22,21 @@ from openjd.sessions import (
 from openjd.sessions import PathMappingRule as OPENJDPathMappingRule
 
 from ...api_models import (
+    BoolListParameter,
+    BoolParameter,
+    FloatListParameter,
     FloatParameter,
+    IntListListParameter,
+    IntListParameter,
     IntParameter,
     JobDetailsData,
     JobAttachmentQueueSettings as JobAttachmentSettingsBoto,
     JobRunAsUser as JobRunAsUserModel,
+    PathListParameter,
     PathMappingRule,
     PathParameter,
+    RangeExprParameter,
+    StringListParameter,
     StringParameter,
     ChunkIntParameter,
 )
@@ -40,7 +48,20 @@ from .validation import Field, validate_object
 def parameters_from_api_response(
     params: dict[
         str,
-        StringParameter | PathParameter | IntParameter | FloatParameter | ChunkIntParameter | str,
+        StringParameter
+        | PathParameter
+        | IntParameter
+        | FloatParameter
+        | ChunkIntParameter
+        | BoolParameter
+        | RangeExprParameter
+        | StringListParameter
+        | PathListParameter
+        | IntListParameter
+        | FloatListParameter
+        | BoolListParameter
+        | IntListListParameter
+        | str,
     ],
 ) -> dict[str, ParameterValue]:
     result = dict[str, ParameterValue]()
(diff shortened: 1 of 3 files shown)


wyongzhi:
Do downstream consumers (openjd-sessions / the rust runtime symbol table) accept string elements for LIST_INT/LIST_FLOAT/LIST_LIST_INT, or do we need the same wire-to-native coercion the DataPlane does before feeding openjd-model (In CR, `_openjd_native_parameter_value`: `["1","2"]` → `[1,2]`)? The unit tests only assert the verbatim mapping, not downstream consumption — an end-to-end test with an intList parameter through a real session would settle it.

src/deadline_worker_agent/sessions/job_entities/job_details.py
```diff
@@ -60,6 +81,38 @@ def parameters_from_api_response(
...
+        elif "boolList" in value:
+            value = cast(BoolListParameter, value)
+            param_value = ParameterValue(type=ParameterValueType.LIST_BOOL, value=value["boolList"])
+        elif "intListList" in value:
+            value = cast(IntListListParameter, value)
+            param_value = ParameterValue(
+                type=ParameterValueType.LIST_LIST_INT, value=value["intListList"]
+            )
         else:
             raise ValueError(f"Parameter {name} -- unknown form in API response: {str(value)}")
```
wyongzhi, line 117: Non-blocking: since we're touching this function — the error message interpolates `str(value)`, which puts customer parameter values into logs/exceptions. The DataPlane side is careful to echo parameter names but never values in its validation errors. Suggest trimming to the key names only, e.g. `f"Parameter {name} -- unknown form in API response: keys={sorted(value)}"`.

test/unit/sessions/job_entities/test_parameters_from_api_response.py
wyongzhi, line 1: Non-blocking, two small test gaps:
1. No negative case pinning the fallthrough behavior — a still-unknown key (e.g. {"futureType": "x"}) should keep raising ValueError. One more parametrize row would lock that in.
2. The new cases assert the verbatim mapping (intList stays ["1","2","3"] as strings), which is correct — the string→int coercion happens engine-side in build_symbol_table (openjd-model rust-bindings symbol_table.rs, from_str_coerce). Worth a one-line comment in the test so a future reader doesn't mistake the string elements for a bug; DP coerces before preprocess_job_parameters for exactly these values, and the asymmetry is intentional.

wyongzhi (approved):

test/unit/sessions/job_entities/test_parameters_from_api_response.py
seant-aws: to answer 2: 

No coercion needed on the worker side. The Rust EXPR engine (build_symbol_table) is the designed coercion point — it accepts ["1","2"] with a list[int] type spec and coerces string elements to native ints internally. Both the v0 Python session and _v1 Rust session converge on this same engine. The DataPlane's _openjd_native_parameter_value() serves the CreateJob validation path (where input arrives as dict[str,str]), not the session execution path.

Confirmed via runtime test: build_symbol_table({"Param.X": ["1","2","3"]}, {"Param.X": "list[int]"}) produces [1, 2, 3] as native ints.

leongdl (approved):

PR #1065, opened 2026-08-19
fix: accept EXPR parameter types in job details entity validation

seant-aws:

Follow-up to #1064. That PR added EXPR parameter parsing to `parameters_from_api_response()`, but `JobDetails.validate_entity_data` runs `_validate_job_parameters()` **before** `from_boto()` reaches the parser. The validator hard-coded the type-key allowlist to `("string", "path", "int", "float")` and required all values to be strings, so job parameters using any EXPR type were rejected at entity validation time — the #1064 branches were unreachable on the BatchGetJobEntity path. Only the UpdateWorkerSchedule task-parameter path was actually fixed.

- Widened `_validate_job_parameters()` with a per-type value-shape check table covering all 12 wire-format types (original 4 scalars + 8 EXPR forms with their native shapes: boolean, list of strings, list of booleans, list of lists of strings).
- Also trimmed the unknown-form `ValueError` in `parameters_from_api_response()` to report key names only, so customer parameter values do not leak into logs (review feedback on #1064).

Job-level EXPR parameters returned by BatchGetJobEntity now pass entity validation and parse end to end. Malformed shapes are still rejected with per-type errors.

- New entity-validation test cases: 1 success case covering all 8 EXPR forms + 5 failure cases for mismatched value shapes. The success case fails against the previous validator (verified by running the same inputs against the prior mainline commit).
- Full unit suite: 3082 passed, 39 skipped. `hatch run lint` clean (ruff + format + mypy).

No documentation changes needed — internal validation logic only.

No.

### DIFF
--- a/src/deadline_worker_agent/sessions/job_entities/job_details.py
+++ b/src/deadline_worker_agent/sessions/job_entities/job_details.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 from dataclasses import dataclass, field
 from pathlib import PurePath, PurePosixPath, PureWindowsPath
-from typing import Any, cast
+from typing import Any, Callable, cast
 import os
 
 from openjd.model import (
@@ -486,6 +486,27 @@ class JobDetails:
 
     @classmethod
     def _validate_job_parameters(cls, job_parameters: dict[str, Any]) -> None:
+        # Maps each accepted wire-format type key to a predicate that checks the
+        # value has the shape the service sends for that type.
+        def _is_str_list(value: Any) -> bool:
+            return isinstance(value, list) and all(isinstance(item, str) for item in value)
+
+        value_shape_checks: dict[str, Callable[[Any], bool]] = {
+            "string": lambda v: isinstance(v, str),
+            "path": lambda v: isinstance(v, str),
+            "int": lambda v: isinstance(v, str),
+            "float": lambda v: isinstance(v, str),
+            "chunkInt": lambda v: isinstance(v, str),
+            "bool": lambda v: isinstance(v, bool),
+            "rangeExpr": lambda v: isinstance(v, str),
+            "stringList": _is_str_list,
+            "pathList": _is_str_list,
+            "intList": _is_str_list,
+            "floatList": _is_str_list,
+            "boolList": lambda v: isinstance(v, list) and all(isinstance(item, bool) for item in v),
+            "intListList": lambda v: isinstance(v, list) and all(_is_str_list(item) for item in v),
+        }
+
         for key, value in job_parameters.items():
             if not isinstance(value, dict):
                 raise ValueError(f'Expected parameters["{key}"] to be a dict but got {type(value)}')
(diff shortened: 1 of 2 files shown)


leongdl (commented):
Reviewed — this correctly fixes the issue from #1064: the validator previously hard-rejected anything outside `("string", "path", "int", "float")` before `parameters_from_api_response()` ever ran, so the EXPR parsing branches were unreachable on the BatchGetJobEntity path. The new per-type shape-check table covers all 8 EXPR forms, and verifying the success case fails against the prior validator is the right falsifiability check. The failure cases cover each shape class well (bool-as-string, non-list, wrong element types, non-nested list), and trimming parameter values out of the unknown-form error is a good PII fix.

Also verified there are no other hardcoded type-key allowlists in the package — the only other caller of the parser (task parameters in `session_queue.py` dequeue) has no pre-validator, so there's no second mismatch site.

LGTM.

leongdl (dismissed):

leongdl:
**Why the validator doesn't have — and doesn't need — `chunkInt`:**

Checked the Open Job Description specification to be sure. The TASK_CHUNKING RFC ([rfcs/0001-task-chunking.md](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md)) introduces `CHUNK_INT` solely as a new task parameter type — it amends the `<TaskParameterDefinition>` grammar:

```
<TaskParameterDefinition> ::= <IntTaskParameterDefinition> | <FloatTaskParameterDefinition> |
                              <StringTaskParameterDefinition> | <PathTaskParameterDefinition> |
                              <ChunkIntTaskParameterDefinition>
```

with the new section 3.4.1.5 `<ChunkIntTaskParameterDefinition>`. The job parameter definition grammar is untouched, so a job parameter can never have type `CHUNK_INT`. The RFC even calls this out directly — chunking is resolved per session action, and "modifying this as a job parameter would not serve a purpose."

So `_validate_job_parameters()` validating jobDetails entities is correct to accept only the 12 job-parameter types (4 base + 8 EXPR). `chunkInt` remains handled by the shared `parameters_from_api_response()` parser exclusively for the task-parameter path (TASK_RUN actions in `session_queue.py`), which is the only place it can legitimately appear.

src/deadline_worker_agent/sessions/job_entities/job_details.py
```diff
@@ -114,7 +114,9 @@ def parameters_from_api_response(
                 type=ParameterValueType.LIST_LIST_INT, value=value["intListList"]
             )
         else:
-            raise ValueError(f"Parameter {name} -- unknown form in API response: {str(value)}")
+            raise ValueError(
+                f"Parameter {name} -- unknown form in API response: keys={sorted(value)}"
```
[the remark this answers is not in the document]
jericht: @seant-aws I think we should undo the `sorted` here - for the `str` case above, would make the error message unreadable

Let's address this one and I think we can merge

src/deadline_worker_agent/sessions/job_entities/job_details.py
```diff
@@ -486,6 +488,26 @@ def validate_entity_data(
...
+            "float": lambda v: isinstance(v, str),
+            "bool": lambda v: isinstance(v, bool),
+            "rangeExpr": lambda v: isinstance(v, str),
+            "stringList": _is_str_list,
+            "pathList": _is_str_list,
+            "intList": _is_str_list,
+            "floatList": _is_str_list,
+            "boolList": lambda v: isinstance(v, list) and all(isinstance(item, bool) for item in v),
+            "intListList": lambda v: isinstance(v, list) and all(_is_str_list(item) for item in v),
+        }
```
jericht, line 508: nit: we have this list of possible value types in multiple places now. Might be good to put the mapping in a variable

also please address the missing `chunkInt` mapping, I think it safer to have a check for it even if it goes unused today rather than assuming callers will never pass `chunkInt` here

seant-aws: yup, added

leongdl (approved):

jericht (approved):

PR #1066, opened 2026-08-19
fix: resolve Rust parameter type enum members by name, not value

seant-aws:

`_to_rust_parameter_values()` resolved the `_v1` binding's enum member via `getattr(type_cls, value.type.value)`. That only works when the enum value equals the member name. Bracketed values are not attribute names: `ParameterValueType.LIST_STRING.value` is `"LIST[STRING]"` while the binding's member is spelled `LIST_STRING`, so the lookup returned `None` and raised "type not defined" for types the binding does define. All six `LIST[*]` job parameter types and `CHUNK[INT]` task parameters were affected under the Rust session runtime. The Python runtime is unaffected.

Resolve by enum member name (`value.type.name`), keeping the human-readable value in the error message. Types genuinely absent from a binding (e.g. EXPR types on `TaskParameterType`, which are job-parameter-only per the Open Job Description specification) still fail loud.

EXPR list-typed job parameters and chunked task parameters now convert correctly on the Rust session runtime.

Parametrized conversion tests for all 8 EXPR job parameter types, a `CHUNK[INT]` task parameter case, and a case asserting EXPR types on the task binding still raise. All new cases were run against the previous lookup: the 7 affected ones fail as expected (`bool`/`rangeExpr` pass under both since their value equals their name). Full unit suite: 3086 passed, 39 skipped. Lint clean.

No — internal conversion logic only.

No.

### DIFF
--- a/src/deadline_worker_agent/sessions/runtime/rust.py
+++ b/src/deadline_worker_agent/sessions/runtime/rust.py
@@ -135,7 +135,10 @@ def _to_rust_parameter_values(
         if isinstance(value, dict):
             converted[name] = value
             continue
-        native_type = getattr(type_cls, value.type.value, None)
+        # Resolve by enum member name, not value: bracketed type values like
+        # "LIST[STRING]" and "CHUNK[INT]" are not valid attribute names, but
+        # their members are spelled LIST_STRING / CHUNK_INT in the binding.
+        native_type = getattr(type_cls, value.type.name, None)
         if native_type is None:
             raise ValueError(
                 f"Parameter {name!r} has type {value.type.value!r}, which the "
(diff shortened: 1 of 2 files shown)


godobyte (approved):

src/deadline_worker_agent/sessions/runtime/rust.py
```diff
@@ -140,7 +140,10 @@ def _to_rust_parameter_values(
         if isinstance(value, dict):
             converted[name] = value
             continue
-        native_type = getattr(type_cls, value.type.value, None)
+        # Resolve by enum member name, not value: bracketed type values like
+        # "LIST[STRING]" and "CHUNK[INT]" are not valid attribute names, but
+        # their members are spelled LIST_STRING / CHUNK_INT in the binding.
+        native_type = getattr(type_cls, value.type.name, None)
```
godobyte, line 146: nit - the comment might be confusing once the prior resolve by value line gets removed.

wyongzhi (approved):

leongdl (approved):

PR #1076, opened 2026-08-25
feat: deliver step-scope let to Python session

leongdl:
> [!IMPORTANT]
> **Blocked on a dependency release.** This needs
> OpenJobDescription/openjd-sessions-for-python#356 merged and released as
> `openjd-sessions` 0.11.1. The pin in this PR points at 0.11.1, so CI cannot
> install until that release exists. Please do not merge before then.

## What changed

`RunStepTaskAction` now passes the step's step-template-scope `let` bindings
(`StepTemplate.let`) through to the session runtime, and the Python runtime
forwards them to `openjd.sessions.Session.run_task(extra_let_bindings=...)`.

Plumbing, five files:

| File | Change |
|---|---|
| `sessions/actions/run_step_task.py` | passes `extra_let_bindings=self._details.step_template.let` |
| `sessions/session.py` | pass-through to the runtime |
| `sessions/runtime/_abc.py` | new parameter on the abstract `run_task` |
| `sessions/runtime/python.py` | forwards to the openjd session |
| `sessions/runtime/rust.py` | accepts and ignores (see below) |

## Why

A step's step-template-scope `let` bindings (RFC 0005 §3.6) never reached the
task when running under the Python session runtime, so any reference to a
step-scope name failed to resolve with `Undefined variable`.

Those bindings normally resolve at job instantiation — `create_job` folds
`StepTemplate.let` into `StepTemplate.script.let` via `resolve_syntax_sugar`.
This agent is handed a service-resolved but **un-instantiated** `StepTemplate`,
where `let` and `script.let` arrive as separate fields, and it never calls
`resolve_syntax_sugar`. So the fold never happened for what this agent sees, and
the step-scope names were absent from the task's symbol table.

Environments already had this channel: `enter_environment` takes
`extra_let_bindings`, and `scheduler/session_queue.py` extracts
`step_template.let` for `ENV_ENTER` actions. Task runs were left out.

## Why the Rust runtime ignores it

The `_v1` session receives the same values inside `resolved_symbol_table_json`,
which `create_job` pre-resolved and which that session uses as the base of its
per-action symbol table. Forwarding the bindings as well would define the same
names twice. This matches how `enter_environment` already drops the parameter on
that path. The behaviour is pinned by a test, so a future change to forward them
trips it rather than passing silently.

## Alternative considered

Fold `StepTemplate.let` into `script.let` in this package by calling
`resolve_syntax_sugar`. Rejected: that transform is not idempotent — a second
call duplicates the bindings — so it puts a double-apply hazard in this code
path, and it duplicates model logic outside the model.

## Tests

13 new tests.

`test/unit/sessions/test_step_scope_let_end_to_end.py` (6) drives the real chain
and asserts on the text the task actually emitted, not on mock call arguments:

```
StepDetails.from_boto  ->  RunStepTaskAction.start  ->  Session.run_task
  ->  PythonSessionRuntime.run_task  ->  openjd.sessions.Session  ->  subprocess
```

Only the scheduler-facing `Session` orchestrator is stood in for; its `run_task`
is a one-line pass-through, pinned separately in `test_session.py`. The entity is
built from the real served `BatchGetJobEntity` payload shape, so the test
exercises the actual condition — `let` and `script.let` as separate unfolded
fields.

Covered: a step-scope binding resolving in the command; both scopes resolving
with script scope shadowing step scope; a binding referencing `Step.Name`; an
unresolvable binding failing the action cleanly; and two negative controls — a
step with no `let`, and script-scope `let` alone — so an over-correction fails
something.

The remaining 7 pin the plumbing in `test_python.py`, `test_rust.py`,
`test_session.py` and `actions/test_run_step_task.py`.

Verification performed:

- **Mutation-checked, 5/5 caught.** Reverting the action to its pre-fix state,
  making it send `None`, dropping the Python adapter's forwarding, dropping the
  `enter_environment` forwarding, and making the Rust adapter forward — each
  fails at least one new test, and the negative controls survive every one.
- **Full unit suite: 3162 passed / 47 skipped**, measured against a
  3149-passed / 47-skipped baseline. The +13 is exactly the tests added; no
  pre-existing test changed behaviour.
- `ruff format`, `ruff check` and `mypy` clean across 217 files.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,10 @@ dependencies = [
     "boto3 >= 1.34.75",
     "deadline-job-attachments == 0.1.3",
     # Pinned to patch version due to Host Config Script runner usage of private OpenJD Sessions API.
-    "openjd-sessions == 0.11.0",
+    # 0.11.1 is the floor: it adds run_task(extra_let_bindings=...), which this
+    # package now passes to deliver step-template-scope `let` to the Python
+    # session. Against 0.11.0 that call raises TypeError.
+    "openjd-sessions == 0.11.1",
     "openjd-model >= 0.11.4, < 0.12",
     # tomli became tomllib in standard library in Python 3.11
     "tomli == 2.0.* ; python_version<'3.11'",
--- a/src/deadline_worker_agent/sessions/actions/run_step_task.py
+++ b/src/deadline_worker_agent/sessions/actions/run_step_task.py
@@ -80,10 +80,17 @@ class RunStepTaskAction(OpenjdAction):
             env_vars["DEADLINE_TASK_ID"] = self.task_id
 
         # The service resolves step template syntax sugar, so script is always present.
+        #
+        # extra_let_bindings: the step template arrives un-instantiated, with
+        # step-scope `let` (StepTemplate.let) and script-scope `let`
+        # (StepTemplate.script.let) as separate fields. Nothing folds the former
+        # into the latter on this path, so pass it explicitly or step-scope names
+        # are missing from the Python session's symbol table.
         session.run_task(
             step_script=cast("StepScript", self._details.step_template.script),
             task_parameter_values=self._task_parameter_values,
             os_env_vars=env_vars,
             step_name=self._details.step_template.name,
             resolved_symbol_table_json=self._details.resolved_symbol_table_json,
+            extra_let_bindings=self._details.step_template.let,
         )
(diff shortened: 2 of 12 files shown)


leongdl:
Superseded by #1077.

#1076 passed step-scope EXPR `let` to the session as source strings. The decision is that the service's resolved symbol table is the single authoritative source for those values, so #1077 forwards the table and removes this channel.

Closing rather than merging, since the two would be redundant. #1077's history contains this change and its removal; the net diff is the table work plus two fixes found while testing it on a fleet.

PR #1084, opened 2026-09-09
feat: default session_runtime to service-selected

seant-aws:
All fleets currently default to the Python session runtime. After a successful six-week canary bake of service-selected routing, the default should flip so all fleets defer runtime selection to the service by default.

Changed the `WorkerSettings.session_runtime` default from `PYTHON` to `SERVICE_SELECTED`, updated the installer config example, and updated the corresponding unit test.

Workers with no explicit `session_runtime` config now default to service-selected routing. The service resolves to Python when no runtime hint is provided, so behavior is unchanged for fleets without service-side routing configured. Operators can pin `session_runtime = "python"` to opt out.

- Red-before-green: flipped test expectation, observed failure, then applied source change
- Unit test matrix (Python 3.9/3.10/3.11): 3187 pass per version
- Lint (ruff + mypy + format): clean
- Package build: clean

Updated the installer config example (`worker.toml.example`) to reflect the new default.

No. Workers upgrade to service-selected routing, but the runtime resolves to Python when no hint is provided. Reverting is a single config change (`session_runtime = "python"`).

### DIFF
--- a/src/deadline_worker_agent/config/settings.py
+++ b/src/deadline_worker_agent/config/settings.py
@@ -95,7 +95,7 @@ class WorkerSettings(BaseSettings):
         If true, then the Worker Agent's logs are structured.
     session_runtime : SessionRuntimeKind
         The session runtime implementation to use. One of 'python', 'rust', or
-        'service-selected'. Defaults to 'python'.
+        'service-selected'. Defaults to 'service-selected'.
     """
 
     farm_id: str = Field(regex=r"^farm-[a-z0-9]{32}$")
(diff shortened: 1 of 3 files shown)


jericht (approved):

jericht:
LGTM after you figure out the unit test failure

seant-aws:
note: unit tests failing. seems to be a upstream change in openjd-model that needs a test update. following up a another PR

seant-aws:
https://github.com/aws-deadline/deadline-cloud-worker-agent/pull/1085

fix was here

leongdl (approved):

PR #1092, opened 2026-09-14
fix: enter step environments that reference step-scope let bindings on the Rust runtime

seant-aws:

A step environment whose template references the declaring step's `let` bindings
could not be entered on the Rust session runtime. The adapter treated the
environment as a *template* and validated it with `decode_environment_template`,
which checks format-string references against template scope — where a step's
bindings do not exist:

```
Action failed to start: 2 validation errors for EnvironmentTemplate
environment -> variables -> MY_LABEL:
	Failed to parse interpolation expression at [0, 11]. Undefined variable: 'label'.
environment -> variables -> MY_VAL:
	Failed to parse interpolation expression at [0, 9]. Undefined variable: 'val'.
```

The failure landed before any session existed, so the resolved symbol table that
supplies those values was never reached. `Step.Name` failed the same way. The
Python runtime was unaffected because it never re-validates: it hands the parsed
model straight to openjd and resolves at enter time.

Stop treating the environment as a template. The service has already instantiated
it, so deserialize it job-side instead — wrap it as the sole step environment of a
placeholder step, call `deserialize_step`, and take the environment back out. This
mirrors how a step script is already deserialized in the same adapter. The job-side
deserializer performs no template-scope reference check, so names that resolve
later — the step's bindings, job parameters, `Step.Name` — pass through unresolved
and are supplied by the resolved symbol table when the environment actually enters.
That is the same ordering the Python runtime has always used.

An earlier revision of this branch injected the step's `let` declarations into the
lifted template instead. That only worked for environments that have a script: an
environment script must declare actions, so a variables-only environment had
nowhere legal to carry a binding list. It is replaced here, along with everything
that propped it up — the script-with-actions guard, the `Step.Name` drop, the
binding-name collision dedup, and the parameter-definition synthesis that existed
only to satisfy the decode. Job parameters still resolve, from the values the
session already holds; extensions are still gated by the profile the session is
built with.

Step environments referencing step-scope bindings now enter on the Rust runtime,
including variables-only environments. No behaviour change on the Python runtime.
Environment templates are no longer validated at enter time on the Rust path — a
malformed template now surfaces from the session rather than from the decode, which
matches the Python runtime's behaviour and is reported through the same
action-failure path.

Unit: red-green throughout. The tests covering the new behaviour were observed
failing against the previous mechanism (`ModelValidationError: Undefined variable`)
before passing against this one. `hatch build`, `hatch run all:test`
(3187 passed, 39 skipped) and `hatch run lint` (ruff check + format + mypy) all pass.

End to end: the OpenJD conformance fixture
[`2023-09/EXPR/jobs/3.6--let-step-bindings-in-step-env`](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/conformance-tests/2023-09/EXPR/jobs/3.6--let-step-bindings-in-step-env.test.yaml)
was submitted through the service to a worker running this branch with the Rust
runtime pinned. It declares two step environments — one variables-only, one
scripted — both referencing the step's `let` bindings, and it fails on the Rust
runtime without this change.

```
Job:     job-76698faa5be54b9fa791ff349cd2a73c   →  SUCCEEDED
Worker:  agent 0.33.2.post2, commit 6ba9d8dc
Runtime: "Selected session runtime: rust (hint='pythonexpr')"
Actions: 9/9 SUCCEEDED

PASS  ENTER_VAL:21          PASS  ENV_VAL:21        PASS  EXIT_VAL:21
PASS  ENTER_LABEL:item_21   PASS  ENV_LABEL:item_21 PASS  EXIT_LABEL:item_21
PASS  TASK_VAL:21
```

```
| variant        | attempt | exit | result                                                 |
|----------------|---------|------|--------------------------------------------------------|
| linux-python   |    1    |  0   | 53 passed, 26 skipped                        (1:31:08) |
| linux-rust     |    1    |  0   | 53 passed, 26 skipped                        (1:33:16) |
| windows-python |    1    |  0   | 51 passed, 20 skipped, 2 xfailed, 6 xpassed  (2:19:07) |
| windows-rust   |    1    |  0   | 51 passed, 20 skipped, 2 xfailed, 6 xpassed  (2:22:44) |

00:28:14Z → 08:16:02Z  (7h48m) · commit 6ba9d8d · wheel 0.33.2.post1.dev2
zero FAILED/ERROR lines in any log · no retries needed
```

All seven expected output lines matched. The `Runtime:` line is the agent's own
log and confirms the configured pin took effect over the service's hint, so the
Rust path is what ran. `ENV_VAL`/`ENV_LABEL` come from the variables-only
environment — the case the previous mechanism could not reach — and resolve from
the service-supplied symbol table.

No — internal behaviour of the Rust session runtime adapter, no user-facing surface
changes.

No.

---

Before this leaves draft: the `step_let_declarations` parameter threaded by the
first commit is unused now that neither runtime lifts the environment, so it and
the stale wording on the abstract method will be dropped and the branch squashed
to a single commit.


seant-aws:
Running conformance test as job templates to submit locally with changes; verifying with fix currently.

leongdl:
Reviewed the diff and traced the openjd-rs side. The fix is correct and I'd ship it. Four asks below, then a refactor recommendation and a follow-up.

## Why the Rust CLI never hit this

Worth recording, because it explains why the bug is specific to the worker's shape rather than to openjd-rs.

`openjd run` decodes the job template once, calls `create_job`, and hands the session the already-instantiated `job::Environment` straight out of the step that declares it — `crates/openjd-cli/src/run/execution.rs:386-393`:

```rust
if let Some(step_envs) = &step.step_environments {
    for env in step_envs {
        ctx.enter_environment(env, &env.name, step.resolved_symtab.clone()).await;
```

Validation happened earlier, in the right scope. At `crates/openjd-model/src/template/validate_v2023_09/format_strings.rs:844-847` the step-environment branch builds its symtab with `is_step_env = true`, which seeds `Step.Name` (line 138 of `build_session_scope_symtab`), and lines 849-866 copy the declaring step's evaluated `let` values into that same table.

The old worker path went somewhere else. `decode_environment_template` routes to `validate_format_strings_environment_template`, whose own doc comment at `format_strings.rs:1008-1010` says it outright: "`Step.Name` is available in neither scope — an environment template is not attached to a step." Line 1026 calls `build_session_scope_symtab(..., is_step_env = false, ...)` and there is no step whose `let` values could be copied in. The CLI only reaches that function for `--environment` files (`execution.rs:94`), which are job-scoped by construction.

So this was never reproducible on the CLI, and the diagnosis in the description is right.

## The approach is sound, for a reason the description doesn't give

I checked the thing the swap actually rests on: whether reading a template-shaped environment body directly as `job::Environment` is equivalent to the `convert_environment` the old path went through.

It is. `convert_environment_with_symtab` (`instantiate.rs:312-345`) is a pure structural copy — clone the name, unwrap `Description`'s newtype, map actions field-for-field, clone `variables` as `FormatString` unchanged. Nothing is resolved or retyped. And the job-side types are wire supersets of the template-side ones: `job::EnvironmentScript` accepts `let` plus a `letBindings` alias where the template accepts only `let`; `job::EmbeddedFile` accepts `fileType` *and* `type` where the template accepts only `type`; the two `CancelationMode` `Deserialize` impls are near-duplicate hand-written parsers over the identical `{"mode": ...}` shape. Serde reaches the same object `convert_environment` would have built.

I also read the binding, since "does not perform template-scope reference validation" is load-bearing. `py_deserialize_step` in openjd-model-for-python `rust-bindings/src/model/create_job_fns.rs:351-362` is `json.dumps` then `serde_json::from_str::<job::Step>`. Nothing else.

## Asks

**1. Drop `step_let_declarations` and everything it drags.** Already flagged in the description. Removal also takes the `step_template.let` read in `session_queue.py`, both new `test_step_env_context.py` cases, `test_enter_environment_accepts_and_ignores_step_let_declarations`, and the `_abc.py` docstring wording about a runtime that "lifts the environment into a standalone document and re-decodes it" — no runtime does that anymore. `step_name` should stay; `python.py` forwards it to the wrapped session, so it has a live consumer.

**2. Soften the extension-gating claim.** "Extensions are still gated by the profile the session is built with" is only true for function availability. `ModelProfile` reaches the session (`rust.py:308`) and `derive_library` at `crates/openjd-sessions/src/session.rs:445-461` uses it for the `FunctionLibrary` and redaction — that is all. The grammar-level checks that lived in `validate_env_format_strings` are gone from this path: "complex expressions require the EXPR extension" on an env variable, and "'let' requires the EXPR extension" on an env script.

This is a code-reading finding, not a measured one; confirming it means running a session with a non-EXPR profile against an env variable containing `{{ 1 + 1 }}`. Two things make it low severity rather than a blocker: the old path passed `RUNTIME_CAPABILITY_EXTENSIONS`, the full ceiling rather than the job's narrowed set, so it would only ever have rejected extensions no runtime supports at all; and the service validated the template at submit time. But that second point is a claim about another system, so name the check you're relying on rather than leaving the sentence to imply a defence that no longer exists.

**3. One verified regression worth a line in the description.** Every template-side type carries `#[serde(deny_unknown_fields)]` — `template::Environment`, `EnvironmentScript`, `EmbeddedFile` (`template/environment.rs:17,26,36`), `Action` (`template/actions.rs:13`). On the job side it survives only on `EnvironmentScript` and `StepScript`; `job::Environment`, `job::Action`, `job::EmbeddedFile` and `job::EnvironmentActions` all lack it. Unknown keys in the environment body now pass silently where they used to be rejected.

**4. Test nits.** `assert type(native_env).__module__ == "openjd.model._v1.job"` is brittle string-matching on a foreign module path; `isinstance` against the imported type pins the same property and survives a module move. Also worth confirming `_build_environment`'s `extensions: ["EXPR"]` is doing work in all three cases — the docstring says undefined-name references parse regardless, which suggests the variables-only case would pass without it.

## Refactor recommendation

Extract the wrapper into one module-level private helper in `rust.py`, and give `run_task`'s twin the same treatment:

```python
def _deserialize_environment(env_dump: dict[str, Any]) -> Environment: ...
def _deserialize_step_script(script_dump: dict[str, Any], step_name: str | None) -> StepScript: ...
```

Six lines of moved code, and it buys three things. The comment explaining why a `"Placeholder"` step exists lives in one place instead of being duplicated at both call sites. The `deserialize_step` coupling — `job::Step`'s required-field set, and the fabricated `{"actions": {"onRun": {"command": "true"}}}` that only exists to satisfy it — is stated once. And it makes the migration below a two-function-body change rather than surgery at two call sites.

The `enter_environment` wrapper is a strictly larger fabrication than `run_task`'s, which is worth being explicit about in the helper's comment: `run_task` wraps a real script, while this one invents a command. Inert today, since the `Step` is discarded and `deserialize_step` is pure serde, but it is a fake command string in the wire document.

## Follow-up: `deserialize_environment`

The wrapper works by shape coincidence, not by contract. `deserialize_step`'s own docstring is explicit that the dict should be a serialized `openjd_model::job::Step`; we're feeding it a template-side environment body inside a fabricated step.

The clean version is a `deserialize_environment(env_dict) -> Environment` binding upstream, and it's small — roughly 30 lines including a test:

| File | Change |
|---|---|
| `rust-bindings/src/model/create_job_fns.rs` | new `py_deserialize_environment`, ~12 lines, a copy of `py_deserialize_step:351-362` swapping `job::Step`→`job::Environment` and `PyStep`→`PyEnvironment` |
| `rust-bindings/src/model/mod.rs:24-27` | one name added to the existing `pub(crate) use create_job_fns::{...}` |
| `rust-bindings/src/lib.rs:225` | one `m.add_function(wrap_pyfunction!(...))` line |
| `src/openjd/_openjd_rs.pyi` | regenerated by `scripts/generate_stubs.sh`, not hand-edited |

`PyEnvironment` already exists (`rust-bindings/src/model/job.rs:400`) and is already exported, so there is no new pyclass — that's why it's small. The real cost is an openjd-model-for-python release plus a floor bump, not code.

Migration is mechanical, because both paths end in the same serde call on the same type: `deserialize_step({..., "stepEnvironments": [E]}).step_environments[0]` and `deserialize_environment(E)` both run `serde_json::from_str::<job::Environment>` over the identical `E`. Nothing about the accepted input, the produced type, or resolution semantics changes. The only behavioural delta is error text — a malformed environment currently surfaces as `failed to deserialize Step: ... stepEnvironments[0].variables ...` and afterwards the path is rooted at the environment, so check nothing asserts on or scrapes that string.

Three notes for whoever picks it up:

- Please file the issue and cite it from both helpers, next to the existing openjd-sessions-for-python#334 reference. Without a tracked issue the wrapper goes permanent by default; the `run_task` one has already survived long enough to get copied.
- The new `enter_environment` tests are migration-safe — `deserialize_step` runs unmocked and the assertions are on the resulting object. `test_rust.py:319-321` is not: it pins `{"name": "Placeholder", "script": ...}` and will need rewriting. Same pattern as the environment-side tests this PR deletes 297 lines of. Pre-existing, out of scope here, worth a note on the issue.
- I'd skip a `try/except ImportError` capability probe. Speculative plumbing for a function that doesn't exist, and it ships two paths where one is untestable until the release lands. A hard floor bump is cleaner.

Two things the migration does **not** fix, so the issue shouldn't claim them. The `deny_unknown_fields` gap stays — `job::Environment` has none whichever door you come through. And reading a template-shaped document as a job-side type stays a coincidence; only the service shipping the environment entity job-side removes that, the way it already ships `job::Step` per `deserialize_step`'s docstring. The type is already built for it — `job::Environment` carries its own `resolvedSymTab` field (`job/mod.rs:152-155`), and `Session::enter_environment`'s doc at `session.rs:1082-1086` treats the argument and the field as interchangeable. That asymmetry, steps job-side and environments template-shaped with a symbol table bolted on beside them, is the underlying defect. Not this PR's job.

## Separately, in openjd-rs

Nothing pins the template-side/job-side serde compatibility this approach depends on:

```
$ grep -rn "from_value::<job::Environment>\|from_str::<job::Environment>" crates/
(no output)
```

So a field added to `template::Environment` passes openjd-rs CI green and is silently dropped on the worker, because `job::Environment` has no `deny_unknown_fields` to object. A test that deserializes a template-shaped environment document as `job::Environment` and asserts equality with `convert_environment` of the same document parsed template-side would guard it. Cheap, lands in openjd-rs, no release coupling, and it guards the wrapper and the eventual binding equally. I'd not reach for adding `deny_unknown_fields` to `job::Environment` instead — that's a behaviour change on the job-side wire, and I haven't checked whether the service already sends keys openjd-rs doesn't model.

## What's good

The e2e evidence is the right shape: the actual conformance fixture through the real service, on a worker with the runtime pinned, with the `Runtime:` log line proving the pin took, plus the four-variant differential matrix. The red-green note is the right evidence for the unit tests. And the test rewrite is a genuine improvement — building a real model through `EnvironmentDetails.from_boto` and running `deserialize_step` unmocked pins behaviour, where the deleted tests pinned a lifted-template dict that was implementation detail. That's most of the −297.

leongdl (approved):

jericht (approved):

AlexTranAmz (approved):

